PR #1, opened 2016-05-17
Add initial version of node problem detector

Random-Liu:
Here is a simple implementation of NodeProblemDetector for https://github.com/kubernetes/kubernetes/issues/23028.

The node problem detector is mainly composed of two parts:
* **ProblemDetector**. It is responsible for:
  * Collecting and consolidating events and conditions from different problem daemons;
  * Translating and aggregating them into `Event`s and `NodeCondition`s;
  * Syncing with apiserver.
* **KernelMonitor** is a simple kernel log monitor for demonstrating the Node Problem API. It is one (the only one for now) of the problem daemons. It monitors the kernel log and reports problems to ProblemDetector.
  * KernelMonitor is in charge of the "kernel condition" node condition.
  * If there is a permanent kernel error detected, it will report "kernel condition" (only `KernelDeadlock` for now) with corresponding reason and message to the problem detector.
  * If there is a temporary error detected, it will report an event to the problem detector.
  * KernelMonitor uses https://github.com/hpcloud/tail to monitor the kernel log, and uses regular expression to match problems in the kernel log. The patterns of problems are configured in the config file `config/kernel-monitor.json`. Now some known kernel bugs (https://github.com/kubernetes/kubernetes/issues/20096, https://github.com/kubernetes/kubernetes/issues/19986) have been added, and it would be easy to support other issue only if the pattern is stable and can be expressed in regular expression.
  * A lot of TODOs:
    * Support more flexible pattern match such as only matching start&end line etc.
    * Support journald.
    * ...

The node problem detector is a DaemonSet. It needs a ConfigMap `node-problem-detector-config` generated from `config/` directory. The bootstrapping is dumb now. To start the node problem detector, you need to:

1. Make sure the config mount, kernel log mount and apiserver endpoint in node-problem-detector.yaml are properly configured.
2. Make sure the log path and issue patterns `config/kernel-monitor.json` are properly configured.
3. Create ConfigMap `node-problem-detector-config` from directory `config/`.
4. Create DaemonSet `node-problem-detector.yaml`.

Now ProblemDetector and KernelMonitor are running in the same container and of course the same DaemonSet. In the future, we may want to:
* Option 1: Move the logic of ProblemDetector into kubelet, and leave problem daemons (such as KernelMonitor) as separate DaemonSets reporting directly to kubelet.
* Option 2: Move ProblemDetector into a DaemonSet, all problem daemons run as separate DaemonSets reporting to ProblemDetector.
* Option 3: Run ProblemDetector and problem daemons in separate containers, but wrap them in one DaemonSet.

This has not been decided yet, and may need more discussion. :)

@dchen1107 
/cc @kubernetes/sig-node


config/kernel-monitor.json
```diff
@@ -0,0 +1,31 @@
...
+		},
+		{
+			"type": "permanent",
+			"reason": "AUFSUmountHung",
+			"pattern": "task umount\\.aufs:\\w+ blocked for more than \\w+ seconds\\."
+		},
+		{
+			"type": "permanent",
+			"reason": "DockerHung",
+			"pattern": "task docker:\\w+ blocked for more than \\w+ seconds\\."
```
dchen1107: We talked this offline. Along with above TaskHung, Docker Hung will be caught twice, and reported twice, one as an event, another one is condition. I think this should be ok.

Random-Liu: To fix this in the future, we could:
* Explicitly exclude overlapping check in regular expression.
* Define simple priority, may just high priority comes first, and only report the first matched rule.

node-problem-detector.yaml
```diff
@@ -0,0 +1,40 @@
...
+metadata:
+  name: node-problem-detector
+spec:
+  template:
+    metadata:
+      labels:
+        app: node-problem-detector
+    spec:
+      hostNetwork: true
+      containers:
```
dchen1107: Please measure resource overhead and update ContainerSpec with request / limit.

pkg/condition/manager.go
```diff
@@ -0,0 +1,134 @@
...
+	// condition for each condition type.
+	c.updates[condition.Type] = condition
+}
+
+func (c *conditionManager) syncLoop() {
+	updateCh := c.clock.Tick(updatePeriod)
+	for {
+		select {
+		case <-updateCh:
+			if c.checkUpdates() || c.checkResync() {
```
dchen1107: I might misread the code here. It looks like every 30 second (through resyncPeriod), problem_detector will update API server with conditions no matter there is change or not?

pkg/kernelmonitor/kernel_log_watcher.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	start, err := k.getStartPoint(path)
+	if err != nil {
+		return nil, err
+	}
+	// TODO(random-liu): If the file gets recreated during this interval, the logic
+	// will go wrong here.
+	// TODO(random-liu): Rate limit tail file.
+	// TODO(random-liu): Figure out what happens if log lines are removed.
+	k.tl, err = tail.TailFile(path, tail.Config{
```
dchen1107: What happens when the old log file is archived?

pkg/kernelmonitor/kernel_log_watcher.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// getStartPoint parses the newest kernel log file and try to find the latest reboot point.
+// Currently we rely on the kernel log timestamp to find the reboot point. The basic idea
+// is straight forward: In the whole lifecycle of a node, the kernel log timestamp should
+// always increase, only when it is reboot, the timestamp will decrease. We just parse the
+// log and find the latest timestamp decreasing, then it should be the latest reboot point.
+// TODO(random-liu): A drawback is that if the node is started long time ago, we'll only get
+// logs in the newest kernel log file. We may want to improve this in the future.
+func (k *kernelLogWatcher) getStartPoint(path string) (int64, error) {
```
dchen1107: Do we really need the latest reboot point? Or we can find from latest report timestamp?

config/kernel-monitor.json
```diff
@@ -0,0 +1,31 @@
...
+			"reason": "OOMKilling",
+			"pattern": "Kill process \\d+ (.+) score \\d+ or sacrifice child\\nKilled process \\d+ (.+) total-vm:\\d+kB, anon-rss:\\d+kB, file-rss:\\d+kB"
+		},
+		{
+			"type": "temporary",
+			"reason": "TaskHung",
+			"pattern": "task \\S+:\\w+ blocked for more than \\w+ seconds\\."
+		},
+		{
+			"type": "permanent",
```
dchen1107: Should all those permanent issues are introduced as NodeConditionTypes? I am ok to do that with following PRs later.

pkg/kernelmonitor/types/types.go
```diff
@@ -0,0 +1,45 @@
...
+	Message   string
+}
+
+// Type is the type of the kernel problem.
+type Type string
+
+const (
+	// Temp means the kernel problem is temporary, only need to report an event.
+	Temp Type = "temporary"
+	// Perm means the kernel problem is permanent, need to change the node condition.
```
dchen1107: cc/ @davidopp We are going to generate a corresponding NodeConditionType for each permanent issue detected by kernelmonitor. Are you ok with this?

dchen1107:
Looks good to me overall.

dchen1107:
@Random-Liu This pr is too large, so my comments are made on one commit. I don't expect you address all the comments above. Let's merge then start a separate PR to address all the issues, WDYT?

Random-Liu:
@dchen1107 Sure.

Makefile
```diff
@@ -0,0 +1,19 @@
+all: push
+
+# See pod.yaml for the version currently running-- bump this ahead before rebuilding!
+TAG = 0.1
+
+# TODO(random-liu): Change the project to google_containers.
+PROJ = google.com/noogler-kubernetes
```
dchen1107: cc/ @mikedanese what is the new policy for managing kubernetes images nowadays?

mikedanese: I don't think it has changed. Do you need instructions on how to push to gcr.io/google-containers

mikedanese: Maybe let @david-mcmahon know we are maintaining a new image there, but you should be set to push this to gcr.io/google-containers

Random-Liu: @mikedanese Thanks! ;)

PR #6, opened 2016-05-19
Change gcr project to official google_containers

Random-Liu:
This update the gcr project to the official one.
I've already upload an image `node-problem-detector:0.1` to google_containers, will send a PR to kubernetes to add it as an add-on pod soon.

@dchen1107


Random-Liu:
Manually merged.

Issue #7, opened 2016-05-23
Unable to build code, missing godep


derekwaynecarr:
I cloned the repo and ran:

```shell
$ make node-problem-detector
CGO_ENABLED=0 GOOS=linux godep go build -a -installsuffix cgo -ldflags '-w' -o node-problem-detector
vendor/gopkg.in/fsnotify.v1/inotify.go:19:2: cannot find package "golang.org/x/sys/unix" in any of:
	/home/decarr/go/src/k8s.io/node-problem-detector/vendor/golang.org/x/sys/unix (vendor tree)
	/usr/local/go/src/golang.org/x/sys/unix (from $GOROOT)
	/home/decarr/go/src/k8s.io/node-problem-detector/Godeps/_workspace/src/golang.org/x/sys/unix (from $GOPATH)
	/home/decarr/go/src/golang.org/x/sys/unix
godep: go exit status 1
Makefile:9: recipe for target 'node-problem-detector' failed
make: *** [node-problem-detector] Error 1
```

Looks like this is missing a godep in master:
https://github.com/kubernetes/node-problem-detector/tree/master/vendor

derekwaynecarr:
Tried using: 

```
$ go version
go version go1.6 linux/amd64
$ godep version
godep v70 (linux/amd64/go1.6)
```

Random-Liu:
@derekwaynecarr Let me see what happened. :)

eparis:
@freeformz    any guesses here?
```
$ godep restore
godep: Dep (gopkg.in/fsnotify.v1) restored, but was unable to load it with error:
	Package (golang.org/x/sys/unix) not found
godep: Error checking some deps.
```

Random-Liu:
Found out why

eparis:
Complete reproducer:
```
$ mkdir -p /tmp/gopath/src/k8s.io
$ cd /tmp/gopath/src/k8s.io/
$ git clone https://github.com/kubernetes/node-problem-detector.git
$ cd node-problem-detector
$ godep restore
godep: Dep (gopkg.in/fsnotify.v1) restored, but was unable to load it with error:
    Package (golang.org/x/sys/unix) not found
godep: Error checking some deps.
```

Random-Liu:
Should be fixed in #8  :)
@derekwaynecarr @eparis Thanks a lot for reporting this!

eparis:
More for @freeformz
```
$ godep restore -v
[...]

PR #10, opened 2016-05-26
Vendor the "right" fsnotify

freeformz:
This is the version that hpcloud/tail vendors


freeformz:
See #7

Random-Liu:
@freeformz IIUC, this should eliminate the dependency on `/sys/unix`, right? Could you also clean that? :) Thanks a lot!

freeformz:
@Random-Liu done.

Random-Liu:
@freeformz Quick question. I'm not quite familiar with Godeps. Could you explain the last 2 commits a bit? The first 2 LGTM.

Random-Liu:
All right, understood the 3rd one. https://github.com/tools/godep/blob/master/Changelog.md
Can you explain the last commit a bit?

freeformz:
Before v71 we didn't preserve permissions of the files copied in, now we do. This is particularly important for executable files.

The last one doesn't really matter that much, it looks like a more exact tag was added at some point and we're now picking it up.

Random-Liu:
@freeformz Thanks a lot! LGTM!

Issue #9, opened 2016-05-26
Node-Problem-Detector should Patch NodeStatus not Update 


saad-ali:
*Problem:*
`kubelet`, `node-controller`, and now the `node-problem-detector` all update the `Node.Status` field.

Normally this is not a problem because if changes happen rapidly, resource version mismatch fails and everything is ok.

When a new field is added to Node Status, however:
* Since `kubelet` and `node-controller` are in the same repository they are recompiled with the new field and Status updates continue to operate normally.
* However, since `node-problem-detector` is in a separate repository and has not been recompiled with the new version, its `Update` calls end up squashing the new (unknown) field, resetting it to `nil`.

*Suggested Solution:*
`node-problem-detector` should do a patch instead of an update to prevent wiping out fields it is not aware of. Incidentally, `kubelet` and `node-controller` should do this as well, but that is not as critical since they are in the same repository and new types are normally added their first.

CC @kubernetes/sig-node @kubernetes/sig-api-machinery @Random-Liu @bgrant0607

caesarxuchao:
You can use the [Patch()](https://github.com/kubernetes/kubernetes/blob/master/pkg/client/typed/dynamic/client.go#L185) method in the dynamic client, but you'll need to modify that function to allow specifying a subresource. cc @krousey

lavalamp:
I didn't know you guys were building this in a different repository. If you aren't going to be version-locked like the other components are, then you need some features that aren't built into our client.

Patch and/or dynamic client seems to be your best way forward.

There are various ways we could have made this easier, but they won't happen in time for your first release...

Random-Liu:
@saad-ali Thanks a lot for reporting!

> You can use the Patch() method in the dynamic client, but you'll need to modify that function to allow specifying a subresource. cc @krousey

I'll try this!

@lavalamp Thanks a lot! Patch is good enough for now~ Let me try :)

derekwaynecarr:
I think we should also have a node e2e that verifies the Kubelet itself
does not regress and drop node conditions it does not manage.  I don't
think we have anything that verifies that today and it would help avoid
regressions.

dchen1107:
Yes, I talked @Random-Liu on e2e test verifying this.

bgrant0607:
@saad-ali Ah, so it was getting clobbered. Great catch!  Where multiple components update a resource, they'll need a way to patch, either via PATCH or via a dedicated subresource. 

[...]

PR #11, opened 2016-05-31
Use patch instead of update status

Random-Liu:
Fix https://github.com/kubernetes/node-problem-detector/issues/9.

Change node-problem-detector to use `Patch` instead of `UpdateStatus`.

@caesarxuchao There is no good fake for [`unversioned.Client`](https://github.com/kubernetes/kubernetes/blob/master/pkg/client/unversioned/client.go#L118), so I can't write proper test for `unversioned.Client` related operations. [`testclient.Fake`](https://github.com/kubernetes/kubernetes/blob/master/pkg/client/unversioned/testclient/testclient.go#L56) is a fake for [`unversioned.Interface`](https://github.com/kubernetes/kubernetes/blob/master/pkg/client/unversioned/client.go#L30), but there is no `Patch` in `unversioned.Interface`, so I have to use `unversioned.Client`. Is there any way to test `unversioned.Client`?

@dchen1107 @saad-ali


caesarxuchao:
> Is there any way to test unversioned.Client?

No, sorry. That package is being deprecated. I opened an issue for adding `Patch` in the generated clientsets (https://github.com/kubernetes/kubernetes/issues/26580), but probably I'll do it after 1.3.

Random-Liu:
@caesarxuchao Cool thanks! :)

dchen1107:
LGTM

PR #12, opened 2016-06-01
Make condition and source configurable

Random-Liu:
Based on #11, only the last commit is new.

This PR:
1. makes node condition and source configurable.
2. adds multiple events and conditions support in problem interface.

We make this change mainly for extensibility, testability and performance.
* Testability: The PR makes it possible to write e2e test which generates test only conditions and events without affecting real node problem detector.
* Extensibility: The PR makes it possible to extend condition types (we only support `KernelDeadlock` before) by just updating configuration.
* Performance: The PR makes it possible to batch reporting events and condition updates.

@dchen1107 
/cc @kubernetes/sig-node

### DIFF
--- a/config/kernel-monitor.json
+++ b/config/kernel-monitor.json
@@ -1,6 +1,14 @@
 {
 	"logPath": "/log/kern.log",
 	"bufferSize": 10,
+	"source": "kernel-monitor",
+	"conditions": [
+		{
+			"type": "KernelDeadlock",
+			"reason": "KernelHasNoDeadlock",
+			"message": "kernel has no deadlock"
+		}
+	],
 	"rules": [
 		{
 			"type": "temporary",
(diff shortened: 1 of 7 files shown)


dchen1107:
Since #11 was merged, can you rebase first?

Random-Liu:
@dchen1107 Done.

config/kernel-monitor.json
```diff
@@ -14,16 +22,19 @@
...
 		{
 			"type": "permanent",
+			"condition": "KernelDeadlock",
 			"reason": "DockerHung",
 			"pattern": "task docker:\\w+ blocked for more than \\w+ seconds\\."
 		},
 		{
 			"type": "permanent",
+			"condition": "KernelDeadlock",
 			"reason": "KernelBug",
```
dchen1107: KernelBug is too generic here since above two can be in this category too. Can we change this to UnregisterNetDeviceIssue?

dchen1107:
Beside one nit, LGTM

config/kernel-monitor.json
Random-Liu: ACK :)

Random-Liu:
@dchen1107 Done. :)

dchen1107:
LGTM

Issue #3, opened 2016-05-17
KernelMonitor: Reconsider whether to use hpcloud/tail


Random-Liu:
For move fast as soon as possible, we used https://github.com/hpcloud/tail for the first version.
However, in fact it has been proved to have some bugs, such as https://github.com/hpcloud/tail/issues/21.

We should reconsider whether we should try to fix the bugs or implement a simple tail lib ourselves based on https://github.com/google/cadvisor/pull/1264.

AdoHe:
@Random-Liu what do you think of this issue now? When I am adding journald support, I want to use log reader like cadvisor, but the tail library we use currently doesn't implement `io.Reader`, anything block use our simple tail lib?

Random-Liu:
@AdoHe, as far as I know, we can just use the simple tail lib https://github.com/google/cadvisor/blob/master/utils/tail/tail.go.

Random-Liu:
@AdoHe One thing is that, we run 2 functions when parsing kernel log:
* `getStartPoint`: This function was initially added to find the recent reboot line in the kernel log.
* `watchLoop`: After finding the reboot line, in side the `watchLoop`, `tail.Tail` is called with `start=head of the reboot line` to watch and parse the following log.

In this implementation, the tail lib needs to support watch from a specific point in the file, which is not supported in the simple tail library.

However, the logic was recently changed. We find start point based on timestamp ([`now-lookback`](https://github.com/kubernetes/node-problem-detector/blob/master/pkg/kernelmonitor/kernel_log_watcher.go#L166-L188)). In fact we can remove the function `getStartPoint` and combine the logic into watchLoop. When receiving a new log line:
* If it is older than `now-lookback`, discard it.
* If not, return translate and add the log line.

AdoHe:
@Random-Liu Thanks very much.

Random-Liu:
This should be closed after #39 got merged.

PR #13, opened 2016-06-03
Hack for unsupported OS distros.

Random-Liu:
This is a hack for unsupported OS distros.
KernelMonitor doesn't support some OS distros for now e.g. GCI. Ideally, we should only run KernelMonitor on nodes on supported OS distro. However, NodeProblemDetector is running as DaemonSet, it has to be deployed on each node:
  * There is no node affinity support for DaemonSet now https://github.com/kubernetes/kubernetes/issues/22205
  * There is no node label for OS Distro, so we can't use node selector, either

If some nodes have unsupported OS distro e.g. the OS distro of master node in gke/gce is GCI, KernelMonitor will throw out error, and NodeProblemDetector will be restarted again and again because it's a DaemonSet.

To avoid this, we decide to add this temporarily hack. When KernelMonitor can't find the kernel log file, it will print a log and then return nil channel and no error. Since nil channel will always be blocked, the NodeProblemDetector will block forever.

I've tried to schedule the NodeProblemDetector on GCI image, the cpu usage is `0.01% core` and memory usage is `< 5MB`.

TODOs:
1. Add journald support to support GCI. Consider reusing [the code in cadvisor](https://github.com/google/cadvisor/blob/master/utils/oomparser/oomparser.go#L168), and maybe replace hpcloud/tail with the simpler implementation in cadvisor https://github.com/kubernetes/node-problem-detector/issues/3
2. Schedule KernelMonitor only on nodes with supported OS distros. (with node label+selector or affinity)

@dchen1107 
/cc @kubernetes/sig-node

### DIFF
--- a/pkg/kernelmonitor/kernel_log_watcher.go
+++ b/pkg/kernelmonitor/kernel_log_watcher.go
@@ -71,6 +71,22 @@ func (k *kernelLogWatcher) Watch() (<-chan *types.KernelLog, error) {
 	if k.cfg.KernelLogPath != "" {
 		path = k.cfg.KernelLogPath
 	}
+	// NOTE(random-liu): This is a hack. KernelMonitor doesn't support some OS distros e.g. GCI. Ideally,
+	// KernelMonitor should only run on nodes with supported OS distro. However, NodeProblemDetector is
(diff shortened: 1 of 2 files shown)


dchen1107:
LGTM

PR #15, opened 2016-06-03
Add entrypoint and copy kernel monitor config into docker image

Random-Liu:
Add entrypoint and config into docker image, so as to simplify node problem detector deployment in Kubernetes.

Addon manager doesn't support ConfigMap now. We could change addon manager to support ConfigMap and also add ConfigMap in addon manifests, but that is not necessary and we should not make that big change before cutting release.
Even though we added entrypoint and config into the docker image, we can always overwrite them in the yaml file in the future.


dchen1107:
LGTM

PR #18, opened 2016-06-09
Remove unnecessary ENV configuration.

Random-Liu:
Thanks for @mikedanese's suggestion.

This PR removed the manually configuration of apiserver ip and port.

Double checked with @freehan. The pod in host network should be able to use the cluster ip in the ENV value to access apiserver directly.

I've no idea what was wrong when I tried that before :(. It didn't work for me at that time...

@dchen1107

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,12 +71,6 @@ spec:
       - name: node-problem-detector
         image: gcr.io/google_containers/node-problem-detector:v0.1
         imagePullPolicy: Always
-        env:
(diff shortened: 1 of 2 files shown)


mikedanese:
LGTM thanks

Random-Liu:
Apply LGTM based on https://github.com/kubernetes/node-problem-detector/pull/18#issuecomment-225052862

PR #22, opened 2016-06-24
Kernel Monitor: Add look back support and kernel panic handling

Random-Liu:
This helps https://github.com/kubernetes/kubernetes/issues/27885.

This PR:
1) Add lookback support in kernel monitor. After started, Kernel monitor will check some old logs to detect old problems which happened before last node reboot.

2) Add `lookback` and `startPattern` in kernel monitor configuration.
  * `lookback` specifies how long time kernel monitor should look back.
  * `startPattern` specifies which log indicates the node is started. kernel monitor will clear all current node conditions once it finds a node start log. This makes sure that old problems won't change the node condition.

3) Add support for kernel panic monitoring, the null pointer and divide 0 kernel panic will be surfaced as event. Usually kernel monitor will report these events during looking back phase.

I've cut a branch v0.1 before this PR, and bump up the image version to v0.2 in this PR.
@dchen1107 
/cc @kubernetes/sig-node

### DIFF
--- a/config/kernel-monitor.json
+++ b/config/kernel-monitor.json
@@ -1,5 +1,7 @@
 {
 	"logPath": "/log/kern.log",
+	"lookback": "10m",
+	"startPattern": "Initializing cgroup subsys cpuset",
 	"bufferSize": 10,
 	"source": "kernel-monitor",
 	"conditions": [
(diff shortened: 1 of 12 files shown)


Random-Liu:
I've updated and verified the PR. With this PR, if there was a kernel panic in last run, node problem detector will report a node event similar with:
```
  10s	10s	1	{kernel-monitor e2e-test-lantaol-minion-group-29qf}		Warning	KernelPanic	divide error: 0000 [#1] SMP
```

config/kernel-monitor.json
```diff
@@ -1,5 +1,7 @@
 {
 	"logPath": "/log/kern.log",
+	"lookback": "1h",
```
dchen1107: looking back 1h might be too long to find panic information? We only care about the most recent, right? Changing it to 5m, or 10m maybe?

Random-Liu:
@dchen1107 Addressed comments.

dchen1107:
LGTM

Issue #28, opened 2016-08-09
NodeProblemDetector tests failing on AWS


zmerlynn:
https://github.com/kubernetes/node-problem-detector/blob/9ab546a2e23613c51e51af7a109952a6079e0651/pkg/problemclient/problem_client.go#L61 assumes that the hostname is the same name as the cloud provider instance name. This is false on AWS. NPD needs to be going through `CurrentNodeName` to get the cloud provider agnostic hostname.

I'm disabling this test on AWS for now, but we're marching towards full support for AWS. Please treat this as if it had broken a build when it was checked in originally.

zmerlynn:
cc @justinsb

zmerlynn:
I'm not sure I agree that this is an `enhancement`, unless the original design was for GCE images/kernels only.

Random-Liu:
@zmerlynn Ha, no other labels in this repo except bug. :)
The best way for now should be using the downward api https://github.com/kubernetes/kubernetes/pull/27880, which is mentioned in the issue https://github.com/kubernetes/node-problem-detector/issues/25.

I'll ask whether we can get https://github.com/kubernetes/kubernetes/pull/27880 in any time soon. If so, we can switch to using it quite soon, if not, I'll think some other ways.

zmerlynn:
I forgot to reopen the bug, but the NPD tests never actually started passing on AWS. I'll look into it soon.

Random-Liu:
This should be solved once https://github.com/kubernetes/kubernetes/pull/35740 is merged.

@dchen1107

Random-Liu:
Close this issue because kubernetes/kubernetes#35740 has been merged. Feel free to reopen if the issue happens again.

PR #29, opened 2016-08-11
NPD: Get node name from pod

Random-Liu:
Fixes https://github.com/kubernetes/node-problem-detector/issues/28.
Suggested in https://github.com/kubernetes/kubernetes/pull/27880#issuecomment-238776172.

This PR makes node problem detector to get node name from the pod where it is on, so that the node name should always be consistent with kubelet.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 all: push
 
 # See pod.yaml for the version currently running-- bump this ahead before rebuilding!
-TAG = v0.1
+TAG = v0.2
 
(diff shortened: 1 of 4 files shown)


dchen1107:
LGTM

Random-Liu:
@dchen1107 Thanks for reviewing. :)

Random-Liu:
Removed the hostnetwork, because we don't need it anymore, now we are getting host name from the apiserver.

PR #31, opened 2016-09-10
NPD: Enable travis test

Random-Liu:
For lack of continuous integration test, there are some errors introduced in the code base.
This PR:
* fixed a failing unit test
* fix a missing godep
* add travis ci

PR #32, opened 2016-09-12
NPD: Remove get start point

Random-Liu:
Based on https://github.com/kubernetes/node-problem-detector/pull/31 to enable the presubmit test.

This PR addressed https://github.com/kubernetes/node-problem-detector/issues/3#issuecomment-246431485 to remove the unnecessary `getStartPoint` function.

With this change @AdoHe should be able to use simple tail library directly.

### DIFF
--- a/pkg/kernelmonitor/kernel_log_watcher.go
+++ b/pkg/kernelmonitor/kernel_log_watcher.go
@@ -17,8 +17,6 @@ limitations under the License.
 package kernelmonitor
 
 import (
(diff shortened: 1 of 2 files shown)


dchen1107:
rebase first, then we are ready to go.

Random-Liu:
@dchen1107 Rebased, apply LGTM based on https://github.com/kubernetes/node-problem-detector/pull/32#issuecomment-246492744.

PR #36, opened 2016-10-27
Share time with the host

Random-Liu:
Kernel monitor discard old kernel log based on the timestamp.

However, usually kernel log timestamp doesn't contain year and timezone information, e.g. `Oct 23 16:59:23`. There is no way to apply the right timezone, if the timezone inside the pod is different from the host.

This PR mount the `/etc/localtime` inside the container to make sure the container in the same zone with the host.

Sometimes `/etc/localtime` may be a symlink file. I've tested that both docker and kubernetes mount will follow the symlink and mount the real content.

@dchen1107

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,11 +81,19 @@ spec:
         - name: log
           mountPath: /log
           readOnly: true
+        - name: localtime
+          mountPath: /etc/localtime
+          readOnly: true
       volumes:
       - name: log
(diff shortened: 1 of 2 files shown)


dchen1107:
I am closing this one since https://github.com/kubernetes/node-problem-detector/pull/39, which covers this issue, was merged. Please reopen it if I were wrong. Thanks!

PR #38, opened 2016-11-10
modify checks to enable journald support

roboll:
No longer defaults kernel log file when it is empty.

Currently, the journald support will never trigger because an empty path is always defaulted [here](https://github.com/kubernetes/node-problem-detector/blob/master/pkg/kernelmonitor/kernel_log_watcher.go#L84), but journald support checks for empty path [here](https://github.com/kubernetes/node-problem-detector/blob/master/pkg/kernelmonitor/kernel_log_watcher.go#L171).

The default config ships with a path in it, so this change shouldn't disrupt existing deployments. When the path is empty, will try journal. When the path is non-empty, existing behavior is intact.

### DIFF
--- a/pkg/kernelmonitor/kernel_log_watcher.go
+++ b/pkg/kernelmonitor/kernel_log_watcher.go
@@ -35,10 +35,6 @@ import (
 	utilclock "github.com/pivotal-golang/clock"
 )
 
-const (
-	defaultKernelLogPath = "/var/log/kern.log"
-)
-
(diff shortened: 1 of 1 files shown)


Random-Liu:
@roboll Thanks a lot for the fix.

https://github.com/kubernetes/node-problem-detector/pull/39 also addressed this issue and some other issues together. :)

roboll:
@Random-Liu got it. closing this in favor of #39

Random-Liu:
@roboll Thanks! :)

PR #39, opened 2016-11-10
Journald support

### LINKED ISSUE #3: KernelMonitor: Reconsider whether to use hpcloud/tail

### LINKED ISSUE #14: KernelMonitor: Add journald support
Add journald support.

Consider reusing [the code in cadvisor](https://github.com/google/cadvisor/blob/master/utils/oomparser/oomparser.go#L168), and maybe replace hpcloud/tail with the simpler implementation in cadvisor https://github.com/kubernetes/node-problem-detector/issues/3

Random-Liu:
Fixes #3.
Fixes #14.
This PR depends on https://github.com/google/cadvisor/pull/1531. I'll remove the 3rd commit and update Godeps after the cadvisor patch is merged.

This PR finished the journald support.

Building `sdjournal` needs `libsystemd-dev` or `libsystemd-journal-dev` which may not be easy to get on some os distros. So I added a knob to disable it:
```
make # wil build journald support by default
ENABLE_JOURNALD=0 make # will not build journald support, thus no libsystemd needed
```

@AdoHe @dchen1107 
/cc @kubernetes/sig-node

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/node-problem-detector/39)


Random-Liu:
@AdoHe Could you help me review this? :)

Random-Liu:
@AdoHe Wait, there is still something wrong. It seems that sdjournal needs `libsystemd.so`, which is really hard to deal with...

AdoHe:
@Random-Liu this [cadvisor#1531](https://github.com/google/cadvisor/pull/1531) has merged, please remove the third commit and update Godep.

pkg/kernelmonitor/kernel_log_reader.go
```diff
@@ -0,0 +1,64 @@
...
+type readerGetter func(WatcherConfig) (io.ReadCloser, error)
+
+// logReaderGetters are the available reader getters.
+var logReaderGetters = []readerGetter{getFileLogReader}
+
+// getLogReader gets select a kernel log reader through the reader getters.
+func getLogReader(cfg WatcherConfig) (io.ReadCloser, error) {
+	var errs []error
+	for name, getter := range logReaderGetters {
+		r, err := getter(cfg)
```
AdoHe: what's the meaning of name here?

AdoHe:
LGTM. @Random-Liu 👍

pkg/kernelmonitor/kernel_log_reader.go
Random-Liu: Will do!

Random-Liu:
@AdoHe In fact there is a problem with current solution. The `go-systemd/sdjournal` replies on the `libsystemd.so` https://github.com/kubernetes/node-problem-detector/blob/master/vendor/github.com/coreos/go-systemd/sdjournal/functions.go#L32-L37, which I think is on the node with systemd installed.

There are several ways to solve this:
1. Mount the host path to make sure we can find the shared library with `dlopen` inside the container. => The shared library may be in different paths on different os distros.
2. Add the shared library inside the container. => We are not sure whether the library will be compatible with systemd on host, at least for systemd < 209 and systemd >= 209, they are different.
3. Use and mount journalctl binary instead.
  a. Mount all necessary path and binary into the container. => I've tried this, and couldn't make it work till now.
  b. just mount the whole root fs as readonly inside the container, and run kernel monitor in a chroot. => I've tried this, and this works fine.

euank:
@Random-Liu I have another option that solves that problem: Don't use journald! See #41 for that solution :grinning:

AdoHe:
@euank as kernel monitor just cares about the kernel log message, access kmsg directly sounds like a good option.

jfilak:
@Random-Liu What are the problems with **Option 2**? ABRT watches for kernel oopses in journald too. The ABRT container ships with own libsystemd and we haven't experienced any problem with backward compatibility.

**Option 3** works for me on `Fedora 24 (Atomic Host)` with `systemd-229-16.fc24.x86_64` and `docker-1.10.3-54.gite03ddb8.fc24.x86_64`:

```bash
docker run -it --privileged --rm -v /:/host fedora sh
journalctl -f -D /host/var/log/journal
```

Random-Liu:
> Option 3 works for me on Fedora 24 (Atomic Host) with systemd-229-16.fc24.x86_64 and docker-1.10.3-54.gite03ddb8.fc24.x86_64

@jfilak Thanks for your information! I'm also using fedora now, because it has newer systemd (229) and supports `+LZ` which is at least needed in GCI.

I'll send an update soon.

Random-Liu:
@AdoHe @euank @jfilak Updated the PR.

* The 1st commit added `Plugin` interface, and implemented `syslog` and `journald` plugin.
* The 2nd commit updated `Makefile` to compile and build container image. Note that:
  * We added a build flag `journald` to enable and disable journald support. We do this because:
    * Compile `go-systemd/sdjournal` needs `libsystemd-dev` or `libsystemd-journal-dev`.
    * Container image become much bigger to support journald, because of the different base image and extra dependencies.
  * We are using fedora as the base image now, because it has newer systemd (229) and supports +LZ which is at least required by GCI.
* The 3rd commit updated the Godeps to include https://github.com/google/cadvisor/pull/1531.
  * Note that cadvisor is using different `go-systemd` version. It confused godeps. This PR has already included all dependencies, but in the future we may need to run `godep save ./...` twice when updating godeps. (Same with https://github.com/kubernetes/node-problem-detector/issues/7)
* The 4th commit updated the `Readme.md` to reflect the new change in kernelmonitor.

@dchen1107 
/cc @kubernetes/sig-node

Makefile
```diff
@@ -1,18 +1,56 @@
...
+
+SOURCES:=$(shell find . -name '*.go')
+
+node-problem-detector: $(SOURCES)
+	CGO_ENABLED=$(CGO_ENABLED) GOOS=linux go build -o node-problem-detector -a -ldflags '-w' $(BUILD_TAGS)
 
 container: node-problem-detector
+	cp Dockerfile.in Dockerfile
+	sed -i 's|BASEIMAGE|$(BASEIMAGE)|g' Dockerfile
+	sed -i 's|SETUP|$(SETUP)|g' Dockerfile
```
jfilak: Nitpicking: I would rather use @BASEIMAGE@ and @SETUP@ (Autotools dialect) in the Dockerfile file, so it would be much clearer that these strings will be replaced. And I usually create a solo rule for the file transformations as follows:
```
Dockerfile: Dockerfile.in
	sed -e 's|BASEIMAGE|$(BASEIMAGE)|g' \
	    -e 's|SETUP|$(SETUP)|g' \
	    $^ >$@
```

What is supposed to be the value of the SETUP variable?

Makefile
```diff
@@ -1,18 +1,56 @@
...
 	gcloud docker push gcr.io/$(PROJ)/node-problem-detector:$(TAG)
 
 clean:
+	rm -f Dockerfile
 	rm -f node-problem-detector
+
+test:
+	go test -v -race ./pkg/... $(BUILD_TAGS)
+
+.PHONY: all
```
jfilak: Just a question: Why do you make only the all target a PHONY target? I believe that the other targets that should run unconditionally should be made PHONY targets too?

Makefile
```diff
@@ -1,18 +1,56 @@
...
+
+# TODO(random-liu): Support different architectures.
+BASEIMAGE:=alpine:3.1
+
+# Disable cgo by default to make the binary statically linked.
+CGO_ENABLED:=0
+
+# NOTE that enable journald will increase the image size.
+ifeq ($(ENABLE_JOURNALD), 1)
+	# Enable journald bulid tag.
```
Random-Liu: Will do. Thanks!

Makefile
Random-Liu: Good suggestion. Will do!

In fact `SETUP` can be removed. Initially I planed to use ubuntu, and needs a `SETUP` to download the libsystemd-dev. However, later I decided to use fedora because systemd on ubuntu doesn't support `+LZ`. Because systemd is already in fedora image, an extra `SETUP` is not needed.

Makefile
Random-Liu: Because pushing to `gcr.io/google_containers` needs some special authentication. I don't think we want to do that every time we do `make`. :)

euank: `PHONY` isn't about whether it happens every time, just about whether it treats the target name as a file to check the existence of (e.g. if you do `touch clean; make clean` it won't run the commands because it treats "clean" as a filename and sees that the target already exists).

`node-problem-detector` isn't a PHONY necessarily, but everything else should be.

Random-Liu: @euank Thanks for the clarification. I'm really a Makefile fresher. :)

Will review your PR and rebase this one.

Random-Liu:
Addressed comments. :)

Makefile
```diff
@@ -1,27 +1,60 @@
...
+	CGO_ENABLED:=1
+endif
 
 PKG_SOURCES := $(shell find pkg -name '*.go')
 
 node-problem-detector: ./bin/node-problem-detector
 
 ./bin/node-problem-detector: $(PKG_SOURCES) node_problem_detector.go
-	CGO_ENABLED=0 GOOS=linux godep go build -a -installsuffix cgo -ldflags '-w' -o node-problem-detector
+	CGO_ENABLED=$(CGO_ENABLED) GOOS=linux go build -o ./bin/node-problem-detector -a -ldflags '-w' $(BUILD_TAGS)
```
euank: Any reason we ever want CGO_ENABLED=0? Only distribution we support is the container, and it'll work fine in the container either way.

Makefile
```diff
@@ -1,27 +1,60 @@
...
 	gcloud docker push gcr.io/$(PROJ)/node-problem-detector:$(TAG)
 
 clean:
+	rm -f Dockerfile
 	rm -f ./bin/node-problem-detector
+
+test:
+	go test -timeout=1m -v -race ./pkg/... $(BUILD_TAGS)
+
+.PHONY: all
```
euank: nit: shoulda been rebased out

Makefile
```diff
@@ -1,27 +1,60 @@
...
 
 node-problem-detector: ./bin/node-problem-detector
 
 ./bin/node-problem-detector: $(PKG_SOURCES) node_problem_detector.go
-	CGO_ENABLED=0 GOOS=linux godep go build -a -installsuffix cgo -ldflags '-w' -o node-problem-detector
+	CGO_ENABLED=$(CGO_ENABLED) GOOS=linux go build -o ./bin/node-problem-detector -a -ldflags '-w' $(BUILD_TAGS)
 
-test:
-	go test -timeout=1m -v -race ./pkg/...
+Dockerfile: Dockerfile.in
```
euank: Should there be a `.gitignore` for Dockerfile?

pkg/kernelmonitor/kernel_log_watcher.go
```diff
@@ -135,79 +112,26 @@ func (k *kernelLogWatcher) watchLoop() {
...
+		buffer.WriteString(line)
 		if err == io.EOF {
-			buffer.WriteString(line)
-			time.Sleep(100 * time.Millisecond)
+			time.Sleep(watchPollInterval)
 			continue
 		}
-		if line == "" {
-			time.Sleep(100 * time.Millisecond)
+		// Trim tailing `\n`.
```
euank: redundant comment imo.

Also, `strings.TrimSpace` would be fine I expect

pkg/kernelmonitor/plugins/types/plugin.go
```diff
@@ -0,0 +1,45 @@
...
+package types
+
+import (
+	"io"
+
+	"k8s.io/node-problem-detector/pkg/kernelmonitor/types"
+)
+
+// Plugin is the interface of a log parsing plugin.
+type Plugin interface {
```
euank: I think an interface that is just: `StreamLogMessages() (<-chan *types.KernelMessage, error)` would be better.

As is, this interface won't work for a kmsg plugin due to the reading abstraction. A `read` of the kmsg device has special semantics which are lost by the attempts to buffer and so on above.

Makefile
Random-Liu: Will do.

Makefile
Random-Liu: My bad. :(

pkg/kernelmonitor/plugins/types/plugin.go
Random-Liu: The advantage of current interface is:
* Less code change, because current code is working like this.
* Plugin can share some code, such as the read loop.

However, delegating the whole logic to the plugin does seem to be more flexible to me:
* We can support kmsg, which has [different read logic](https://github.com/euank/node-problem-detector/blob/d834639cf9181850a408f6adde3c9154d4b9fb25/vendor/github.com/euank/go-kmsg-parser/kmsgparser/kmsgparser.go#L115-L130).
* We can use sdjournal api directly instead of the reader interface which has an extra buffer inside.

Comparing these 2 options, I prefer the later. :)

/cc @AdoHe WDYT?

Makefile
jfilak: If you don't add Dockerfile to PHONY targets, users will be confused. Consider this scenario:
```bash
make
# The Dockerfile file didn't exist and make created it.
# @BASEIMAGE@ is Fedora and systemd-journal is enabled.
make ENABLE_JOURNALD=0
# The Dockerfile file already exists and make skipped the Dockerfile target. 
# However the user believes that @BASEIMAGE@ is set to "alpine:3.4" and
# systemd-journal is disabled.
```

Random-Liu: @jfilak I see. You are right. Thanks for explaining. :)

Makefile
Random-Liu: If the binary is dynamically linked, and there is no corresponding dynamic library in the container, it won't work, right?

pkg/kernelmonitor/plugins/types/plugin.go
AdoHe: I prefer the later too. kmsg has different read logical, which could be regarded as a `new` kernel log watcher, but not suitable for current buffer approach plugin.

pkg/kernelmonitor/kernel_log_watcher.go
Random-Liu: Done

Makefile
Random-Liu: Done

Random-Liu:
I updated the PR:
1. Move the interface to higher level based on: https://github.com/kubernetes/node-problem-detector/pull/39#discussion_r89027913. This makes it possible to use `kmsg` directly https://github.com/kubernetes/node-problem-detector/pull/41.
2. Change the journald integration to use `Next()`, `GetEntry()` and `Wait()` directly.
3. Addressed above comments.

I think the PR is ready to go.

/cc @euank @jfilak @AdoHe

node-problem-detector.yaml
```diff
@@ -26,6 +26,11 @@ spec:
         - name: log
           mountPath: /log
           readOnly: true
+        # Make sure node problem detector is in the same timezone
+        # with the host.
+        - name: localtime
+          mountPath: /etc/localtime
+          readOnly: true
```
jfilak: FYI yesterday, Dan Walsh posted this on Atomic devel list (Atomic is basically Fedora):
https://lists.projectatomic.io/projectatomic-archives/atomic-devel/2016-December/msg00014.html

In short, if the `/etc/localtime` files is are symbolic link, then you end up with the localtime in the container not updated.

However, I don't think you can do something to avoid this issue.

Random-Liu: That's wired. When I initially decided to go with this approach, I remembered I've tried the symlink situation, it worked.

I'll try again then.

AdoHe:
Finally ready to go @Random-Liu 👍

pkg/kernelmonitor/logwatchers/journald/log_watcher.go
```diff
@@ -0,0 +1,163 @@
...
+		return nil, fmt.Errorf("failed to create journal client from path %q: %v", path, err)
+	}
+	// Seek journal client based on the lookback duration.
+	start := time.Now().Add(-since)
+	err = journal.SeekRealtimeUsec(uint64(start.UnixNano() / 1000))
+	if err != nil {
+		return nil, fmt.Errorf("failed to lookback %q: %v", since, err)
+	}
+	// TODO(random-liu): Make this configurable to support parsing other logs.
+	kernelMatch := sdjournal.Match{
```
AdoHe: should this be configurable? The `TRANSPORT` could be other value?

Random-Liu: Some people may want to use this to monitor other logs (e.g. docker log) to catch some known docker issue. (See https://github.com/kubernetes/node-problem-detector/issues/44_

pkg/kernelmonitor/logwatchers/syslog/helpers_test.go
```diff
@@ -49,15 +54,13 @@ func TestDefaultTranslator(t *testing.T) {
...
-				t.Errorf("case %d: expect error should occur, got %+v, %v", c+1, log, err)
-			}
+		log, err := translate(test.input)
+		if (err != nil) != test.err {
+			t.Errorf("case %d: error assertion failed, got log: %+v, error: %v", c+1, log, err)
 			continue
 		}
-		if test.timestamp != log.Timestamp || test.message != log.Message {
-			t.Errorf("case %d: expect %v, %q; got %v, %q", c+1, test.timestamp, test.message, log.Timestamp, log.Message)
+		if !reflect.DeepEqual(test.log, log) {
```
euank: We should really be using testify's `assert.Equal`

pkg/kernelmonitor/logwatchers/syslog/log_watcher.go
```diff
@@ -0,0 +1,123 @@
...
+	var buffer bytes.Buffer
+	for {
+		select {
+		case <-s.tomb.Stopping():
+			glog.Infof("Stop watching syslog")
+			return
+		default:
+		}
+
+		line, err := s.reader.ReadString('\n')
```
euank: You can get hung here while `s.tomb.Stopping` is true.

To fix this, you could have a watcher for tomb in another goroutine that closes s.reader when that happens so this no longer "hangs", and then the two goroutines could synchronize and exit.

What I wrote for the kmsg also does the right thing (see https://github.com/kubernetes/node-problem-detector/blob/d834639cf9181850a408f6adde3c9154d4b9fb25/pkg/kernelmonitor/kernel_log_watcher.go#L114-L135)

pkg/kernelmonitor/logwatchers/journald/log_watcher_test.go
```diff
@@ -0,0 +1,66 @@
...
+			log: &kerntypes.KernelLog{
+				Timestamp: time.Unix(0, 987654321*1000),
+				Message:   "",
+			},
+		},
+	}
+
+	for c, test := range testCases {
+		log := translate(test.entry)
+		if !reflect.DeepEqual(test.log, log) {
```
euank: assert would be nice.

pkg/kernelmonitor/logwatchers/journald/log_watcher.go
```diff
@@ -0,0 +1,163 @@
...
+	}()
+	for {
+		select {
+		case <-j.tomb.Stopping():
+			glog.Infof("Stop watching journald")
+			return
+		default:
+		}
+		// Get next log entry.
+		n, err := j.journal.Next()
```
euank: Same comment as the syslog one, you could get hung here while `tomb.Stopping` is trying to signal. Same fix as well.

pkg/kernelmonitor/util/utils.go
```diff
@@ -0,0 +1,29 @@
...
+limitations under the License.
+*/
+
+package util
+
+import (
+	"time"
+)
+
+func ParseDuration(s string) (time.Duration, error) {
```
euank: Honestly, I don't like this helper. It's saving the caller 3 lines of code, and that code is more obvious in the parent context (of defaulting a config) than here.

Renaming this to "ParseDurationWithDefault" or something could make it better, but I'd rather just get rid of this function.

Random-Liu: Thanks. I decided to add a function to apply default on the configuration passed in. Then we don't need this function.

euank:
@Random-Liu I gave a suggestion; using the kernel boot id (as provided in journald) to accurately determine whether it's relevant for the current boot, not parsing for strings that happen to exist on your box.

Speaking of your box, grepping for that log message does not work on CoreOS or other distros with a modern kernel. As of 4.5, the message being searched for [is a pr_debug](https://github.com/torvalds/linux/commit/a5ae989957cbc3f3b7bc40677bd9e459c0917528) message and therefore quite unlikely to appear.

Perhaps that's the reason my CoreOS nodes get marked as kernel unhealthy even after a reboot; that "start" message certainly never appears.

Random-Liu:
@euank Good to know. GCI is using 4.4.21+, no wonder it works on GCI.

> using the kernel boot id (as provided in journald) to accurately determine whether it's relevant for the current boot, not parsing for strings that happen to exist on your box.

However, it's still hard to correlate boot id with the corresponding kernel log. It's easy for journald. But for syslog, I need to investigate more to see whether there is a way. :)

euank:
@Random-Liu let's take it to #48, I suggested something there for syslog as well.

pkg/kernelmonitor/logwatchers/syslog/helpers_test.go
Random-Liu: Done.

pkg/kernelmonitor/logwatchers/journald/log_watcher_test.go
Random-Liu: Done.

pkg/kernelmonitor/logwatchers/syslog/log_watcher.go
Random-Liu: On a second thought, I feel like this is working as intended.

1) We should not read/close the file concurrently, or else there may be a race https://github.com/golang/go/issues/10001.
2) Stop function is added to make sure all resources being reclaimed properly, it should wait for files to be closed properly. So we have to wait for the last read, if there is one on going.

Random-Liu:
@euank Addressed comments.

node-problem-detector.yaml
Random-Liu: I read your link and understand what you mean. Thanks for sharing!

IIUC, it means that if `/etc/localtime` inside the container is a symlink (fedora is just using symlink), say `/etc/localtime -> ../usr/share/zoneinfo/UTC`. `../usr/share/zoneinfo/UTC` will be mounted instead.

This should not cause trouble to [golang](https://github.com/golang/go/blob/964639cc338db650ccadeafb7424bc8ebb2c0f6c/src/time/zoneinfo_unix.go#L59), but it's worth fixing to avoid confusion.

Random-Liu: Done.

pkg/kernelmonitor/logwatchers/journald/log_watcher.go
euank: The 5 second `waitLogTimeout` here means that it won't be blocked for too long at least, so this one isn't too bad.

pkg/kernelmonitor/logwatchers/syslog/log_watcher.go
```diff
@@ -0,0 +1,123 @@
...
+		buffer.WriteString(line)
+		if err == io.EOF {
+			time.Sleep(watchPollInterval)
+			continue
+		}
+		line = strings.TrimSpace(buffer.String())
+		buffer.Reset()
+		log, err := translate(line)
+		if err != nil {
+			glog.Infof("Unable to parse line: %q, %v", line, err)
```
euank: This should be at warning or error I think.

pkg/kernelmonitor/logwatchers/syslog/log_watcher.go
```diff
@@ -0,0 +1,123 @@
...
+		if err != nil && err != io.EOF {
+			glog.Errorf("Exiting syslog watch with error: %v", err)
+			return
+		}
+		buffer.WriteString(line)
+		if err == io.EOF {
+			time.Sleep(watchPollInterval)
+			continue
+		}
+		line = strings.TrimSpace(buffer.String())
```
euank: both parseLine and this do a `TrimSpace`. Just one of them should be fine.

euank:
Couple more comments, overall lgtm.

pkg/kernelmonitor/logwatchers/syslog/log_watcher.go
Random-Liu: Will do.

pkg/kernelmonitor/logwatchers/syslog/log_watcher.go
Random-Liu: Will do.

Random-Liu:
@euank Addressed comments.

Random-Liu:
Rebased the PR.

Thanks for the review @euank @jfilak @AdoHe.

@dchen1107 Would you like to take a final look?

pkg/kernelmonitor/README.md
```diff
@@ -43,14 +39,23 @@ with new rule definition:
...
+* journald: `logPath` is the journal log directory, usually `/var/log/journal`.
 
-## Support Other Log Format
+### New Log Plugin
 
-Kernel monitor uses [`Translator`](https://github.com/kubernetes/node-problem-detector/blob/master/pkg/kernelmonitor/translator/translator.go)
-plugin to translate kernel log the internal data structure. It is easy to
-implement a new translator for a new log format.
+Kernel monitor uses [Log
+Plugin](https://github.com/kubernetes/node-problem-detector/blob/master/pkg/kernelmonitor/plugins/types.go) to support different log management tools.
```
Random-Liu: The doc needs to be updated.

Makefile
```diff
@@ -1,24 +1,73 @@
...
 TARBALL := node-problem-detector-$(VERSION).tar.gz
 
-IMAGE := gcr.io/$(PROJ)/node-problem-detector:$(TAG)
+# IMAGE is the image name of the node problem detector container image.
+IMAGE := $(PROJ)/node-problem-detector:$(TAG)
+
+# ENABLE_JOURNALD enables build journald support or not. Building journald support needs libsystemd-dev
+# or libsystemd-journal-dev.
+# TODO(random-liu): Build NPD inside container.
+ENABLE_JOURNALD?=1
```
ankon: Nit: spaces around the assignment operators missing to keep this in-sync with the other variables here.

Makefile
```diff
@@ -30,14 +79,17 @@ version:
 	@echo $(VERSION)
 
 ./bin/node-problem-detector: $(PKG_SOURCES)
-	GOOS=linux go build -o bin/node-problem-detector \
+	CGO_ENABLED=$(CGO_ENABLED) GOOS=linux go build -o bin/node-problem-detector \
 	     -ldflags '-w -extldflags "-static" -X $(PKG)/pkg/version.version=$(VERSION)' \
```
ankon: As per the comment above the "-static" here looks fishy, and probably should only be here when not building with journald support?

andyxning: @ankon 

Yes, you're right. `-static` should only be used when journald is disabled. For more info please consult the comments about [this review](https://github.com/kubernetes/node-problem-detector/pull/54#pullrequestreview-14090197)

Random-Liu: @ankon Good catch. This is introduced during last rebase. As is mentioned in https://github.com/kubernetes/node-problem-detector/pull/54#pullrequestreview-14090197, we should not try to static link when journald is enabled.

Makefile
Random-Liu: ACK. Will fix. Thanks.

Random-Liu:
> One thing that didn't work at all was the magic with /etc/localtime, and I ended up using an init container to create a symlink instead of trying to mount the file directly.

What problem did you hit?

ankon:
> What problem did you hit?

It (not sure who exactly - docker, kubelet, kubernetes?) refused to start the pod, because it couldn't mkdir the _directory_ `/etc/localtime` (ending in the usual CrashLoopBackOff behavior).

FWIW: 
```sh
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"5", GitVersion:"v1.5.1", GitCommit:"82450d03cb057bab0950214ef122b67c83fb11df", GitTreeState:"clean", BuildDate:"2016-12-14T00:57:05Z", GoVersion:"go1.7.4", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"5", GitVersion:"v1.5.1", GitCommit:"82450d03cb057bab0950214ef122b67c83fb11df", GitTreeState:"clean", BuildDate:"1970-01-01T00:00:00Z", GoVersion:"go1.7.1", Compiler:"gc", Platform:"linux/amd64"}
$ minikube version
minikube version: v0.15.0
$ uname -r
4.9.3-200.fc25.x86_64
```

Makefile
Random-Liu: Done.

Makefile
Random-Liu: Done.

Random-Liu:
> It (not sure who exactly - docker, kubelet, kubernetes?) refused to start the pod, because it couldn't mkdir the directory /etc/localtime (ending in the usual CrashLoopBackOff behavior).

Could you post the log here? In fact, I don't know whether npd works for minikube. :)

dchen1107:
LGTM

Random-Liu:
@dchen1107 Thanks a lot for reviewing! :)

Makefile
andyxning: @Random-Liu @jfilak 

According to [#39(discussion)](https://github.com/kubernetes/node-problem-detector/pull/39#discussion_r89067352), we should also make `./bin/node-problem-detector` target phony. WDYT.

PR #40, opened 2016-11-10
Support getting node name from k8s API

apatil:
Hi @Random-Liu, I noticed that you were grabbing the node name from the downward API, which isn't supported in k8s 1.3.7 and below. This patch adds a fallback to using the k8s API at runtime. What do you think?

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/node-problem-detector/40)

### DIFF
--- a/pkg/problemclient/problem_client.go
+++ b/pkg/problemclient/problem_client.go
@@ -65,6 +65,18 @@ func NewClientOrDie() Client {
 	// 2. For some cloud providers, os.Hostname is different from the real hostname.
 	c.nodeName = os.Getenv("NODE_NAME")
 	if c.nodeName == "" {
+		// k8s v1.3.7 and earlier doesn't provide the node name in the dowwnward API.
+		// However, we can get it from the k8s API.
+		var err error
+		pod, err := c.client.Pods(os.Getenv("POD_NAMESPACE")).Get(os.Getenv("POD_NAME"))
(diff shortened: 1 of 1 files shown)


Random-Liu:
@apatil Thanks for the patch!

Could you sign the agreement?
I tried this before https://github.com/kubernetes/node-problem-detector/pull/29, but I remember there will be something wrong... Let me have a try again, if there is no problem then let's get this in for backward compatibility.

apatil:
I signed it!

Random-Liu:
I remember the problem. https://github.com/kubernetes/kubernetes/issues/30491

This doesn't work on master node, because pod can't access apiserver on master node.
However, this is not your problem. NPD on master can't even update node condition if it can't access apiserver.

andyxning:
@Random-Liu @apatil Maybe we can choose to change `node-problem-detector.yaml` file not to use downward api to get `NODE_NAME` when used in Kubernetes under 1.4. I prefer we continue to use `NODE_NAME` environment to store node ip. When used under kubernetes under 1.4, users can choose whatever method they prefer to get the node ip and store it in `NODE_NAME` environment. 

For those using Kubernetes with Service type NodePort, [#34923] (https://github.com/kubernetes/kubernetes/issues/34923) is one way to get node ip inside a pod.

PR #41, opened 2016-11-14
logwatchers: add new kmsg-based kernel log watcher

### LINKED ISSUE #14: KernelMonitor: Add journald support

euank:
All the other loggers (afaik), like rsyslog and journald and what have you, simply read `/dev/kmsg`.

It's probably easier for everyone if we just read that directly too!

I put together a bit of code doing so [over here](https://github.com/euank/go-kmsg-parser/blob/v1.0.0/kmsgparser/kmsgparser.go), and this commit leverages that code for the node-problem-detector.

I plan to replace cadvisor's similar hacky log parsing with the same code as well.

So, what are the implications?
Well, one thing this can't do as well is read far into the past. This can't read further back than the ring buffer extends, so on startup it's quite possible it won't get as many old messages. That means really large lookback values will behave differently now because they'll read less.

That could have already happened depending on logrotation policy, but now it's a lot harder for the user to control (kernel dmesg size is much harder to change :).

However, I think the significant simplicity and cross-distro gains by doing this are worth it.

~~I also preemptively removed the old integrations as I think this makes them fully redundant, as well as the translators thingy since why would you want to translate when you're reading the one and only true source?~~

Edit: I added this as an additional plugin, but didn't change any defaults or such.

Fixes #14 #39 

cc @Random-Liu @AdoHe @derekwaynecarr

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/node-problem-detector/41)

### DIFF
--- a/pkg/systemlogmonitor/README.md
+++ b/pkg/systemlogmonitor/README.md
@@ -8,10 +8,11 @@ the configuration files. (
 [`config/kernel-monitor.json`](https://github.com/kubernetes/node-problem-detector/blob/master/config/kernel-monitor.json) as an example).
 The rule list is extensible.
 
-## Limitations
+## Supported sources
 
-* System Log Monitor only supports file based log and journald now, but it is easy
-  to extend it with [new log watcher](#new-log-watcher)
+* System Log Monitor currently supports file-based logs, journald, and kmsg.
+  Additional sources can be added by implementing a [new log
+  watcher](#new-log-watcher).
 
 ## Add New NodeConditions
 
@@ -44,10 +45,10 @@ with new rule definition:
 
 System log monitor supports different log management tools with different log
 watchers:
-* [filelog](https://github.com/kubernetes/node-problem-detector/blob/master/pkg/systemlogmonitor/logwatchers/filelog): Log watcher for
+* [filelog](./logwatchers/filelog): Log watcher for
 arbitrary file based log.
-* [journald](https://github.com/kubernetes/node-problem-detector/blob/master/pkg/systemlogmonitor/logwatchers/journald): Log watcher for
-journald.
+* [journald](.//logwatchers/journald): Log watcher for
+* [kmsg](./logwatchers/kmsg): Log watcher for the kernel ring buffer device, /dev/kmsg.
 Set `plugin` in the configuration file to specify log watcher.
 
 ### Plugin Configuration
@@ -66,6 +67,7 @@ Log watcher specific configurations are configured in `pluginConfig`.
   * timestampFormat: The format of the timestamp. The format string is the time
     `2006-01-02T15:04:05Z07:00` in the expected format. (See
     [golang timestamp format](https://golang.org/pkg/time/#pkg-constants))
+* **kmsg**
 
 ### Change Log Path
 
@@ -78,6 +80,6 @@ field in the configurtion file is the log path. You can always configure
 
 ### New Log Watcher
 
-System log monitor uses [Log
-Watcher](https://github.com/kubernetes/node-problem-detector/blob/master/pkg/systemlogmonitor/logwatchers/types/log_watcher.go) to support different log management tools.
-It is easy to implement a new log watcher.
+System log monitor uses [Log Watcher](./logwatchers/types/log_watcher.go) to
+support different log management tools.  It is easy to implement a new log
+watcher.
--- a/pkg/systemlogmonitor/logwatchers/kmsg/log_watcher.go
+++ b/pkg/systemlogmonitor/logwatchers/kmsg/log_watcher.go
@@ -1,0 +1,115 @@
+/*
+Copyright 2017 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package kmsg
+
+import (
+	"bufio"
+	"fmt"
+	"strings"
+	"time"
+
+	utilclock "code.cloudfoundry.org/clock"
+	"github.com/euank/go-kmsg-parser/kmsgparser"
+	"github.com/golang/glog"
+
+	"k8s.io/node-problem-detector/pkg/systemlogmonitor/logwatchers/types"
+	logtypes "k8s.io/node-problem-detector/pkg/systemlogmonitor/types"
+	"k8s.io/node-problem-detector/pkg/systemlogmonitor/util"
+)
+
+type kernelLogWatcher struct {
+	cfg    types.WatcherConfig
+	logCh  chan *logtypes.Log
+	tomb   *util.Tomb
+	reader *bufio.Reader
+
+	kmsgParser kmsgparser.Parser
+	clock      utilclock.Clock
+}
+
+// NewKmsgWatcher creates a watcher which will read messages from /dev/kmsg
+func NewKmsgWatcher(cfg types.WatcherConfig) types.LogWatcher {
+	kmsgparser.NewParser()
+	return &kernelLogWatcher{
+		cfg:  cfg,
+		tomb: util.NewTomb(),
+		// Arbitrary capacity
+		logCh: make(chan *logtypes.Log, 100),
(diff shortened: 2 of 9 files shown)


euank:
Testing done:

I manually built and published an image (available at `euank/node-problem-detector:v0.6`). I then edited the daemonset to reference that image, deployed it, and ssh'd into a node.

I ran "echo "BUG: unable to handle kernel NULL pointer dereference at some-location" > /dev/kmsg" as root, verified it showed up in `dmesg`, and then verified `kubectl describe no` had an event of `1m		1m		1	{kernel-monitor k8s1}			Warning		KernelPanic	BUG: unable to handle kernel NULL pointer dereference at some-location`.

Random-Liu:
> Well, one thing this can't do as well is read far into the past. 

Yeah, this is the reason why I didn't use it before. We want to know whether the node panicked last time and the panic reason.

Random-Liu:
However, I think we could keep reading from `/dev/kmsg` as an optional log reader, since it supports more os distro, although with less look back support.

euank:
@Random-Liu That makes some sense, but an alternate solution would be checkpointing. In general, we don't care about lookbehind on the initial boot/setup, only on reboots where the daemonset starts up again.
In those cases, if we checkpoint then we can be our own super limited logdaemon for the last $lookBehind duration of logs.

That might actually be simpler than integrating journald and syslog parsers.

The only benefit those loggers have over us checkpointing is that journald at some point was gonna integrate with pstore (https://www.kernel.org/doc/Documentation/ABI/testing/pstore) and, because it plans to delete after consuming, it'll be the only source of truth for it..... but that's not being done yet so that's a future theoretical feature!

Random-Liu:
> For example, for kernel panic (assuming the node is configured as reboot on kernel panic), we have to look back to figure out what happened. I think this is a useful feature that we want to keep, given that we already encountered some kernel panics:

I recently found out that there is no easy way to log kernel panics into logs, because for safety, kernel won't want to write anything when it's already panicked. So, my argument is wrong.

However, I still think look back is needed when NPD is firstly started or restarted. :)

However,

dchen1107:
@euank, #39 was merged. Can you please rebase this one, so that we can continue reviewing? Thanks!

euank:
Rebased, manually tested that it worked as I expected.

Random-Liu:
@euank Thanks a lot! Will try to review it soon.

Random-Liu:
@euank Will review this PR soon. Sorry for the delay, and thanks a lot for helping!

pkg/systemlogmonitor/README.md
```diff
@@ -44,10 +45,10 @@ with new rule definition:
 
 System log monitor supports different log management tools with different log
 watchers:
-* [filelog](https://github.com/kubernetes/node-problem-detector/blob/master/pkg/systemlogmonitor/logwatchers/filelog): Log watcher for
+* [filelog](./logwatchers/filelog): Log watcher for
 arbitrary file based log.
-* [journald](https://github.com/kubernetes/node-problem-detector/blob/master/pkg/systemlogmonitor/logwatchers/journald): Log watcher for
-journald.
+* [journald](.//logwatchers/journald): Log watcher for
```
Random-Liu: for journald

pkg/systemlogmonitor/README.md
```diff
@@ -66,6 +67,7 @@ Log watcher specific configurations are configured in `pluginConfig`.
   * timestampFormat: The format of the timestamp. The format string is the time
     `2006-01-02T15:04:05Z07:00` in the expected format. (See
     [golang timestamp format](https://golang.org/pkg/time/#pkg-constants))
+* **kmsg**
```
Random-Liu: **kmsg**: no configurations for now.

pkg/systemlogmonitor/logwatchers/kmsg/log_watcher.go
```diff
@@ -0,0 +1,115 @@
...
+	"k8s.io/node-problem-detector/pkg/systemlogmonitor/logwatchers/types"
+	logtypes "k8s.io/node-problem-detector/pkg/systemlogmonitor/types"
+	"k8s.io/node-problem-detector/pkg/systemlogmonitor/util"
+)
+
+type kernelLogWatcher struct {
+	cfg    types.WatcherConfig
+	logCh  chan *logtypes.Log
+	tomb   *util.Tomb
+	reader *bufio.Reader
```
Random-Liu: Not used.

pkg/systemlogmonitor/logwatchers/kmsg/log_watcher.go
```diff
@@ -0,0 +1,115 @@
...
+
+	kmsgParser kmsgparser.Parser
+	clock      utilclock.Clock
+}
+
+// NewKmsgWatcher creates a watcher which will read messages from /dev/kmsg
+func NewKmsgWatcher(cfg types.WatcherConfig) types.LogWatcher {
+	kmsgparser.NewParser()
```
Random-Liu: Unused code, remove it.

pkg/systemlogmonitor/logwatchers/kmsg/log_watcher.go
```diff
@@ -0,0 +1,115 @@
...
+			return
+		case msg := <-kmsgs:
+			glog.V(5).Infof("got kernel message: %+v", msg)
+			if msg.Message == "" {
+				continue
+			}
+
+			// Discard too old messages
+			if k.clock.Since(msg.Timestamp) > lookback {
+				glog.V(5).Infof("throwing away msg %v for being too old: %v > %v", msg.Message, msg.Timestamp.String(), lookback.String())
```
Random-Liu: nit: s/throwing/Throwing

pkg/systemlogmonitor/logwatchers/kmsg/log_watcher.go
```diff
@@ -0,0 +1,115 @@
...
+		close(k.logCh)
+		k.tomb.Done()
+	}()
+	kmsgs := k.kmsgParser.Parse()
+
+	for {
+		select {
+		case <-k.tomb.Stopping():
+			glog.Infof("Stop watching kernel log")
+			k.kmsgParser.Close()
```
Random-Liu: nit: log error of `Close`.

pkg/systemlogmonitor/logwatchers/kmsg/log_watcher_test.go
```diff
@@ -0,0 +1,160 @@
...
+		timeout := time.After(100 * time.Millisecond)
+		select {
+		case log := <-logCh:
+			t.Errorf("unexpected extra log: %+v", *log)
+		case <-timeout:
+		}
+	}
+}
+
+type fakeKmsgReader struct {
```
Random-Liu: Not used.

Random-Liu:
@euank Do you have time to address the comments in this PR?

Random-Liu:
The PR LGTM overall, only several small cleanups are needed.

Offline discussed with @dchen1107, since we want this PR badly for 1.7, I'll merge this first, and send another PR to address the comments. :)

pkg/systemlogmonitor/README.md
Random-Liu: Done

pkg/systemlogmonitor/README.md
Random-Liu: Done

pkg/systemlogmonitor/logwatchers/kmsg/log_watcher.go
Random-Liu: Done

pkg/systemlogmonitor/logwatchers/kmsg/log_watcher.go
Random-Liu: Done.

pkg/systemlogmonitor/logwatchers/kmsg/log_watcher.go
Random-Liu: Done.

pkg/systemlogmonitor/logwatchers/kmsg/log_watcher.go
Random-Liu: Done

pkg/systemlogmonitor/logwatchers/kmsg/log_watcher_test.go
Random-Liu: Done.

PR #42, opened 2016-11-21
makefile: Make misc improvements

euank:
* Add test target and re-use in travis
* Add timeout test target
* Build binary into ./bin folder
* Phony everything

I thought I had made a PR with this, but nope, it was just sitting around in my branch for the last week.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+.PHONY: all container push clean node-problem-detector
+
 all: push
 
 # See node-problem-detector.yaml for the version currently running-- bump this ahead before rebuilding!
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -1,18 +1,27 @@
...
+
 all: push
 
 # See node-problem-detector.yaml for the version currently running-- bump this ahead before rebuilding!
 TAG = v0.2
 
 PROJ = google_containers
 
-node-problem-detector: node_problem_detector.go
+PKG_SOURCES := $(shell find pkg -name '*.go')
```
Random-Liu: Why not include go files in `vendor`? We should also rebuild if `vendor` is updated, right?

Makefile
```diff
@@ -1,18 +1,27 @@
+.PHONY: all phony container push clean node-problem-detector
```
Random-Liu: What does the `phony` mean here?

Makefile
euank: It's a premature optimization. Managing the timestamps for that many files is a bit slower and having to do `make clean` after a vendor update doesn't seem like that big of a deal.

Random-Liu: Hm, if so. Let's not check `vendor` for now. :)

Random-Liu:
LGTM. Will merge this after test passed.

PR #53, opened 2016-12-21
add fmt and vet make target

andyxning:
Add `fmt` and `vet` make target. And, add `fmt` and `vet` dependencies for `node-problem-detector` target.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/node-problem-detector/53)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: all container push clean node-problem-detector
+.PHONY: all container push clean node-problem-detector vet fmt
 
 all: push
 
@@ -9,7 +9,13 @@ PROJ = google_containers
 
 PKG_SOURCES := $(shell find pkg -name '*.go')
 
-node-problem-detector: $(PKG_SOURCES) node_problem_detector.go
+vet:
+	go list ./... | grep -v "./vendor/*" | xargs go vet
+
+fmt:
+	find . -type f -name "*.go" | grep -v "./vendor/*" | xargs gofmt -s -w -l
+
+node-problem-detector: $(PKG_SOURCES) node_problem_detector.go fmt vet
 	GOOS=linux go build -ldflags '-w -extldflags "-static"' -o node-problem-detector
 
 test:


andyxning:
@Random-Liu PTAL.

Random-Liu:
@andyxning Seems like something wrong with the rebase. :)

andyxning:
@Random-Liu Done.

Random-Liu:
@andyxning There are still 2 duplicated commits. :)

andyxning:
@Random-Liu Done Again. :)

Makefile
```diff
@@ -9,7 +9,13 @@ PROJ = google_containers
 
 PKG_SOURCES := $(shell find pkg -name '*.go')
 
-node-problem-detector: $(PKG_SOURCES) node_problem_detector.go
+vet:
+	go list ./... | grep -v "./vendor*" | xargs go vet
```
Random-Liu: This seems a little weird to me although it works for now.

go list is listing packages, while `grep -v "./vendor*"` seems to assume that it is listing files.
In the future, if there is a package called "abc/vendorefg", it will also be excluded.

Makefile
```diff
@@ -9,7 +9,13 @@ PROJ = google_containers
...
+
+fmt:
+	find . -type f -name "*.go" | grep -v "./vendor*" | xargs gofmt -s -w
```
Random-Liu: How about `grep -v "./vendor/*" to make it less ambiguous.

Random-Liu: How about `-l` to also list bad files?

andyxning: @Random-Liu Agree to add `-l` to list bad files. Done.

Makefile
andyxning: @Random-Liu Just like next comment, we can change `grep -v "./vendor*"` to `grep -v "./vendor/*"` to minimize the impact.

Random-Liu: Just to make sure, is do you expect `.` to match a dot or matching anything?

andyxning: No, What i mean is that i want to exclude node-problem-detector binary file under root directory. I forgot that it does not need a dot prefix. I have confirmed this according to  http://stackoverflow.com/a/3637678/1889327

andyxning: What `.` means is that it means anything prefixing vendor directory. Because we run `go list ./...` in the root directory, it output
```
k8s.io/node-problem-detector
k8s.io/node-problem-detector/pkg/condition
k8s.io/node-problem-detector/pkg/kernelmonitor
k8s.io/node-problem-detector/pkg/kernelmonitor/translator
k8s.io/node-problem-detector/pkg/kernelmonitor/types
k8s.io/node-problem-detector/pkg/kernelmonitor/util
k8s.io/node-problem-detector/pkg/problemclient
k8s.io/node-problem-detector/pkg/problemdetector
k8s.io/node-problem-detector/pkg/types
k8s.io/node-problem-detector/pkg/util
k8s.io/node-problem-detector/pkg/util/nethealth
k8s.io/node-problem-detector/vendor/code.cloudfoundry.org/clock
k8s.io/node-problem-detector/vendor/code.cloudfoundry.org/clock/fakeclock
k8s.io/node-problem-detector/vendor/github.com/blang/semver
k8s.io/node-problem-detector/vendor/github.com/coreos/go-oidc/http
```

What i mean is to match`.*/vendor/.*` in Golang regexp expression.

Random-Liu: @andyxning Yeah, I think you should be assuming to match everything.

Random-Liu:
LGTM. And could you help me review this https://github.com/kubernetes/node-problem-detector/pull/56?

PR #55, opened 2016-12-21
Remove dead code in kernel monitor

andyxning:
Check for `linux` os with `runtime.GOOS` instead of `syscall.Sysinfo_t` which is not available on MacOS.

This will is possible to develop and `make node-problem-detector` on MacOS when #39 is merged with supporting disable `systemd` with an environment variable.

All available `runtime.GOOS` values can be found in [this page](https://golang.org/doc/install/source#environment).

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/node-problem-detector/55)

### DIFF
--- a/pkg/kernelmonitor/kernel_monitor.go
+++ b/pkg/kernelmonitor/kernel_monitor.go
@@ -20,7 +20,6 @@ import (
 	"encoding/json"
 	"io/ioutil"
 	"regexp"
-	"syscall"
 	"time"
 
 	kerntypes "k8s.io/node-problem-detector/pkg/kernelmonitor/types"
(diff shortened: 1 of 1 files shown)


andyxning:
@Random-Liu PTAL.

Random-Liu:
In fact, I've removed this line https://github.com/kubernetes/node-problem-detector/pull/39/files#diff-0375f911d80ffbc3325bbbe3e41ccf02L85.

Previously, I used `Sysinfo_t` to get system boot time, and use it to calculate the kernel log time stamp. Now, we don't use that anymore. I think we can remove it now. :)

andyxning:
Sorry for not checking #39 before making a PR.

pkg/kernelmonitor/kernel_monitor.go
```diff
@@ -81,10 +81,8 @@ func NewKernelMonitorOrDie(configPath string) KernelMonitor {
 		panic(err)
 	}
 	glog.Infof("Finish parsing log file: %+v", k.config)
-	var info syscall.Sysinfo_t
-	err = syscall.Sysinfo(&info)
-	if err != nil {
-		panic(err)
+	if runtime.GOOS != "linux" {
```
Random-Liu: Even we check operating system. This may not be the place to do that.

Random-Liu:
@andyxning Let's just remove the dead code here. We don't need to do the operating system check now. :)

pkg/kernelmonitor/kernel_monitor.go
andyxning: Done.

andyxning:
@Random-Liu Done.

Random-Liu:
LGTM

Issue #68, opened 2017-01-11
Add --hostname-override flag.


Random-Liu:
Currently NPD gets hostname override from `NODE_NAME` env. https://github.com/kubernetes/node-problem-detector/blob/master/pkg/problemclient/problem_client.go#L73

This is not friendly for standalone mode.

We should add a `--hostname-override` flag, and set the flag with `NODE_NAME` from downward api in the yaml file. https://github.com/kubernetes/node-problem-detector/blob/master/node-problem-detector.yaml#L21

andyxning:
@Random-Liu I suppose sticking with `NODE_NAME`  for node-problem-detector wherever it is running. This makes the command line consistent when running  node-problem-detector in both situation. :)

We may need to add more clear info about this behavior in README.md instead of only recording it in CHANGELOG.md. 

Today we may use `systemd` or other process manager tools to manager npd daemon. We can simple wrap npd in a shell script where a `NODE_NAME` environment variable should be customized and exported when needed.

shyamjvs:
I agree with @andyxning on this. Adding a --hostname-override flag wouldn't really help much as we would need to provide the value for the flag most probably through an env variable anyway if we do not know the node name statically. For instance a replication controller for a pod in which the npd is a container, so you pretty much don't know the name of a specific replica. An example of this is the kubemark [hollow-node](https://github.com/shyamjvs/kubernetes/blob/25c5430e50384babb83f018836193a1fc8893d69/test/kubemark/resources/hollow-node_template.json#L153).

Random-Liu:
@andyxning @shyamjvs 
For DaemonSet, if we configure the pod to use HostNetwork, we can also let the pod to use os.Hostname.
For Standalone, if kubelet registers the node with --hostname-override, we need to change the hostname NPD using, too.

So in fact, there is no much difference between DaemonSet and Standalone on the hostname setting. That's why I prefer a --hostname-override flag which explicitly overrides the hostname. Environment variable doesn't seem to be a good user interface. :)

shyamjvs:
> I prefer a --hostname-override flag which explicitly overrides the hostname and has clear description when we run help command. Environment variable doesn't seem to be a good user interface to me. :)
sgtm

andyxning:
Sounds reasonable. @Random-Liu I think i can add this functionality and at the same time adding more docs about `--hostname-override` in README by updating #72

Random-Liu:
[...]

PR #72, opened 2017-01-12
detail how node-problem-detector get node name in README

andyxning:
Ref: #68

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/node-problem-detector/72)

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ For example, to run without auth, use the following config:
 http://APISERVER_IP:APISERVER_PORT?inClusterConfig=false
 ```
 Refer [heapster docs](https://github.com/kubernetes/heapster/blob/1e40b0f4b5eeb3f02e11ee22c2b6fda36b6e6ea1/docs/source-configuration.md#kubernetes) for a complete list of available options.
+* `-hostname-override`: A customized node name used for node-problem-detector to update conditions and emit events. node-problem-detector gets node name first from `hostname-override`, then `NODE_NAME` environment variable and finally fall back to `os.Hostname`.
 
 ## Build Image
 Run `make` in the top directory. It will:
--- a/cmd/node_problem_detector.go
+++ b/cmd/node_problem_detector.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/node-problem-detector/pkg/version"
 
 	"github.com/golang/glog"
+	"fmt"
 )
 
 // TODO: Move flags to options directory.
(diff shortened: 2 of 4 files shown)


andyxning:
@Random-Liu PTAL.

README.md
```diff
@@ -62,6 +62,7 @@ For example, to run without auth, use the following config:
 http://APISERVER_IP:APISERVER_PORT?inClusterConfig=false
 ```
 Refer [heapster docs](https://github.com/kubernetes/heapster/blob/1e40b0f4b5eeb3f02e11ee22c2b6fda36b6e6ea1/docs/source-configuration.md#kubernetes) for a complete list of available options.
+* `-hostname-override`: A customized node name used for node-problem-detector to update conditions and emit events. node-problem-detector gets node name first from `hostname-override`, then `NODE_NAME` environment variable and finally fall back to `os.Hostname` when node name can not be inferred from neither `hostname-override` nor `NODE_NAME` environment variable.
```
Random-Liu: `when node name can not be inferred from neither `hostname-override` nor `NODE_NAME` environment variable` seems unnecessary to me.

cmd/node_problem_detector.go
```diff
@@ -41,6 +42,36 @@ func validateCmdParams() {
 	}
 }
 
+func mustGetNodeName() string {
+	var nodeName string
+
+	// Check hostname override first for customized node name.
+	if *hostnameOverride != "" {
```
Random-Liu:
```
if *hostnameOverride != "" {
  return *hostnameOverride
}
```

cmd/node_problem_detector.go
```diff
@@ -41,6 +42,36 @@ func validateCmdParams() {
...
+	}
+
+	// Get node name from environment variable NODE_NAME
+	// By default, assume that the NODE_NAME env should have been set with
+	// downward api or user defined exported environment variable. We prefer it because sometimes
+	// the hostname returned by os.Hostname is not right because:
+	// 1. User may override the hostname.
+	// 2. For some cloud providers, os.Hostname is different from the real hostname.
+	nodeName = os.Getenv("NODE_NAME")
+	if nodeName == "" {
```
Random-Liu: Less indent is preferred. My original code is not good. :)
```
nodeName := os.Getenv("NODE_NAME")
if nodeName != "" {
  return nodeName
}

nodeName, err := os.Hostname()
if err != nil {
  panic("Failed to get host name: %v", err)
}

return nodeName
```

cmd/node_problem_detector.go
```diff
@@ -41,6 +42,36 @@ func validateCmdParams() {
 	}
 }
 
+func mustGetNodeName() string {
```
Random-Liu: We usually use `OrDie` for function which may panic,`getNodeNameOrDie`

cmd/node_problem_detector.go
```diff
@@ -50,7 +81,9 @@ func main() {
 		os.Exit(0)
 	}
 
+	nodeName := mustGetNodeName()
+
 	k := kernelmonitor.NewKernelMonitorOrDie(*kernelMonitorConfigPath)
-	p := problemdetector.NewProblemDetector(k, *apiServerOverride)
+	p := problemdetector.NewProblemDetector(k, *apiServerOverride, nodeName)
```
Random-Liu: Can we create the client here, and pass the client into `problemdetector` so that we don't need to pass these variables all the way down?

README.md
andyxning: It is indeed verbose. :)

cmd/node_problem_detector.go
andyxning: Done.

cmd/node_problem_detector.go
andyxning: Having thought that `must` prefix is a convention for function which may panic in Golang. Since `OrDie` is the convention for kubernetes, we should stick with it. :)

cmd/node_problem_detector.go
andyxning: Refactored

andyxning: Refactored

cmd/node_problem_detector.go
andyxning: @Random-Liu I agree with you.

BTW, as we may add more command line argument, we may need to add a `Conf` struct and make all command line arguments as its fields. When we need a command line argument, we just need to reference an `Conf` instance. This will also make the code more readable. :)

I suggest we make another PR to make this refactor. WDYT.

andyxning: @Random-Liu I agree with you.

BTW, as we may add more command line argument, we may need to add a `Conf` struct and make all command line arguments as its fields. When we need a command line argument, we just need to reference an `Conf` instance. This will also make the code more readable. :)

I suggest we make another PR to make this refactor. WDYT.

andyxning: @Random-Liu I agree with you.

BTW, as we may add more command line argument, we may need to add a `Conf` struct and make all command line arguments as its fields. When we need a command line argument, we just need to reference an `Conf` instance. This will also make the code more readable. :)

I suggest we make another PR to make this refactor. WDYT.

cmd/node_problem_detector.go
Random-Liu: @andyxning Thanks.

cmd/node_problem_detector.go
Random-Liu: Yeah, I agree.

In fact, following the K8s convention, we should add an `option/` directory under `cmd/` and define the config object and flags there.

Feel free to file an issue or send a PR. :)

Random-Liu:
@andyxning LGTM. Thanks!

Random-Liu:
/lgtm

---







*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/node-problem-detector/72#-:-KamxP2fXsTPqIKLJ3Kf:b-8ks5qa)*

cmd/node_problem_detector.go
andyxning: @Random-Liu Will do this later.

PR #75, opened 2017-01-13
Made authFile override logic not fallback to in-cluster config

### LINKED ISSUE #73: apiserver-override flag in standalone mode not working in kubemark
So I create npd containers inside hollow-nodes (just for clarity: these are actually pods) of kubemark by passing the address of the kubemark master and the kubeconfig file through the override flag as follows:
```
"name": "hollow-node-problem-detector",
"image": "gcr.io/google_containers/node-problem-detector:v0.3",
"env": [
						{
							"name": "NODE_NAME",
							"valueFrom": {
								"fieldRef": {
									"fieldPath": "metadata.name"
								}
							}
						}
],
"command": [
						"/node-problem-detector",
						"--kernel-monitor=/config/kernel-monitor.json",
						"--apiserver-override=https://104.198.41.48:443?inClusterConfig=false&auth=/kubeconfig/npd_kubeconfig",
						"--alsologtostderr",
						"1>>/var/logs/npd_$(NODE_NAME).log 2>&1"
],
"volumeMounts": ....
```

I even checked inside the container using `kubectl exec` that the right kubeconfig file is present on the container and NODE_NAME env var is defined. Here's the kubeconfig:
```
apiVersion: v1
kind: Config
users:
- name: node-problem-detector
  user:
    token: 8qt8RLdwIKeZQ0QeVrSWg4BrqK3Cs8H8
clusters:
- name: kubemark
  cluster:
    insecure-skip-tls-verify: true
    server: https://104.198.41.48
contexts:
- context:
    cluster: kubemark
    user: node-problem-detector
  name: kubemark-npd-context
current-context: kubemark-npd-context
```

Despite this, the npd seems to communicate with the wrong master (the master of the real cluster underneath the kubemark cluster). It might probably be still using the inClusterConfig from the real node underneath the hollow-node, as that seems to be the only way by which it can get hold of the wrong master's IP address.

cc @Random-Liu @andyxning

shyamjvs:
Fixes #73 

Changed the function `GetKubeClientConfig`  to use `NewNonInteractiveClientConfig` instead of `NewNonInteractiveDeferredLoadingClientConfig` as we do not want to fallback to in-cluster-config if authFile is specified.
(Inspired from this example https://github.com/kubernetes/kubernetes/blob/master/cmd/kubelet/app/bootstrap.go#L149)
We would also want to have such a similar change in heapster.

cc @Random-Liu @andyxning

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/node-problem-detector/75)

### DIFF
--- a/vendor/k8s.io/heapster/common/kubernetes/configs.go
+++ b/vendor/k8s.io/heapster/common/kubernetes/configs.go
@@ -104,9 +104,19 @@ func GetKubeClientConfig(uri *url.URL) (*kube_client.Config, error) {
 		}
 
 		if authFile != "" {
-			if kubeConfig, err = kubeClientCmd.NewNonInteractiveDeferredLoadingClientConfig(
-				&kubeClientCmd.ClientConfigLoadingRules{ExplicitPath: authFile},
-				configOverrides).ClientConfig(); err != nil {
+			// Load structured kubeconfig data from the given path.
+			loader := &clientcmd.ClientConfigLoadingRules{ExplicitPath: authFile}
+			loadedConfig, err := loader.Load()
+			if err != nil {
+				return nil, err
+			}
+
+			// Flatten the loaded data to a particular restclient.Config based on the current context.
+			if kubeConfig, err = clientcmd.NewNonInteractiveClientConfig(
+				*loadedConfig,
+				loadedConfig.CurrentContext,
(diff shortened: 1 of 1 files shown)


Random-Liu:
Offline discussed with @shyamjvs, we should not change the vendor directly.

We should send this patch to heapster. :)

PR #81, opened 2017-01-21
Fix kernel monitor issues

### LINKED ISSUE #47: "unregister_netdevice" isn't necessarily a KernelDeadlock
I have a node running CoreOS 1221.0.0 with kernel version 4.8.6-coreos.

The node-problem-detector marked it with "KernelDeadlock 	True 	Sun, 04 Dec 2016 18:56:20 -0800 	Wed, 16 Nov 2016 00:03:33 -0800 	UnregisterNetDeviceIssue 	unregister_netdevice: waiting for lo to become free. Usage count = 1".

If I check my kernel log, I see the following:

```
$ dmesg -T | grep -i unregister_netdevice -C 3
[Wed Nov 16 08:02:19 2016] docker0: port 5(vethfd2807b) entered blocking state
[Wed Nov 16 08:02:19 2016] docker0: port 5(vethfd2807b) entered forwarding state
[Wed Nov 16 08:02:19 2016] IPv6: eth0: IPv6 duplicate address fe80::42:aff:fe02:1206 detected!
[Wed Nov 16 08:03:33 2016] unregister_netdevice: waiting for lo to become free. Usage count = 1
[Wed Nov 16 08:14:35 2016] vethafecb94: renamed from eth0
[Wed Nov 16 08:14:35 2016] docker0: port 2(veth807b9e2) entered disabled state
[Wed Nov 16 08:14:35 2016] docker0: port 2(veth807b9e2) entered disabled state
```

Clearly, the node managed to continue to perform operations after printing that message. In addition, pods continue to function just fine and there aren't any long-term issues for me on this node.

I know that the config of what counts as a deadlock is configurable, but perhaps the default configuration shouldn't include this, or the check should be more advanced for it, since as-is it could be quite confusing.

### LINKED ISSUE #48: "startPattern" is fragile and wrong on newer kernels
Broken out from [here](https://github.com/kubernetes/node-problem-detector/pull/39#issuecomment-265877841)

Currently, the config has a default of `"startPattern": "Initializing cgroup subsys cpuset",`

This pattern is meant to detect a node's boot process. Prior to the 4.5 kernel, this message was typically printed during boot of a node. After 4.5 however, due to [this change](https://github.com/torvalds/linux/commit/a5ae989957cbc3f3b7bc40677bd9e459c0917528), it is quite unlikely for that message to appear.

Furthermore, there's rarely a reason to detect whether a message is for the current boot in such a fragile way.

With the `kern.log` reader, every message is for the current boot because kern.log is usually handled where each `kern.log` file corresponds to one boot (e.g. `kern.log` is this boot, `kern.log.1` is the boot before, `kern.log.2.gz` the one before, etc). (EDIT: I'm wrong about this for gci at least)

With journald, the boot id is annotated in messages, and so it can accurately be correlated with the current boot id (see the "_BOOT_ID" record in journald messages).

With a kmsg reader, all messages will only be the current boot because kmsg is not persistent.

In none of those cases is `startPattern` useful. Each kernel log parsing plugin should be responsible for doing the right thing itself I think.

Random-Liu:
Based on https://github.com/kubernetes/node-problem-detector/pull/39. Will rebase after https://github.com/kubernetes/node-problem-detector/pull/39 gets merged.

Only the last commit is new.

This PR:
* Change `unregister_netdevice` to be an event to fix #47. /cc @euank 
* Change `KernelPanic` to `KernelOops` because we can't really handle kernel panic by simply parsing kernel log.
  * We can consider supporting [`ramoops`](http://lxr.free-electrons.com/source/Documentation/ramoops.txt), but GCE doesn't support it now.
  * Other options are to use `serial console` or `PVPANIC`, but it could not be achieved inside the node scope. We may introduce a cluster level node health monitoring component in the future.
* Use system boot time instead of [`StartPattern`](https://github.com/kubernetes/node-problem-detector/blob/master/config/kernel-monitor.json#L4) to fix #48 /cc @euank . Since `KernelPanic` is not in the scope any more, we can only care about current boot now.

/cc @dchen1107 @jfilak

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/node-problem-detector/81)

### DIFF
--- a/config/kernel-monitor.json
+++ b/config/kernel-monitor.json
@@ -1,8 +1,7 @@
 {
 	"plugin": "journald",
 	"logPath": "/var/log/journal",
-	"lookback": "10m",
-	"startPattern": "Initializing cgroup subsys cpuset",
+	"lookback": "5m",
 	"bufferSize": 10,
 	"source": "kernel-monitor",
 	"conditions": [
@@ -25,12 +24,17 @@
 		},
 		{
 			"type": "temporary",
-			"reason": "KernelPanic",
+			"reason": "UnregisterNetDevice",
+			"pattern": "unregister_netdevice: waiting for \\w+ to become free. Usage count = \\d+"
+		},
+		{
+			"type": "temporary",
+			"reason": "KernelOops",
 			"pattern": "BUG: unable to handle kernel NULL pointer dereference at .*"
 		},
 		{
 			"type": "temporary",
-			"reason": "KernelPanic",
+			"reason": "KernelOops",
 			"pattern": "divide error: 0000 \\[#\\d+\\] SMP"
 		},
 		{
@@ -44,12 +48,6 @@
 			"condition": "KernelDeadlock",
 			"reason": "DockerHung",
 			"pattern": "task docker:\\w+ blocked for more than \\w+ seconds\\."
-		},
-		{
-			"type": "permanent",
-			"condition": "KernelDeadlock",
-			"reason": "UnregisterNetDeviceIssue",
-			"pattern": "unregister_netdevice: waiting for \\w+ to become free. Usage count = \\d+"
 		}
 	]
 }
--- a/pkg/kernelmonitor/config.go
+++ b/pkg/kernelmonitor/config.go
@@ -34,8 +34,6 @@ type MonitorConfig struct {
 	DefaultConditions []types.Condition `json:"conditions"`
 	// Rules are the rules kernel monitor will follow to parse the log file.
 	Rules []kerntypes.Rule `json:"rules"`
-	// StartPattern is the pattern of the start line
-	StartPattern string `json:"startPattern, omitempty"`
 }
 
 // applyDefaultConfiguration applies default configurations.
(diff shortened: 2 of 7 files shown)


dchen1107:
@Random-Liu #39 was merged. Can you rebase this one so that I can review it?

Random-Liu:
@dchen1107 Rebased.

dalehamel:
:+1: PTAL, we just had a few heart attacks after seeing this show up in our node descriptions, thinking there was some sort of serious problem.

Curiously, it seems like it occurred when our nodes ran out of space.

Any nodes that did run out of inodes (using overlay2) showed this issue ones that didn't, didn't.

Random-Liu:
> :+1: PTAL, we just had a few heart attacks after seeing this show up in our node descriptions, thinking there was some sort of serious problem.

Initially this is added for a serious kernel issue https://github.com/docker/docker/issues/5618, https://github.com/docker/docker/issues/5618. However, it seems that for newer kernel, it is not a serious issue any more. :)

> Curiously, it seems like it occurred when our nodes ran out of space.

Hm, that's an interesting observation. In fact, I don't really know what happened behind the scene for the "unregister_netdevice" bug. :(

dchen1107:
LGTM

Issue #58, opened 2017-01-07
NPD Kubernetes 1.6 Planning


Random-Liu:
NPD (node problem detector) is introduced in Kubernetes 1.3 as a default add-on in GCE cluster.

At that time, it is mainly targeted on default GCE Kubernetes setup. However, as time goes by, some limitations were found such as [Journald support](#14), [Authentication Issue](#21), [Scalability Issue](#37)  which affected the adoption of NPD in many other environment.

In Kubernetes 1.6, we plan to invest some time to improve NPD, make it production ready and rollout it in GKE.

Here are the working items and priorities:
- [x] **[P0] Journald support.** Many important OS distros are using systemd now, such as GCI, CoreOS, CentOS etc. This is essential for NPD adoption. (Issue:  https://github.com/kubernetes/node-problem-detector/issues/14, PR: https://github.com/kubernetes/node-problem-detector/pull/39, https://github.com/kubernetes/node-problem-detector/pull/33, @AdoHe) 
- [x] **[P0] Apiserver client option override.** By default, NPD is running as DaemonSet and [use `InClusterConfig` to access apiserver](http://kubernetes.io/docs/user-guide/accessing-the-cluster/#accessing-the-api-from-a-pod). However, this does not work when service account is not available. (Issue: https://github.com/kubernetes/node-problem-detector/issues/27, https://github.com/kubernetes/node-problem-detector/issues/21). We should make the apiserver client option configurable, so that user can customize it based on their cluster setup. **This is prerequisite of Standalone mode** (PR: https://github.com/kubernetes/node-problem-detector/pull/49, @andyxning)
- [x] **[P0] Standalone mode.** Make it possible to run NPD standalone, possibly as a systemd service. DaemonSet is easy to deploy and manage. However, docker still stops all containers when it's dead ([`live-restore`](https://docs.docker.com/engine/admin/live-restore/) is still [in validation](https://github.com/kubernetes/kubernetes/issues/38303)). Because of this, NPD may not be able to detect problems when docker is unresponsive. (Issue: #76)
- [ ] **[P1] Integrate NPD with K8s e2e framework.** NPD is already running in e2e cluster, but the information it collects is not well-surfaced from the test framework. We should make it visible by failing the test or collecting via a dashboard (Issue: https://github.com/kubernetes/kubernetes/issues/30811).
- [x] **[P1] Scalability and performance.** https://github.com/kubernetes/node-problem-detector/issues/85
[...]

PR #84, opened 2017-01-27
Only change transition timestamp when condition is changed.

Random-Liu:
Previously, we update transition timestamp whenever a new problem is matched from the log.
However, the kernel problem sometimes shows up for multiple even infinity times, for example the hung task problem. This will cause a lot of unnecessary condition updates.

This PR changes kernel monitor to only update the transition timestamp and message when problem status or reason is changed. This makes sure that we only update condition when it actually changes.

This change is also for scalability (Ref https://github.com/kubernetes/node-problem-detector/issues/58)

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/node-problem-detector/84)

### DIFF
--- a/pkg/kernelmonitor/kernel_monitor.go
+++ b/pkg/kernelmonitor/kernel_monitor.go
@@ -161,10 +161,15 @@ func (k *kernelMonitor) generateStatus(logs []*kerntypes.KernelLog, rule kerntyp
 			condition := &k.conditions[i]
 			if condition.Type == rule.Condition {
 				condition.Type = rule.Condition
+				// Update transition timestamp and message when the condition
+				// changes. Condition is considered to be changed only when
+				// status or reason changes.
+				if !condition.Status || condition.Reason != rule.Reason {
+					condition.Transition = timestamp
+					condition.Message = message
+				}
 				condition.Status = true
-				condition.Transition = timestamp
 				condition.Reason = rule.Reason
-				condition.Message = message
 				break
 			}
 		}
--- a/pkg/kernelmonitor/kernel_monitor_test.go
+++ b/pkg/kernelmonitor/kernel_monitor_test.go
@@ -36,12 +36,13 @@ func TestGenerateStatus(t *testing.T) {
 		{
 			Type:       testConditionA,
 			Status:     true,
-			Transition: time.Now(),
+			Transition: time.Unix(500, 500),
+			Reason:     "initial reason",
 		},
 		{
 			Type:       testConditionB,
 			Status:     false,
-			Transition: time.Now(),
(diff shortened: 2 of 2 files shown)


Random-Liu:
@andyxning Would you like to help me review this? Thanks a lot!

andyxning:
@Random-Liu Will review this later. :)

andyxning:
/lgtm

---

Reviewed 2 of 2 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/node-problem-detector/84#-:-KcODUSsvrY6Gst6z_G5:b-8ks5qa)*

Random-Liu:
Apply LGTM based on https://github.com/kubernetes/node-problem-detector/pull/84#issuecomment-278037477.

andyxning:
@Random-Liu 

> Only change transition timestamp when condition is changed.

This PR may make a confusion between condition recovery and condition continuing. 
IIUC, code before this PR will update condition transition timestamp when the condition continues to happen. According to the `LastTransitionTime` in `kubectl describe node NODENAME`, we can infer  whether condition is still happening. 

After this PR, if a similar condition continues to happen, node problem detector will not update `LastTransitionTime`. Thus we can not make sure whether specific condition is still happening. What  we can conclude is that condition has happened **recently** in `LastTransitionTime`.

Random-Liu:
> IIUC, code before this PR will update condition transition timestamp when the condition continues to happen. According to the LastTransitionTime in kubectl describe node NODENAME, we can infer whether condition is still happening.

> After this PR is merged, if a similar condition continues to happen, node problem detector will not update LastTransitionTime. Thus we can not make sure whether specific condition is still happening. What we can conclude is that condition has happened recently in LastTransitionTime.

We may need to define more types of problems and behaviors in the future. But for now, we only have 2 types of problems: for serious ones which permanently change the node condition, we'll update the node condition; for less serious ones, we'll only report an event.

As for the serious one, once it happens, the node is considered to transit into a bad condition. At that time, we update the node condition, and set the `LastTransitionTime` to indicate that "The node becomes bad at this time". After that, no matter how many same problems are detected, they all indicate that the node is still in the bad condition, so we don't need to update the `LastTransitionTime`.

> Also, node problem detector will not update any condition with Status set to False(except the initial condition) indicating that condition has disappeared and recovered. :)

This is only the behavior of kernel monitor (or log monitor soon). For serious kernel problem, usually there is no easy recovery except rebooting the node, so there is no `reset` pattern to recover the condition for now. In the future, we may add it if needed.

PR #87, opened 2017-02-03
change long option prefix to short option prefix

andyxning:
This PR change long command line option prefix to short command line option prefix. I.e., `--` -> `-`.

@Random-Liu PTAL.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/node-problem-detector/87)

### DIFF
--- a/Dockerfile.in
+++ b/Dockerfile.in
@@ -20,4 +20,4 @@ RUN test -h /etc/localtime && rm -f /etc/localtime && cp /usr/share/zoneinfo/UTC
 
 ADD ./bin/node-problem-detector /node-problem-detector
 ADD config /config
-ENTRYPOINT ["/node-problem-detector", "--kernel-monitor=/config/kernel-monitor.json"]
+ENTRYPOINT ["/node-problem-detector", "-kernel-monitor=/config/kernel-monitor.json"]
--- a/node-problem-detector.yaml
+++ b/node-problem-detector.yaml
@@ -12,8 +12,8 @@ spec:
       - name: node-problem-detector
         command:
         - /node-problem-detector
-        - --logtostderr
-        - --kernel-monitor=/config/kernel-monitor.json
+        - -logtostderr
+        - -kernel-monitor=/config/kernel-monitor.json
         image: gcr.io/google_containers/node-problem-detector:v0.2
         imagePullPolicy: Always
         securityContext:


Random-Liu:
@andyxning I think we may want to use double dash for the long options. See http://serverfault.com/questions/387935/whats-the-difference-betwen-the-single-dash-and-double-dash-flags-on-shell-comm

andyxning:
@Random-Liu Agreed that we should use double dash for long command line option. However
* the output of `./bin/node-problem-detector -h` is like below which uses single dash for long command line option.
```
Usage of ./bin/node-problem-detector:
  -alsologtostderr
    	log to standard error as well as files
  -apiserver-override string
    	Custom URI used to connect to Kubernetes ApiServer
  -hostname-override string
    	Custom node name used to override hostname
  -kernel-monitor string
    	The path to the kernel monitor config file (default "/config/kernel-monitor.json")
  -log_backtrace_at value
    	when logging hits line file:N, emit a stack trace
  -log_dir string
    	If non-empty, write log files in this directory
  -logtostderr
    	log to standard error instead of files
  -stderrthreshold value
    	logs at or above this threshold go to stderr
  -v value
    	log level for V logs
  -version
    	Print version information and quit
  -vmodule value
    	comma-separated list of pattern=N settings for file-filtered logging
```
* the reason for the above is that we use `flag` package from golang instread of [spf13/pflag](https://github.com/spf13/pflag) which [kubectl_apply uses](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/util/factory_client_access.go#L343-L356)

This PR is just to make sure the example in docs or configurations be consistent with the output of  the help info which is true for kubelet, AFAIK, who is also using `spf13/pflag` package.

Random-Liu:
> the output of ./bin/node-problem-detector -h is like below which uses single dash for long command line option.

@andyxning Make sense to me. Thanks for your explanation. If so, I'm totally ok with this PR.

> If we need the same behavior of kubelet/kubectl for npd . I will make another PR. :)

Yeah, let's change to `spf13/pflag` in another PR. Thanks!

Random-Liu:
LGTM

andyxning:
@Random-Liu Yep. Will fire another PR later.

PR #89, opened 2017-02-03
fix npd binary generation

andyxning:
This PR fix [#39(discussion1)](https://github.com/kubernetes/node-problem-detector/pull/39#discussion_r89067352), [#39(discussion2)](https://github.com/kubernetes/node-problem-detector/pull/39#discussion_r99326431).

Way to repreduce(adjusted from [#39(discussion1)](https://github.com/kubernetes/node-problem-detector/pull/39#discussion_r89067352))

```
make
# The ./bin/node-problem-detector file didn't exist and make created it.
# systemd-journal is enabled.
make ENABLE_JOURNALD=0
# The ./bin/node-problem-detector file already exists and make skipped the ./bin/node-problem-detector target. 
# However the user believes that node-problem-detector is compiled with systemd-journal disabled.
```

/cc @dchen1107 @jfilak @kubernetes/node-problem-detector-reviewers

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/node-problem-detector/89)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@
 
 # Build the node-problem-detector image.
 
-.PHONY: all build-container build-tar build push-container push-tar push clean vet fmt version Dockerfile
+.PHONY: all build-container build-tar build push-container push-tar push clean vet fmt version Dockerfile node_problem_detector
 
 all: build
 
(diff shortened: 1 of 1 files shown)


Random-Liu:
Actually, I'm not sure how we should deal with this.
This PR makes the npd binary to be built every time we run make, even though source code is not changed, which is also not what we want.

However, as you see. The original implementation will not rebuild when we change the ENV.

andyxning:
@Random-Liu I Agree that we should skip rebuild npd binary when necessary since this may take long time. 

Since we can not simply make sure whether existing node problem detector binary satisfy the build requirement. IMO, we should first keep it right then performance. :)

andyxning:
@Random-Liu PTAL.

andyxning:
@Random-Liu Could please take a look at this? :)

andyxning:
@Random-Liu Could you take a look at this. :)

PR #92, opened 2017-02-07
Generalize the kernel monitor code.

Random-Liu:
For https://github.com/kubernetes/node-problem-detector/issues/44.
Based on https://github.com/kubernetes/node-problem-detector/pull/81 and https://github.com/kubernetes/node-problem-detector/pull/88.

his PR:
* Change kernel monitor to system log monitor, because we want to generalize kernel monitor to detect problems of different system daemons.
* Change `syslog` to `filelog`, because actually it does not only support syslog now, it should be able to support most file based logs.
* Update the README.md.

/cc @dchen1107 @kubernetes/node-problem-detector-reviewers

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/node-problem-detector/92)

PR #88, opened 2017-02-03
Add arbitray system log support

Random-Liu:
For https://github.com/kubernetes/node-problem-detector/issues/44.

Based on https://github.com/kubernetes/node-problem-detector/pull/81.

This PR:
* Add `PluginConfig` for plugin specific configuration.
* Add `source` in journald plugin configuration to make sdjournal `SYSTEM_IDENTIFIER` configurable.
* Add `timestamp` regular expression, `message` regular expression and `timestampFormat` configuration in syslog plugin configuration. (We plan to change this to be `filelog`, because it can handle not only syslog)

TODOs for this work:
* Add multiple configuration files support. NPD should start a system log monitor for each passed in configurtion, so that it can monitor different services at the same time.
* Rename `syslog` to `filelog`.
* Rename `kernel monitor` to `system log monitor`, change all "kernel" reference in the code.
* Optimize the efficiency to make the system log monitor more light weight.
* Add document for the new behavior.
* Update kubernetes e2e test to test the new behavior.

/cc @dchen1107 @kubernetes/node-problem-detector-reviewers

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/node-problem-detector/88)

Issue #44, opened 2016-11-22
Generalize kernel monitor.


Random-Liu:
Discussed with @apatil.

Kernel monitor was initially introduced to monitor kernel log and detect kernel issues.

However, in fact it could be extended to monitor other logs such as docker log, systemd log etc. by adding new [`translator`](https://github.com/kubernetes/node-problem-detector/blob/master/pkg/kernelmonitor/translator/translator.go). Currently it is already doable, but not very intuitive because:
1. All files, types and functions are named as `kernel xxx`.
2. Translator is not configurable.

We should refactor the code to make it easier and more intuitive to extend kernel monitor:
* Change `kernelmonitor` to `logmonitor`. We'll only use log monitor to monitor kernel log for K8s, but it should be easy for other users to reconfigure and extend it to monitor other logs.
* Extend the configuration to make translator and log source configurable after https://github.com/kubernetes/node-problem-detector/pull/41 landed, including:
  * Make the journald log filter configurable.
  * Make the translate function configurable.

/cc @kubernetes/sig-node

jfilak:
With systemd-journal plugin enabled you can monitor the node for user space core files (either generated by ABRT or systemd-coredump). ABRT also logs uncaught Python, Ruby and Java exceptions to systemd-journal (systemd developers are working on their own Python exception handler). In Kernel you don't need to limit yourself to oopses, you can also detect Hardware issues in the form of MCEs and pstore oopses.

Random-Liu:
@jfilak Thanks for your suggestion.

* On one hand, we'd like NPD to detect more issues. As you mentioned, MCEs and pstore are all very useful signals for bare metal.
* On the other hand, we'd also like to keep the default behavior minimum so as to make it light-weight. After all, problem is rare case, we should only catch problems people really care and avoid consuming too much resource.

That's why we want to make NPD as configurable and plug-able as possible. I'd like to:
* Make kernel log monitor more generic on the premise of efficiency. So that people can use it to monitor different kinds of system logs if they really need the last line of defense. However, the default behavior should be only monitoring kernel log to detect known kernel issues. If people have configurations for bare metal or other environments, it's welcome to share!
* Integrate with different problem detection solution.

Any ideas and suggestions are welcome!

Random-Liu:
We plan to add arbitrary log support in 1.6, there are following working items need to be done:
[...]

PR #94, opened 2017-02-08
Add multiple system log monitor support

Random-Liu:
For #44.
Based on #81, #88 and #92.

Only the last commit is new.

This PR change the `--log-monitor` flag to `--log-monitors`. Multiple log monitor config separated by comma can be passed through `--log-monitors`, NPD will start a separate log monitor for each of them.

This makes it possible to use NPD to monitor different system log at the same time.

/cc @dchen1107 @kubernetes/node-problem-detector-reviewers

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/node-problem-detector/94)

### DIFF
--- a/Dockerfile.in
+++ b/Dockerfile.in
@@ -20,4 +20,4 @@ RUN test -h /etc/localtime && rm -f /etc/localtime && cp /usr/share/zoneinfo/UTC
 
 ADD ./bin/node-problem-detector /node-problem-detector
 ADD config /config
-ENTRYPOINT ["/node-problem-detector", "--system-log-monitor=/config/kernel-monitor.json"]
+ENTRYPOINT ["/node-problem-detector", "--system-log-monitors=/config/kernel-monitor.json"]
--- a/README.md
+++ b/README.md
@@ -54,8 +54,10 @@ List of supported problem daemons:
 # Usage
 ## Flags
 * `--version`: Print current version of node-problem-detector.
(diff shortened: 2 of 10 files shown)


Random-Liu:
@dchen1107 Rebased. PTAL.

dchen1107:
I only reviewed one commit (Add multiple log monitoring support) in this pr since all other should be covered / merged through different pr. 

Two comments I have: 
1) can we have unittest for parsing --system-log-monitors? at least make sure the corresponding system-log-monitoring goroutine is invoked, and there is no leakage, etc.
2) can we have a node / cluster e2e tests covering the new feature too?

Random-Liu:
> can we have unittest for parsing --system-log-monitors? at least make sure the corresponding system-log-monitoring goroutine is invoked, and there is no leakage, etc.

Will do.

> can we have a node / cluster e2e tests covering the new feature too?

I manually tested this, and this works. I'll add a test case in NPD e2e test for this feature.

Random-Liu:
@dchen1107 Add a new commit to test goroutine leak.

In fact, with the unit test, we are still not sure whether a failed log monitor will leak any other resources instead of goroutine. However, resource leaking is originally hard to capture in unit test.

Random-Liu:
@dchen1107 The PR is ready for another review. PTAL!

dchen1107:
LGTM

PR #95, opened 2017-02-21
fix misspell "timestamp"

fate-grand-order:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/node-problem-detector/95)


Random-Liu:
The failure is a flake, we should fix it. I'll let it re-run.

PR #98, opened 2017-03-02
Fix journald plugin to only look at the current boot.

Random-Liu:
Previously I thought `go-system/sdjournal` will only look at current boot by default.

However, it turns out that is not the case when I verified npd with some reboot test.

Ideally we should use [`sd_journal_seek_monotonic_usec`](https://www.freedesktop.org/software/systemd/man/sd_journal_seek_head.html), or [`sd_journal_seek_cursor`](https://www.freedesktop.org/software/systemd/man/sd_journal_seek_head.html), but sdjournal doesn't support it very well. So in this PR, we use the system uptime to make sure we npd only looks at current boot.

@dchen1107 @kubernetes/node-problem-detector-reviewers

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/node-problem-detector/98)

### DIFF
--- a/pkg/systemlogmonitor/logwatchers/journald/log_watcher.go
+++ b/pkg/systemlogmonitor/logwatchers/journald/log_watcher.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


dchen1107:
LGTM

PR #133, opened 2017-08-10
given the daemonset a name in the docs

mhristof:
otherwise, `kubectl create -f` was complaining with

```
The DaemonSet "node-problem-detector" is invalid: spec.template.metadata.labels: Invalid value: map[string]string(nil): `selector` does not match template `labels`
```

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/node-problem-detector/133)

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,9 @@ metadata:
   name: node-problem-detector
 spec:
   template:
+    metadata:
+      labels:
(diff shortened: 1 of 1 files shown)


mhristof:
bumping this for the CLA checks

PR #134, opened 2017-08-23
return an error when error happens in SetConditions()

CaoShuFeng:
@Random-Liu

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/node-problem-detector/134)


Random-Liu:
My mistake, thanks for fixing!

PR #136, opened 2017-08-29
introduce boilerplate test

CaoShuFeng:
This change is introduced because I found that some copyritht headers are different from each other.

This change add boilerplate test copyed from:
https://github.com/kubernetes/kubernetes/tree/master/hack/boilerplate

I found that most file in node-problem-detector has a different
copyright format from kubernetes/kubernetes:

node-problem-detector:
```
Copyright 2017 The Kubernetes Authors All rights reserved.
```

kubernetes:
```
Copyright 2017 The Kubernetes Authors.
```

@Random-Liu should we use the same copyright with kubernetes?

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/node-problem-detector/136)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-# Copyright 2017 The Kubernetes Authors.
+# Copyright 2017 The Kubernetes Authors All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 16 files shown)


CaoShuFeng:
Hi, @Random-Liu , what's your opinion about this test?

If this test is not necessary, at least we can reformat all the copyright headers, I think.

PR #137, opened 2017-08-29
Adding in prometheus exporter for events and conditions

cstuntz:
**Reason for Pull**

Prometheus is becoming more and more popular to track metrics in Kubernetes clusters, and it is a very lightweight method of tracking node problems across a cluster. This PR adds in basic prometheus exporter for all events and conditions tracked by node problem detector, which makes tracking problems much easier.

**Three Main Changes**

- Add new file that is used to contain the map of counters.
-- This lets us dynamically add Prometheus counters from events and conditions as they appear, instead of needing to hard-code all of them.
- Using the Fetch method from the containers file, I add new counters or increment existing ones as the events are processed in problem_detector.go.
- Update the reason names so there aren't any duplicates.
-- Prometheus can't have any duplicate metric names registered.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/node-problem-detector/137)


cstuntz:
I've completed the CLA, and now fixing the dependencies

dhananjaysathe:
Bump

Issue #150, opened 2017-11-19
Tracking issue for adding custom plugins


andyxning:
This is a tracking issue for what we still need to do after adding custom plugins in #145 .

- [x] Complete `pkg/custompluginmonitor/README.md`
- [x] Support per-rule interval
- [x] Emitting events when Condition changes.  
* Added in https://github.com/kubernetes/node-problem-detector/pull/151
- [x] support unknown condition status. 
* Added in https://github.com/kubernetes/node-problem-detector/pull/151

Random-Liu:
> Emitting events when Condition changes

Comparing with actual event, condition change should be rare. If node condition changes too frequently, it will cause even more issues, e.g. kubelet can't update node condition.

So I think it's fine to generate an event for condition change.

Random-Liu:
We also need to add integration test for custom plugin in https://github.com/kubernetes/kubernetes/blob/master/test/e2e_node/node_problem_detector_linux.go

byxorna:
Is this still in flight? `master` never registers the custom plugin with `registerLogWatcher(customPluginName, ...)`. Is this something that is actively being worked on?

andyxning:
@byxorna For now, i am ok to work on any problem you have. But as @Random-Liu has no time to focus on this project for now, i can only help to address any issues or review PRs.

For merging or any other problem, we can only wait @Random-Liu to come back. :)

byxorna:
@andyxning cool, thank you! Is the custom plugin in a state where you will accept a PR to enable it?

Its unclear to me whether it is feature complete (but not hooked up) or still a WIP, because master has https://github.com/kubernetes/node-problem-detector/blob/master/config/custom-plugin-monitor.json committed, but will refuse to load that config due to `No create function found for plugin "custom"`

andyxning:
@byxorna 

For now it is usable in the master branch. But there do have an enhance PR from @Random-Liu to enhance custom plugin manager in https://github.com/kubernetes/node-problem-detector/pull/151.

Actually, i have used custom plugin manager internally with #151 . I will dig into the reason about `No create function found for plugin "custom"`.

andyxning:
@byxorna What command line did you use in your usage?

andyxning:
@byxorna Seems like that you run custom plugin monitor with `--system-log-monitors`. Just like something: `--system-log-monitors=custom-plugin-monitor.json`.

[...]

PR #145, opened 2017-10-18
add custom problem detector plugin

andyxning:
This PR will add custom plugin problem detector interface to node-problem-detector.

Proposal: [https://docs.google.com/document/d/1jK_5YloSYtboj-DtfjmYKxfNnUxCAvohLnsH5aGCAYQ/edit#](https://docs.google.com/document/d/1jK_5YloSYtboj-DtfjmYKxfNnUxCAvohLnsH5aGCAYQ/edit#)

Major changes:
* move `ApplyDefaultConfiguration` for `MonitorConfig` from function to method. [Ref]()
* move `ValidateRules ` for `MonitorConfig` from function to method for. [Ref](https://github.com/kubernetes/node-problem-detector/pull/145/files#diff-ae34550839d4dd21f93fdb2193f0798f)
* move `k8s.io/node-problem-detector/pkg/systemlogmonitor/util` to `k8s.io/node-problem-detector/pkg/util/tomb`. [Ref](https://github.com/kubernetes/node-problem-detector/pull/145/files#diff-fbf4ff688541f0b6d6243de3c50245cb)
* move `Rule` type for system log config to pkg `k8s.io/node-problem-detector/pkg/types`. [Ref](https://github.com/kubernetes/node-problem-detector/pull/145/files#diff-34dfcb9f9c18e85cfeb1525b94b82572)
* move `LogMonitor` interface to `k8s.io/node-problem-detector/pkg/types` and rename it to `Monitor`. [Ref](https://github.com/kubernetes/node-problem-detector/pull/145/files#diff-6d10bcbde757a47bc5d87ac0510e0215)
* Code clean. [Ref](https://github.com/kubernetes/node-problem-detector/pull/145/files#diff-6d10bcbde757a47bc5d87ac0510e0215L147)

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/node-problem-detector/145)

PR #151, opened 2017-11-21
Improve cpm

Random-Liu:
Based on https://github.com/kubernetes/node-problem-detector/pull/145.

Addressed part of https://github.com/kubernetes/node-problem-detector/issues/150.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/node-problem-detector/151)

### DIFF
--- a/pkg/custompluginmonitor/custom_plugin_monitor.go
+++ b/pkg/custompluginmonitor/custom_plugin_monitor.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/node-problem-detector/pkg/custompluginmonitor/plugin"
 	cpmtypes "k8s.io/node-problem-detector/pkg/custompluginmonitor/types"
 	"k8s.io/node-problem-detector/pkg/types"
+	"k8s.io/node-problem-detector/pkg/util"
 	"k8s.io/node-problem-detector/pkg/util/tomb"
 )
 
@@ -124,10 +125,16 @@ func (c *customPluginMonitor) generateStatus(result cpmtypes.Result) *types.Stat
 		for i := range c.conditions {
 			condition := &c.conditions[i]
 			if condition.Type == result.Rule.Condition {
-				status := result.ExitStatus >= cpmtypes.NonOK
+				status := toConditionStatus(result.ExitStatus)
 				if condition.Status != status || condition.Reason != result.Rule.Reason {
 					condition.Transition = timestamp
 					condition.Message = result.Message
+					events = append(events, util.GenerateConditionChangeEvent(
+						condition.Type,
+						status,
+						result.Rule.Reason,
+						timestamp,
+					))
 				}
 				condition.Status = status
 				condition.Reason = result.Rule.Reason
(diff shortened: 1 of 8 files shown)


Random-Liu:
I feel like we must support unknown to make this feature work.

As for the event for condition change, it's good to have. However, it doesn't need too much change, let's do it?

andyxning:
@Random-Liu Agree. But I am afraid i have no enough time to review this for now. Too busy this week. Will do this in about 5days.

Random-Liu:
@andyxning I'm not sure whether we could cut a release without the unknown support. It feels wrong to me that we report problem when the script itself time out.

Today is Kubernetes code freeze, let me see how to deal with this.

config/custom-plugin-monitor.json
```diff
@@ -15,12 +15,6 @@
     }
   ],
   "rules": [
-    {
-      "type": "temporary",
-      "reason": "NTPIsDown",
```
andyxning: Since each condition is followed by an event, this can be deleted as we have added `NTPIsDown` condition. But i prefer we add another different temporary event as an example. :)

pkg/custompluginmonitor/custom_plugin_monitor.go
```diff
@@ -159,8 +177,9 @@ func initialConditions(defaults []types.Condition) []types.Condition {
 	conditions := make([]types.Condition, len(defaults))
 	copy(conditions, defaults)
 	for i := range conditions {
-		// TODO(random-liu): Validate default conditions
-		conditions[i].Status = false
+		if conditions[i].Status == "" {
```
andyxning: IIUC, by changing to this, NPD will allow user to custom the initial `status` field value.

pkg/custompluginmonitor/custom_plugin_monitor.go
```diff
@@ -143,6 +150,17 @@ func (c *customPluginMonitor) generateStatus(result cpmtypes.Result) *types.Stat
 	}
 }
 
+func toConditionStatus(s cpmtypes.Status) types.ConditionStatus {
+	switch s {
+	case cpmtypes.OK:
+		return types.True
```
andyxning: when the check script returns `OK`, the condition status should return `types.False` which indicates that the condition is `False`.

pkg/custompluginmonitor/custom_plugin_monitor.go
```diff
@@ -143,6 +150,17 @@ func (c *customPluginMonitor) generateStatus(result cpmtypes.Result) *types.Stat
...
+	case cpmtypes.NonOK:
+		return types.False
```
andyxning: `return types.True`

config/custom-plugin-monitor.json
Random-Liu: Will keep this as an example.

Random-Liu: Done

pkg/custompluginmonitor/custom_plugin_monitor.go
Random-Liu: True.

Random-Liu: Done

pkg/custompluginmonitor/custom_plugin_monitor.go
Random-Liu: Done

pkg/custompluginmonitor/custom_plugin_monitor.go
Random-Liu: Hm, let's not do that.

Random-Liu: Done

andyxning:
@Random-Liu Thanks for the updating! :)

/lgtm

PR #152, opened 2017-11-27
Network problem script

rramkumar1:
Added a script that checks for important networking events. This script will be run by the custom plugin monitor. 

Currently the script only checks if the conntrack table is full and sends a temporary event if it is. More checks will be added in the future.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/node-problem-detector/152)

### DIFF
--- a/config/plugin/network_problem.sh
+++ b/config/plugin/network_problem.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+# This plugin checks for common network issues. Currently, it only checks
+# if the conntrack table is full. 
+
+conntrack_max=$(cat /proc/sys/net/ipv4/netfilter/ip_conntrack_max)
+conntrack_count=$(cat /proc/sys/net/ipv4/netfilter/ip_conntrack_count)
+if (( conntrack_count >= conntrack_max )); then
+  echo "Conntrack table full"
+  exit 1
+fi
+
+echo "Conntrack table available"
+exit 0
+
--- a/config/plugin/network_problem.sh
+++ b/config/plugin/network_problem.sh
@@ -3,13 +3,21 @@
 # This plugin checks for common network issues. Currently, it only checks
 # if the conntrack table is full. 
 
+OK=0
+NONOK=1
+UNKNOWN=2
+
+[ -f /proc/sys/net/ipv4/netfilter/ip_conntrack_max ] || echo $UNKNOWN
+[ -f /proc/sys/net/ipv4/netfilter/ip_conntrack_count ] || echo $UNKNOWN
+
 conntrack_max=$(cat /proc/sys/net/ipv4/netfilter/ip_conntrack_max)
 conntrack_count=$(cat /proc/sys/net/ipv4/netfilter/ip_conntrack_count)
+
 if (( conntrack_count >= conntrack_max )); then
   echo "Conntrack table full"
-  exit 1
+  exit $NONOK
 fi
 
 echo "Conntrack table available"
-exit 0
+exit $OK
 
--- a/config/network-problem-monitor.json
+++ b/config/network-problem-monitor.json
@@ -1,0 +1,18 @@
+{
+  "plugin": "custom",
+  "pluginConfig": {
+    "invoke_interval": "30s",
+    "timeout": "5s",
+    "max_output_length": 80,
+    "concurrency": 3
+  },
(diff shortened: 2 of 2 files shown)


config/plugin/network_problem.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+# This plugin checks for common network issues. Currently, it only checks
+# if the conntrack table is full. 
+
```
andyxning: I would prefer we keep the plugin pattern for defining exit code. Which is illustrated in [check_ntp.sh](https://github.com/kubernetes/node-problem-detector/blob/master/config/plugin/check_ntp.sh#L6-L8)
```
OK=0
NONOK=1
UNKNOWN=2
```

config/plugin/network_problem.sh
```diff
@@ -0,0 +1,15 @@
...
+conntrack_max=$(cat /proc/sys/net/ipv4/netfilter/ip_conntrack_max)
+conntrack_count=$(cat /proc/sys/net/ipv4/netfilter/ip_conntrack_count)
+if (( conntrack_count >= conntrack_max )); then
+  echo "Conntrack table full"
+  exit 1
```
andyxning: `exit $OK`

config/plugin/network_problem.sh
```diff
@@ -0,0 +1,15 @@
...
+fi
+
+echo "Conntrack table available"
+exit 0
```
andyxning: `exit $OK`

config/plugin/network_problem.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+# This plugin checks for common network issues. Currently, it only checks
+# if the conntrack table is full. 
+
+conntrack_max=$(cat /proc/sys/net/ipv4/netfilter/ip_conntrack_max)
```
andyxning: I think we need to check the file existence firstly, otherwise if one of the two files do not exit, the check script will return error exit code.

andyxning:
@rramkumar1 @Random-Liu 

I am not sure we need to add all available custom plugins in node-problem-detector repo. But, imo, i am fine with it.

config/plugin/network_problem.sh
rramkumar1: Done

config/plugin/network_problem.sh
rramkumar1: Done

config/plugin/network_problem.sh
rramkumar1: Done

config/plugin/network_problem.sh
rramkumar1: Done

rramkumar1:
@andyxning Addressed your comments, PTAL. Also I agree that it would make sense to add all available custom plugins to the repo since they should be useful for anyone.

andyxning:
lgtm

@Random-Liu PTAL.

rramkumar1:
ping @Random-Liu

Random-Liu:
LGTM

config/plugin/network_problem.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+# This plugin checks for common network issues. Currently, it only checks
+# if the conntrack table is full. 
+
+OK=0
+NONOK=1
+UNKNOWN=2
+
+[ -f /proc/sys/net/ipv4/netfilter/ip_conntrack_max ] || echo $UNKNOWN
```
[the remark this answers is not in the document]
r4j4h: @rramkumar1 @Random-Liu  I see this is merged now and wanted to bring attention to this line just incase :+1:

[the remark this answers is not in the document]
r4j4h: I created !155 to handle it since I messed this up so poorly!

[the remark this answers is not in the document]
andyxning: > is echo $UNKNOWN intentional or should it be exit $UNKNOWN?

Apparently it should be `exit $UNKNOWN`. Sorry for not realizing this early.

dchen1107:
30s interval is too often, and I don't think any remedy system will take an action every minute. Can we increase the interval to 5 mins at least? cc/ @Random-Liu

PR #155, opened 2018-01-05
Changed unknown status echo to exit in network_problem custom plugin

r4j4h:
My comment was eaten by github in #152 and wanted to raise attention incase this was meant to be an exit instead of an echo, otherwise feel free to close!

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/node-problem-detector/155)


rramkumar1:
@r4j4h Good catch! @Random-Liu PTAL

PR #169, opened 2018-04-08
update example

zhangxiaoyu-zidif:
the original ds example could not be created.


andyxning:
Since daemonset has been promoted to `apps/v1` since Kubernetes 1.9.0, this should be fine.
/lgtm

PR #177, opened 2018-05-30
fix typo in daemonset labelSelector

rlguarino:
The label on the pod is `app: node-problem-detector` while the
label in the `labelSelector` was `name: node-problem-detector`.
This CL updates the `labelSelector` to use `app` rather than `name`.


andyxning:
/lgtm
/approve

Thanks @rlguarino for helping take care of this. 👍

PR #180, opened 2018-06-18
Add log-counter plugin written in go

dashpole:
This PR adds a new binary to the node-problem-detector repository: log-counter.
The new binary uses the kmsg log watcher to get kmsg log events, and checks the number of events that occurred.
The binary accepts command-line flags for the pattern, count, and period of time to look back.
It sets the condition `NodeRecreationRequired` when it sees the unregister_netdevice error 3 times in 20 minutes, and runs every 10 minutes.

/assign @Random-Liu

### DIFF
--- a/cmd/logcounter/log_counter.go
+++ b/cmd/logcounter/log_counter.go
@@ -1,0 +1,46 @@
+/*
+Copyright 2018 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/pflag"
+
+	"k8s.io/node-problem-detector/cmd/logcounter/options"
+	"k8s.io/node-problem-detector/pkg/custompluginmonitor/types"
+	"k8s.io/node-problem-detector/pkg/logcounter"
+)
+
+func main() {
+	fedo := options.NewLogCounterOptions()
+	fedo.AddFlags(pflag.CommandLine)
+	pflag.Parse()
+
+	counter, err := logcounter.NewKmsgLogCounter(fedo)
+	if err != nil {
+		fmt.Print(err)
+		os.Exit(int(types.Unknown))
+	}
+	actual := counter.Count()
+	if actual >= fedo.Count {
+		fmt.Printf("Found %d matching logs, which meets the threshold of %d\n", actual, fedo.Count)
+		os.Exit(int(types.NonOK))
+	}
+	os.Exit(int(types.OK))
+}
--- a/config/kernel-monitor-counter.json
+++ b/config/kernel-monitor-counter.json
@@ -1,0 +1,31 @@
+{
+  "plugin": "custom",
+  "pluginConfig": {
+    "invoke_interval": "5m",
+    "timeout": "1m",
+    "max_output_length": 80,
+    "concurrency": 1
+  },
+  "source": "kernel-monitor",
+  "conditions": [
+    {
+      "type": "FrequentUnregisterNetDevice",
+      "reason": "NoFrequentUnregisterNetDevice",
+      "message": "node is functioning properly"
+    }
+  ],
+  "rules": [
+    {
+      "type": "permanent",
+      "condition": "FrequentUnregisterNetDevice",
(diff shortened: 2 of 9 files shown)


cmd/logcounter/log_counter.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+func main() {
+	fedo := options.NewLogCounterOptions()
+	fedo.AddFlags(pflag.CommandLine)
+	pflag.Parse()
+
+	counter, err := logcounter.NewKmsgLogCounter(fedo)
+	if err != nil {
+		glog.Error(err)
```
Random-Liu: Use `fmt.Print`.

We only support `stdout` now. https://github.com/kubernetes/node-problem-detector/blob/master/pkg/custompluginmonitor/plugin/plugin.go#L112

config/kernel-monitor-counter.json
```diff
@@ -0,0 +1,26 @@
...
+  "pluginConfig": {
+    "invoke_interval": "10m",
+    "timeout": "1m",
+    "max_output_length": 80,
+    "concurrency": 1
+  },
+  "source": "kernel-monitor",
+  "conditions": [
+    {
+      "type": "NodeRecreationRequired",
```
Random-Liu: This seems to be a auto repair decision. Let's keep this as a problem description.

pkg/logcounter/log_counter.go
```diff
@@ -0,0 +1,78 @@
...
+
+type logCounter struct {
+	logCh   <-chan *systemtypes.Log
+	buffer  systemlogmonitor.LogBuffer
+	pattern string
+	clock   clock.Clock
+}
+
+func NewKmsgLogCounter(options *options.LogCounterOptions) (types.LogCounter, error) {
+	watcher := kmsg.NewKmsgWatcher(watchertypes.WatcherConfig{Lookback: options.Lookback})
```
Random-Liu: It is possible that there is no kernel log for a long time.

In that way, this plugin will always block for 1 min. Is that desired?

cmd/logcounter/log_counter.go
dashpole: should I use Printf, or should I pass --logtostderr to the binary?  I am leaning toward the second, but whatever you want works :p

pkg/logcounter/log_counter.go
dashpole: There is a timeout in the reader loop that will terminate if we don't get a new kmsg for a second.

config/kernel-monitor-counter.json
dashpole: renamed to FrequentUnregisterNetDevice

cmd/logcounter/log_counter.go
dashpole: passed --logtostderr to the binary

dashpole: changed to printf

cmd/logcounter/log_counter.go
```diff
@@ -0,0 +1,49 @@
...
+	fedo.AddFlags(pflag.CommandLine)
+	pflag.Parse()
+
+	counter, err := logcounter.NewKmsgLogCounter(fedo)
+	if err != nil {
+		fmt.Print(err)
+		os.Exit(exitUnknown)
+	}
+	if counter.Count() >= fedo.Count {
+		os.Exit(exitNotOK)
```
Random-Liu: Add message for this case?

dashpole:
can you tripple check the changes to the Makefile?  I am not really sure what the proper structure is.  This set of changes is mostly a guess.

cmd/logcounter/log_counter.go
dashpole: done

config/kernel-monitor-counter.json
```diff
@@ -0,0 +1,31 @@
...
+      "reason": "NoFrequentUnregisterNetDevice",
+      "message": "node is functioning properly"
+    }
+  ],
+  "rules": [
+    {
+      "type": "permanent",
+      "condition": "FrequentUnregisterNetDevice",
+      "reason": "UnregisterNetDevice",
+      "path": "/home/kubernetes/bin/log-counter",
```
dashpole: I couldn't get this working without an absolute path.  Any ideas on what the path should be?

dashpole:
This is working now!

config/kernel-monitor-counter.json
Random-Liu: We can use `https://stackoverflow.com/questions/18537257/how-to-get-the-directory-of-the-currently-running-file`. And document that the path is relative to `node-problem-detector` binary.

It is also fine to enforce absolute path, but we should validate it inside NPD.

dashpole: it didn't seem like it was relative to the npd binary...

dashpole: We do validate the path inside NPD, so it should be alright...

Random-Liu: > it didn't seem like it was relative to the npd binary...

I mean we can make it so.

Another option is to add a config to config the script root directory. Anyway, I'm fine with starting from this for now.

PR #186, opened 2018-07-08
make VERSION value more flexible

### LINKED ISSUE #179: make ./bin/node-problem-detector expects a git repository, hinders builds from releases
The `make ./bin/node-problem-detector` [command](https://github.com/kubernetes/node-problem-detector/blob/2f915ecf29dcb3370611f51bbd8297c9180109bc/Makefile#L80)  uses the VERSION [variable](https://github.com/kubernetes/node-problem-detector/blob/2f915ecf29dcb3370611f51bbd8297c9180109bc/Makefile#L22)

Version variable executes a git command that expects a repo. 

If you try to build after extracting the tarfile from a release, the code will not have a .git directory and  build fails with 

```+ make ./bin/node-problem-detector
fatal: Not a git repository (or any of the parent directories): .git```

One could use TAG in the make ./bin/node-problem-detector rule or one could make VERSION optional too with 

`VERSION?=$(shell git describe --tags --dirty)`

### PR DESCRIPTION
**Pull request by andyxning**
Fix #179 

This PR will make the `VERSION` value in Makefile more flexible and the build output less error info log.
It will first check whether this is a git repo by checking whether a `.git` directory exists or not. If it is a git repo then `VERSION` value will be `git describe --tags --dirty`, otherwise it will use a default value `UNKNOWN`

### CODE REVIEW

**Comment by andyxning** [comment]
/cc @Random-Liu 
@ahakanbaba

File: Makefile [hunk 1]
```diff
@@ -19,7 +19,7 @@
 all: build
 
 # VERSION is the version of the binary.
-VERSION:=$(shell git describe --tags --dirty)
+VERSION:=$(shell if [ -d .git ]; then echo `git describe --tags --dirty`; else echo "UNKNOWN"; fi)
```
Random-Liu: I think we can make this optional. Silently return `UNKNOWN` doesn't seem good as well.

Making it optional, so that people build it can specify a string themselves.

andyxning: Good suggestion. Will do.

andyxning: Done

PR #188, opened 2018-07-09
support go 1.10 in travis ci

andyxning:
Since Golang 1.10 has been released for a long time, it should be tested in travis ci. :)

/cc @Random-Liu

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,8 @@ sudo: required
 dist: trusty
 language: go
 go:
- - 1.8
  - 1.9
+ - "1.10"
 services:
  - docker
 before_install:


.travis.yml
```diff
@@ -4,8 +4,8 @@ sudo: required
 dist: trusty
 language: go
 go:
- - 1.8
  - 1.9
+ - "1.10"
```
Random-Liu: nit: Remove quote for code style consistency?

I'm fine with doing this in another PR.

andyxning: yaml can not parse `1.10` correctly without any quote. :)
FYI：https://twitter.com/davecheney/status/964669282855485440

PR #193, opened 2018-07-17
Adding missing CONTRIBUTING.md

### LINKED ISSUE #190: Missing CONTRIBUTING.md file
All K8s subrepositories should have a CONTRIBUTING.md file, which at the minimum should point to https://github.com/kubernetes/community/blob/master/contributors/guide/README.md. Care should be taken that all information is in sync with the contributor guide.

Subrepositories may also have contributing guidelines specific to that repository. They should be explicitly documented and explained in the CONTRIBUTING.md

Ref:  https://github.com/kubernetes/community/issues/1832

poothia:
It adds contribution guidelines for node-problem-detector
fixes #190

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,52 @@
+# Contributing guidelines
+
+The Kubernetes node-problem-detector project accepts contributions via GitHub pull requests. This document outlines the process to help get your contribution accepted.
+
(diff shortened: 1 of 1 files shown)


ghanshyammann:
we should have a standard format and content of contributor guide across projects but it is ok as of now because this commit adding good doc for new contributor and it can be standardize later.

PR #203, opened 2018-09-10
fix custom plugin monitor condition change

andyxning:
Fix #202 

/cc @aga20

PR #205, opened 2018-09-12
Always update the message field of custom conditions

zadoryzsolt:


andyxning:
/close 

Close this via #203

PR #210, opened 2018-10-09
Fix error path for plugin

duffqiu:
remove the "."


andyxning:
@duffqiu This is correct, IMHO. The relative path usage is because that the NPD process is running in the root directory.

Actually, it is consistent with the path settings in https://github.com/kubernetes/node-problem-detector/blob/master/config/custom-plugin-monitor.json. You can adjust the path directory before you use NPD custom plugin in the production env.

PR #221, opened 2018-11-20
Detect readonly filesystem

wangzhen127:
The PR adds one more rule to kernel monitor to detect read-only filesystem. It also changes the plugin from journald to kmsg, because when filessystem becomes readonly, no new logs can be saved into `/var/log`. Then journald does not have new data for problem detection.

### DIFF
--- a/config/kernel-monitor.json
+++ b/config/kernel-monitor.json
@@ -1,9 +1,6 @@
 {
-	"plugin": "journald",
-	"pluginConfig": {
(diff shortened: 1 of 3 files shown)


wangzhen127:
@andyxning, the reading part is fine. The problem is that when fs is read-only, no new logs can be saved into `/var/log`. Then journald does not have new data for problem detection. Updated the PR description to explain that part.

Issue #202, opened 2018-09-05
Reason change once and it stay in wrong state with custom plugins


aga20:
Hello,

I found something again I think but maybe I follwed wrong way. 
I'm using a custom config based on the ntp example: 

```
{
    "plugin": "custom",
    "pluginConfig": {
        "invoke_interval": "30s",
        "timeout": "5s",
        "max_output_length": 80,
        "concurrency": 3
    },
    "source": "ntp-custom-plugin-monitor",
    "conditions": [
        {
            "type": "CustomProblem",
            "reason": "CustomIsUp",
            "message": "Status of the custom service"
        }
    ],
    "rules": [
        {
            "type": "permanent",
            "condition": "CustomProblem",
            "reason": "CustomIsDown",
            "path": "/usr/bin/custom.sh",
            "timeout": "3s"
        }
    ]
}

```
The /usr/bin/custom.sh script is very simple: exit with 0 or 1. 

So when the node problem detector start it set the condition:

```
Conditions:
  Type             Status  LastHeartbeatTime                 LastTransitionTime                Reason                       Message
  ----             ------  -----------------                 ------------------                ------                       -------
  CustomProblem    False   Wed, 05 Sep 2018 14:40:26 +0200   Wed, 05 Sep 2018 14:40:25 +0200   CustomIsUp                   Status of the custom service
  OutOfDisk        False   Wed, 05 Sep 2018 14:40:23 +0200   Thu, 30 Aug 2018 17:16:47 +0200   KubeletHasSufficientDisk     kubelet has sufficient disk space available
  MemoryPressure   False   Wed, 05 Sep 2018 14:40:23 +0200   Thu, 30 Aug 2018 17:16:47 +0200   KubeletHasSufficientMemory   kubelet has sufficient memory available
  DiskPressure     False   Wed, 05 Sep 2018 14:40:23 +0200   Wed, 05 Sep 2018 13:17:06 +0200   KubeletHasNoDiskPressure     kubelet has no disk pressure
  PIDPressure      False   Wed, 05 Sep 2018 14:40:23 +0200   Thu, 30 Aug 2018 17:16:47 +0200   KubeletHasSufficientPID      kubelet has sufficient PID available
  Ready            True    Wed, 05 Sep 2018 14:40:23 +0200   Thu, 30 Aug 2018 17:35:04 +0200   KubeletReady                 kubelet is posting ready status

```
[...]

Issue #200, opened 2018-09-03
Condition message doesn't update when the status is same


aga20:
Hello,

we want to use the node problem detector  with custom plugins in our env where we have an internally developed alarm handling solution. We will implement a layer between kubernetes an our alarm handling solution which can raise alarm based on conditions and events. 
First we should check the free spaces on the nodes' partitions but we have a lot of node types with another volumes and partions (storage, etc...) so we can't able cover all node with one config (the npd run as a daemon set so the per node config is a little bit hard). We try to figure out a general solution and our final idea is next: we use only one general config and run one script which check the partitions/volume on the nodes. If one partition is full it exit with code 1 and print out the useful information for example /mnt/part 1 full on the node. This is ok when condition's status is change between True/False. But we want to update the condition message if an another partition for example part2 is full but the contdition status didn't change back to false yet (so the part1 is still full). As I see until the status is same you don't update the message [here](https://github.com/kubernetes/node-problem-detector/blob/master/pkg/custompluginmonitor/custom_plugin_monitor.go#L129).
I want to know this is the expected behaviour and what is the reason or we can change this behaviour? For us it can help a lot. 

Thanks!

aga20:
@andyxning Could you comment it? Thanks!

andyxning:
@aga20 
> But we want to update the condition message if an another partition for example part2 is full but the contdition status didn't change back to false yet (so the part1 is still full). As I see until the status is same you don't update the message 

This is mainly used to decrease the updates that needs to be done in between NPD and apiserver.

andyxning:
IIUC, you mean that you need to recognize which partition to check and which partition is run out of space. You need to make the condition message structured in order to parse it easily and correctly.


Actually you can change the behavior. But what you should note is that this may make some pressure on your apiserver.

aga20:
Hello,

yes we need a structured message of course but we want to use the message field to keep update our alarms. For example during a condition's status is "true" we should know if another partion is full. So ok, during a run I see 2 partition is full, I print out in a structured message. The sysadmin recognise the alarm and start to investigate the problem. The condition still in "true" state and the script run again and see another 3rd partition is full, print out the new message but the npd doesn't update the condition's message and we missed the new information. After a little the original 2 pratition is ok because someone made some free space but the condition is still "true" and the message stay same (what represent the two partition not the 3rd) because as the script report the 3rd partition is still full. 
This is why we need an updated message field if this is possible. 
[...]

PR #228, opened 2018-11-27
More fix to custom plugin monitor condition change

wangzhen127:
When `condition.Status` does not change, we should check `condition.Reason` instead of `condition.Message` because Message could be spammy and change a lot. 

This is part of #202.


andyxning:
@wangzhen127 @Random-Liu Please see https://github.com/kubernetes/node-problem-detector/issues/200, end users may need a switch to enable condition change when the need to bring some info through message.

PR #231, opened 2018-11-28
Update base image for CVE fix.

Random-Liu:
There are several CVEs in `k8s.gcr.io/debian-base-amd64:0.3`.
image

Update the base image to fix those CVEs.
image


Random-Liu:
I'll cut a `0.6.1` release once this is merged. Once this is done, we can promote the image to `k8s.gcr.io`.

PR #238, opened 2018-12-10
enable codnition updaet when message change for custom plugin

### LINKED ISSUE #200: Condition message doesn't update when the status is same

andyxning:
Fix #200 

This PR adds a custom global config to control whether NPD should update condition status when message changes.

### DIFF
--- a/pkg/custompluginmonitor/custom_plugin_monitor.go
+++ b/pkg/custompluginmonitor/custom_plugin_monitor.go
@@ -176,8 +176,12 @@ func (c *customPluginMonitor) generateStatus(result cpmtypes.Result) *types.Stat
 
 					condition.Status = status
 					condition.Reason = result.Rule.Reason
-				} else if condition.Status == status && condition.Reason != result.Rule.Reason {
-					// change 4: Condition status do not change. condition reason changes.
+				} else if condition.Status == status &&
+					(condition.Reason != result.Rule.Reason ||
+						(*c.config.PluginGlobalConfig.EnableMessageChangeBasedConditionUpdate && condition.Message != result.Message)) {
+					// change 4: Condition status do not change.
+					// condition reason changes or
+					// condition message changes when message based condition update is enabled.
 					condition.Transition = timestamp
 					condition.Reason = result.Rule.Reason
 					condition.Message = result.Message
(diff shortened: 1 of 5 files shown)


pkg/custompluginmonitor/custom_plugin_monitor.go
```diff
@@ -176,7 +176,7 @@ func (c *customPluginMonitor) generateStatus(result cpmtypes.Result) *types.Stat
 
 					condition.Status = status
 					condition.Reason = result.Rule.Reason
-				} else if condition.Status == status && condition.Reason != result.Rule.Reason {
+				} else if condition.Status == status && (condition.Reason != result.Rule.Reason || (*c.config.PluginGlobalConfig.EnableMessageChangeBasedConditionUpdate && condition.Message != result.Message)) {
```
wangzhen127: The comment on line 180 needs update.

Can you also update the custom plugin readme to include this new option?

andyxning: @wangzhen127 Thanks for the tips. Will do.

andyxning: @wangzhen127 Done. PTAL.

Random-Liu: This line is way too long... can we break it into 2 lines at some point?

andyxning:
@Random-Liu Done. PTAL.

PR #245, opened 2018-12-25
add resource

### LINKED ISSUE #240: Add Resource Limits to the node-problem-detector-pod
Please add resource limits to this project to ensure stability!

andyxning:
Fix #240 

The requests and limits are only references. They have no requirements. End users can adjust this themselves. 

/cc @Random-Liu @wangzhen127

### DIFF
--- a/deployment/node-problem-detector.yaml
+++ b/deployment/node-problem-detector.yaml
@@ -18,6 +18,13 @@ spec:
         - --logtostderr
         - --system-log-monitors=/config/kernel-monitor.json,/config/docker-monitor.json
         image: k8s.gcr.io/node-problem-detector:v0.6.1
+        resources:
+          limits:
+            cpu: 10m
+            memory: 80Mi
+          requests:
+            cpu: 10m
+            memory: 80Mi
         imagePullPolicy: Always
         securityContext:
           privileged: true


deployment/node-problem-detector.yaml
```diff
@@ -18,6 +18,13 @@ spec:
         - --logtostderr
         - --system-log-monitors=/config/kernel-monitor.json,/config/docker-monitor.json
         image: k8s.gcr.io/node-problem-detector:v0.5.0
+        resources:
+          limits:
+            cpu: 300m
+            memory: 100Mi
+          requests:
+            cpu: 300m
+            memory: 100Mi
```
wangzhen127: I tested v0.6.0 that the resource consumption is:
- CPU: 2m
- MEM: 30Mi

Maybe we should lower the numbers to something smaller, like cpu 10m and memory 50Mi to better reflect the resource consumption of NPD? WDYT?

andyxning: According to your test, it should be ok to run NPD with cpu 10m. I suggest we add more memory to avoid OOM. @wangzhen127

andyxning: Finanlly, cpu requests/limits is 10m, memory requests/limits is 80Mi.

andyxning:
@wangzhen127 Rebased. PTAL.

PR #253, opened 2019-01-30
Empty LogPath will use journald's default path.

stribb:
At present, the empty LogPath will use the _code's_ default journal path, without regard for the _host's_ default journal path.

This code changes the default to use the host's path.


stribb:
CLA signed.

stribb:
Is there some way I can make this simpler to review?

Thanks

stribb:
Forgive my ignorance, but how long should I expect to wait before this PR is reviewed?

Thanks

wangzhen127:
Sorry for the delay. I will take a look today. Feel free to ping me on slack if you don't see any response after a day or two.

pkg/systemlogmonitor/logwatchers/journald/log_watcher.go
```diff
@@ -135,27 +135,28 @@ func (j *journaldWatcher) watchLoop() {
...
 	// Get journal log path.
-	path := defaultJournalLogPath
-	if cfg.LogPath != "" {
-		path = cfg.LogPath
-	}
-	// If the path doesn't present, NewJournalFromDir will create it instead of
-	// returning error. So check the path existence ourselves.
-	if _, err := os.Stat(path); err != nil {
-		return nil, fmt.Errorf("failed to stat the log path %q: %v", path, err)
+	path := ""
```
wangzhen127: The value of `path` seems never updated?

pkg/systemlogmonitor/logwatchers/journald/log_watcher.go
```diff
@@ -135,27 +135,28 @@ func (j *journaldWatcher) watchLoop() {
...
-
 	// configSourceKey is the key of source configuration in the plugin configuration.
 	configSourceKey = "source"
 )
 
 // getJournal returns a journal client.
 func getJournal(cfg types.WatcherConfig, startTime time.Time) (*sdjournal.Journal, error) {
 	// Get journal log path.
-	path := defaultJournalLogPath
-	if cfg.LogPath != "" {
```
wangzhen127: I think the idea here is that if someone doesn't want to use `/var/log/journal`, he can just set `cfg.LogPath`. Can you explain more on your use case and why current approach is not enough?

stribb: Some of our nodes have different log paths from others. In both cases, they use the log path that is returned by NewJournal().

It would seem better to me to fall back to journald's configured path if path is not specified, rather than hard-code a default.

wangzhen127: That's fair.

pkg/systemlogmonitor/logwatchers/journald/log_watcher.go
```diff
@@ -135,27 +135,28 @@ func (j *journaldWatcher) watchLoop() {
...
+	} else {
+		// If the path doesn't exist, NewJournalFromDir will
+		// create it instead of returning error. So check the
+		// path existence ourselves.
+		if _, err = os.Stat(cfg.LogPath); err != nil {
+			return nil, fmt.Errorf("failed to stat the log path %q: %v", path, err)
+		}
+		// Get journal client from the log path.
+		journal, err = sdjournal.NewJournalFromDir(cfg.LogPath)
 	}
```
wangzhen127: Now we fall back to system default path, can you print a message here on what path it is used?

andyxning: +1 for the comment above. @stribb Do you have tiem to update this?

stribb: I agree it would be nice. But there's no API to find that out as far as I can tell.

https://www.freedesktop.org/software/systemd/man/sd_journal_open.html#

stribb:
@wangzhen127 Sorry, I'm new to GitHub PRs. Does my PR look reasonable or did I mess up? :)

Thanks!

stribb:
Thanks, I've squashed all the commits. Sorry for the delay: I was flying to Barcelona.

PR #264, opened 2019-03-30
Support building NPD images in different arch

yuwenma:
Test: `make` generates a .tar.gz file and the exec file node-problem-detector can be upstarted.


yuwenma:
/assign @Random-Liu 

BTW, can I get a ok-to-test label?

PR #267, opened 2019-04-14
add labels to daemonset metadata

ringtail:
Add labels to Daemonset metadata.

### DIFF
--- a/deployment/node-problem-detector.yaml
+++ b/deployment/node-problem-detector.yaml
@@ -2,6 +2,8 @@ apiVersion: apps/v1
 kind: DaemonSet
 metadata:
   name: node-problem-detector
+  labels:
+    app: node-problem-detector
 spec:
   selector:
     matchLabels:


deployment/node-problem-detector.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: apps/v1
 kind: DaemonSet
 metadata:
   name: node-problem-detector
+  labels:
+    app: node-problem-detector
```
wangzhen127: nit: s/app/k8s-app

ringtail:
@wangzhen127 Hmm. I don't think it is caused by this PR. Could you rerun the test?
```
--- FAIL: TestGoroutineLeak (0.00s)
assertions.go:199: Error Trace:	log_watcher_test.go:194
Error:	Not equal: 9 (expected)
!= 8 (actual)
```

deployment/node-problem-detector.yaml
ringtail: I found the label below is app: node-problem-detector. So I think app: node-problem-detector may be more suitable

wangzhen127: Ah, forgot about that. lgtm

Issue #284, opened 2019-06-03
Implement NPD metrics mode


xueweiz:
Here is the [proposal](https://docs.google.com/document/d/1SeaUz6kBavI283Dq8GBpoEUDrHA2a795xtw0OvjM568/edit#heading=h.xw2rl6ej7zhr).

This issue tracks the implementation of the proposal:
- [x] Check-in the proposal to NPD project's [README](https://github.com/kubernetes/node-problem-detector/tree/v0.6.3#links)   (#285)
- [x] Refactor problem daemon registration and initialization, so that they are more modular, and rely on less [per-daemon initialization code duplication](https://github.com/kubernetes/node-problem-detector/blob/4880be842d32dfcf31d2a8f01070d17473732a88/cmd/node_problem_detector.go#L73).  (#275)
- [x] Refactor events/condition reporting into k8s_exporter. (#275)
- [x] Add a SystemStatsMonitor problem daemon, to monitor some system metrics; and add an optional Prometheus metrics exporter. (#286 #275)
- [x] Implement feature in System Log Monitor and Custom Plugin Monitor, to allow them to report metrics. (#300 #315 )
- [x] Implement option for disabling problem daemons at compile time. (#294 )
- [x] ~~Implement option for disabling exporters at compile time.~~ (Current exporters includes a k8s exporter and Prometheus exporter. Since they are commonly used, compile time disabling is not high priority for now. We can re-prioritize if the situation changes.)

xueweiz:
Now that #315 is merged, this issue is finally complete! It will be release in NPD v0.7.0

Huge thanks to Zhen, Lantao and Andy for the reviews and feedback!

There are a few follow up improvements, most importantly, e2e tests for NPD (#296). I really want e2e test to happen, so that we can have more confidence when changing NPD code.

PR #275, opened 2019-05-14
node-problem-detector: report disk queue length in Prometheus format

xueweiz:
This change is to address below items from #284 :

- [x] Refactor problem daemon registration and initialization, so that they are more modular, and rely on less per-daemon initialization code duplication.
- [x] Refactor events/condition reporting into k8s_exporter.
- [x] Add a SystemStatsMonitor problem daemon, to monitor some system metrics; and add an optional Prometheus metrics exporter.

Since this PR will need OpenCensus and gopsutil library, I submitted PR #289 for it.

Deprecated (but not removed yet) flags in this PR:

- --custom-plugin-monitors
- --system-log-monitors

New flags introduced in this PR:

- --config.custom-plugin-monitor
- --config.system-log-monitor
- --config.system-stats-monitor
- --enable-k8s-exporter
- --prometheus-address
- --prometheus-port

Test steps:

1. Verifying existing command works:
> make ./bin/node-problem-detector && ./bin/node-problem-detector --system-log-monitors=config/kernel-monitor.json --apiserver-override=http://127.0.0.1:8080?inClusterConfig=false --logtostderr

2. Testing new flags for existing behavior:
> make ./bin/node-problem-detector && ./bin/node-problem-detector --config.system-log-monitor=config/kernel-monitor.json --apiserver-override=http://127.0.0.1:8080?inClusterConfig=false --logtostderr

3. Test NPD for disk monitor with no k8s_exporter
> make ./bin/node-problem-detector && ./bin/node-problem-detector --config.system-log-monitor=config/kernel-monitor.json --apiserver-override=http://127.0.0.1:8080?inClusterConfig=false --logtostderr --config.system-stats-monitor=config/system-stats-monitor.json --enable-k8s-exporter=false

> curl http://localhost:20257/metrics

Follow up:
I will later update test infra to use these new flags.

PR #286, opened 2019-06-03
Update dependencies in Godep for #275

xueweiz:
As part of #284 , in #275 , we adds in new dependencies on gopsutil and OpenCensus.
This PR updates Godep so that NPD can build correctly.

I separated these code out from #275 , so that #275  can focus on real code changes, rather than dependency management.


andyxning:
/lgtm
/approve

@xueweiz You need to force push in #275.

PR #287, opened 2019-06-04
Migrate from godep to modules and bump versions

wangzhen127:
This PR migrates the dependency management system from godep to modules and bumps the package versions.

Kubernetes is now using modules. See this k8s [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-architecture/2019-03-19-go-modules.md). NPD is still using vendoring with modules.

The side effect of using golang modules is that it requires go 1.11+ to build node-problem-detector afterwards.


wangzhen127:
Still work in progress.

wangzhen127:
This is ready for review.
/assign @andyxning 
/assign @Random-Liu

Makefile
```diff
@@ -69,7 +69,7 @@ ifeq ($(ENABLE_JOURNALD), 1)
 endif
 
 vet:
-	go list ./... | grep -v "./vendor/*" | xargs go vet $(BUILD_TAGS)
+	GO111MODULE=off go list ./... | grep -v "./vendor/*" | GO111MODULE=off xargs go vet $(BUILD_TAGS)
```
andyxning: Why is this `off`?

wangzhen127: Updated

wangzhen127:
Note that I didn't bump versions of the following:
```
	k8s.io/api v0.0.0-20180127130940-acf347b865f2
	k8s.io/apimachinery v0.0.0-20180126010752-19e3f5aa3adc
	k8s.io/client-go v0.0.0-20180103015815-9389c055a838
	k8s.io/heapster v0.0.0-20180704153620-b25f8a16208f
```

I tried to bump their versions to latest. But it seems that `make test` will complain. I dig deeper and it is actually `go vet`'s problem. Somehow `go vet` tries to check the imported vendor code, which I thought it should not do. I spent some time on it, but didn't figure out a workaround. I decide it does not worth more time to bump those packages, because they work fine now. We can look into them later when needed.

andyxning:
Seems flakes.
/test pull-npd-test

PR #288, opened 2019-06-04
Run custom plugins immediately on startup

yguo0905:
`time.Ticker` doesn't fire instantly so the custom plugins are not run immediately on NPD startup. Some plugins may be configured with long pull interval.

We'd like to see their first run when the cluster is just created, so changing it accordingly.

Following the advice on https://github.com/golang/go/issues/17601#issuecomment-319109006.

/assign @Random-Liu 
/assign @wangzhen127

### DIFF
--- a/pkg/custompluginmonitor/plugin/plugin.go
+++ b/pkg/custompluginmonitor/plugin/plugin.go
@@ -53,46 +53,62 @@ func (p *Plugin) GetResultChan() <-chan cpmtypes.Result {
 }
 
 func (p *Plugin) Run() {
+	defer func() {
+		glog.Info("Stopping plugin execution")
+		p.tomb.Done()
+	}()
+
 	runTicker := time.NewTicker(*p.config.PluginGlobalConfig.InvokeInterval)
+	defer runTicker.Stop()
+
+	runner := func() {
+		glog.Info("Start to run custom plugins")
+
+		for _, rule := range p.config.Rules {
+			p.syncChan <- struct{}{}
+			p.Add(1)
+
+			go func(rule *cpmtypes.CustomRule) {
+				defer p.Done()
+				defer func() {
+					<-p.syncChan
+				}()
+
+				start := time.Now()
+				exitStatus, message := p.run(*rule)
+				end := time.Now()
+
+				glog.V(3).Infof("Rule: %+v. Start time: %v. End time: %v. Duration: %v", rule, start, end, end.Sub(start))
+
+				result := cpmtypes.Result{
+					Rule:       rule,
+					ExitStatus: exitStatus,
+					Message:    message,
+				}
+
+				p.resultChan <- result
+
+				glog.Infof("Add check result %+v for rule %+v", result, rule)
+			}(rule)
+		}
+
+		p.Wait()
+		glog.Info("Finish running custom plugins")
+	}
+
+	select {
+	case <-p.tomb.Stopping():
+		return
+	default:
+		runner()
+	}
 
 	for {
 		select {
 		case <-runTicker.C:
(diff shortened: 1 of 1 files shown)


pkg/custompluginmonitor/plugin/plugin.go
```diff
@@ -55,41 +55,53 @@ func (p *Plugin) GetResultChan() <-chan cpmtypes.Result {
 func (p *Plugin) Run() {
 	runTicker := time.NewTicker(*p.config.PluginGlobalConfig.InvokeInterval)
```
xueweiz: I think it's good to also clean up this code a little bit.
We could add a `defer runTicker.Stop()` line after the initialization of runTicker:
`runTicker := time.NewTicker(*p.config.PluginGlobalConfig.InvokeInterval)`

pkg/custompluginmonitor/plugin/plugin.go
```diff
@@ -55,41 +55,53 @@ func (p *Plugin) GetResultChan() <-chan cpmtypes.Result {
...
+		glog.Info("Finish running custom plugins")
+	}
 
-					glog.Infof("Add check result %+v for rule %+v", result, rule)
-				}(rule)
-			}
+	select {
+	case <-p.tomb.Stopping():
+		glog.Info("Stopping plugin execution")
+		p.tomb.Done()
```
xueweiz: Perhaps it's better to just move this one `p.tomb.Done()` and the one below to the top, as a `defer p.tomb.Done()`

pkg/custompluginmonitor/plugin/plugin.go
```diff
@@ -55,41 +55,53 @@ func (p *Plugin) GetResultChan() <-chan cpmtypes.Result {
...
 
-			p.Wait()
-			glog.Info("Finish running custom plugins")
+	for {
+		select {
+		case <-runTicker.C:
+			runner()
 		case <-p.tomb.Stopping():
 			glog.Info("Stopping plugin execution")
 			p.tomb.Done()
```
xueweiz: This is the second `p.tomb.Done()` I'd prefer to be `defer`-ed

pkg/custompluginmonitor/plugin/plugin.go
yguo0905: Good catch.

pkg/custompluginmonitor/plugin/plugin.go
yguo0905: I don't see `Stop()` is invoked anywhere. Do we support graceful shutdown?

But I changed it to use defer.

xueweiz: Seems like `tomb.Stop()` is called at here:
https://github.com/kubernetes/node-problem-detector/blame/v0.6.3/pkg/custompluginmonitor/custom_plugin_monitor.go#L84

+cc @andyxning for the context, as he wrote the code.

xueweiz:
/lgtm
/approve

This PR looks good to me. Although I probably don't have permission to approve. This is just FTR.

pkg/custompluginmonitor/plugin/plugin.go
yguo0905: I meant the `customPluginMonitor`'s `Stop()` doesn't seem to be used anywhere

andyxning: Sorry for the late response. Actually `customPluginMonitor`'s `Stop()` is not invoked anywhere. It is just to meet the requirement about implementing `Monitor` interface. 

`Stop` should be called once `p.Run` errors and is about to exit. @yguo0905 Could you take a look at this?

PR #289, opened 2019-06-06
Add OpenCensus and gopsutil as dependency

xueweiz:
These two libraries are needed for #275  to compile.

The changes were made & merged first at #286 . But since then, NPD has switched from Godep to Go Modules (#287 ). This change is re-adding these libraries using Go Modules.

Test: rebase #275  on top of this, and then #275  compiles correctly.


xueweiz:
This is also part of #284

go.sum
```diff
@@ -131,35 +164,51 @@ golang.org/x/sys v0.0.0-20180905080454-ebe1bf3edb33/go.mod h1:STP8DvDyc/dI5b8T5h
...
+gopkg.in/yaml.v1 v1.0.0-20140924161607-9f9df34309c0/go.mod h1:WDnlLJ4WF5VGsH/HVa3CI79GS0ol3YnhVnKP89i0kNg=
 gopkg.in/yaml.v2 v2.2.1 h1:mUhvW9EsL+naU5Q3cakzfE91YhliOondGd6ZrsDBHQE=
 gopkg.in/yaml.v2 v2.2.1/go.mod h1:hI93XBmqTisBFMUTm0b8Fm+jr3Dg1NNxqwp+5A1VGuI=
 honnef.co/go/tools v0.0.0-20190102054323-c2f93a96b099/go.mod h1:rf3lG4BRIbNafJWhAfAdb/ePZxsR/4RtNHQocxwk9r4=
 k8s.io/api v0.0.0-20180127130940-acf347b865f2 h1:tKttsdYar5HhU8Ps0R5lelrd4YTZ84urwTOg9q4JcEk=
 k8s.io/api v0.0.0-20180127130940-acf347b865f2/go.mod h1:iuAfoD4hCxJ8Onx9kaTIt30j7jUFS00AXQi6QMi99vA=
 k8s.io/apimachinery v0.0.0-20180126010752-19e3f5aa3adc h1:+YMPfU/2f9LMz0Sy68+JodreTTgo9PNSpkLw3PakNbs=
 k8s.io/apimachinery v0.0.0-20180126010752-19e3f5aa3adc/go.mod h1:ccL7Eh7zubPUSh9A3USN90/OzHNSVN6zxzde07TDCL0=
-k8s.io/client-go v0.0.0-20180103015815-9389c055a838 h1:3eHQsqKEOZ7zBLitCyNCfQkpyE996hTv3I4qAbXzhHU=
+k8s.io/client-go v0.0.0-20180103015815-9389c055a838 h1:gguf74OUD2sGHanFJxot0PxbBMmldZeNwnbe5i1SMm0=
```
xueweiz: There are a couple of these checksum changes in the PR.

The change is not caused by any code changes/upgrades.
This is actually a `go mod` [bug](https://github.com/golang/go/issues/29278) in pre-go 1.11.4, and the [workaround](https://github.com/golang/go/issues/29278#issuecomment-447540234) is to manually edit the go.sum checksums.

andyxning:
> The changes were made & merged first at #286 . But since then, NPD has switched from Godep to Go Modules (#287 ). This change is re-adding these libraries using Go Modules.

If go modules are added after #286, then it these packages should be transformed to go modules. Anyway, let's ensure that they are added by go modules.

@xueweiz Thanks for the job.

go.sum
andyxning: Thanks for the deep diagnose. 

I am not quite sure about this. But, if golang module does not complain about this when building or testing, we can just let it go since its a golang bug.

xueweiz:
Thanks so much for the prompt review @andyxning !

> If go modules are added after #286, then these packages should be transformed to go modules.
Yeah, but the problem is that as part of #287 , the unused libraries (i.e. included in /vendor, but not imported any where) got trimmed.
The code that imports OpenCensus and gopsutil are still in the review queue. So these two libraries got identified as unused libraries, and got deleted :)

go.sum
xueweiz: Sorry I didn't explain myself clearly :P

The symptom of this golang bug happens when developer manually runs `go mod vendor && go mod download`, as according to NPD's [README](https://github.com/kubernetes/node-problem-detector/blob/master/README.md#dependency-management) and the [go modules doc](https://github.com/golang/go/wiki/Modules#daily-workflow).

Running `go mod vendor` is supposed to update the vendor directory and insert the newly added dependencies. But the command will fail with error message like (same as [here](https://github.com/golang/go/issues/29278#issue-391337254)):

> go: verifying k8s.io/client-go@v0.0.0-20180103015815-9389c055a838: checksum mismatch
downloaded: h1:gguf74OUD2sGHanFJxot0PxbBMmldZeNwnbe5i1SMm0=
go.sum: h1:3eHQsqKEOZ7zBLitCyNCfQkpyE996hTv3I4qAbXzhHU=

And then the command will just fail without doing anything.

To make it proceed, [one must manually edit the go.sum checksums](https://github.com/golang/go/issues/29278#issuecomment-447540234).

andyxning:
> The code that imports OpenCensus and gopsutil are still in the review queue. So these two libraries got identified as unused libraries, and got deleted :)

You reminds me. :)

PR #290, opened 2019-06-06
Remove TestGoroutineLeak unit tests.

xueweiz:
We are seeing some flakes on these tests because some goroutine fluctuation:
https://github.com/kubernetes/node-problem-detector/pull/275#issuecomment-499306727

Removing the tests, as it's most robust to test leakage in a soak/stress test, rather than unit test.

### DIFF
--- a/pkg/systemlogmonitor/log_monitor_test.go
+++ b/pkg/systemlogmonitor/log_monitor_test.go
@@ -17,15 +17,10 @@ limitations under the License.
 package systemlogmonitor
 
 import (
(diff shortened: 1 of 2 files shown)


xueweiz:
Seems Lantao wrote the test, so:
/assign @Random-Liu

Random-Liu:
The test doesn't make sense. Not sure why someone (me...) added it.

xueweiz:
Thanks Lantao!

Issue #291, opened 2019-06-07
Add system component health monitor


wangzhen127:
This issue tracks the work that adds system component health monitor.

This monitor will monitor critical system component and report metrics. This will be based on the [metrics mode](https://docs.google.com/document/d/1SeaUz6kBavI283Dq8GBpoEUDrHA2a795xtw0OvjM568/). What it does include:
- [x] Count how many times the component has been started/restarted.
- [ ] Run health check command to determine if the system component is in responsive. Count how many times the component is non-responsive after timeout.
- [ ] Provide optional and configurable remedy method for the component.
- [ ] Additional health check types can be added.

The initial targets include: kubelet, docker, containerd and NPD itself. One can think about this as a NPD version that could replace [health-monitor.sh](https://github.com/kubernetes/kubernetes/blob/release-1.15/cluster/gce/gci/health-monitor.sh) for GKE. But this monitor is not limited to those.

So far, the design has been considering NPD running in standalone mode only.

wangzhen127:
#426 adds the health-check monitor.

/assign @abansal4032
/cc @yguo0905

PR #292, opened 2019-06-13
Add systemd monitor for kubelet, docker, and containerd restart events

wangzhen127:
This PR adds systemd monitor for kubelet, docker, and containerd restart events. This is part of https://github.com/kubernetes/node-problem-detector/issues/291.

This PR also enables using custom flags for `build.sh`, so that we can include the systemd monitor in the test jobs later.

To test locally:
1. Run the following in one terminal:
```
./bin/node-problem-detector --system-log-monitors=config/systemd-monitor.json --apiserver-override=http://127.0.0.1:8080?inClusterConfig=false --logtostderr
```
2. In a separate terminal, inject systemd logs.
```
echo 'Started Kubernetes kubelet.' | systemd-cat -t systemd
echo 'Starting Docker Application Container Engine...' | systemd-cat -t systemd
echo 'Starting containerd container runtime...' | systemd-cat -t systemd
```
3. Check the logs in the first terminal that the events are generated.

To test in a cluster:
1. Copy the file to a GKE node
2. Modify the `lookback` to a large value to cover the start of the journald logs, so that the first few restarts can be detected.
3. Update `/etc/systemd/system/node-problem-detector.service` to include `systemd-monitor.json` as part of `--system-log-monitors`.
4. `sudo systemctl daemon-reload`
5. `sudo systemctl restart node-problem-detector.service`
6. `sudo journalctl -u node-problem-detector` to check the NPD logs that the events are generated
7. `kubectl get events` to check the events show up

### DIFF
--- a/config/systemd-monitor.json
+++ b/config/systemd-monitor.json
@@ -1,0 +1,28 @@
+{
+	"plugin": "journald",
+	"pluginConfig": {
+		"source": "systemd"
+	},
+	"logPath": "/var/log/journal",
+	"lookback": "",
+	"bufferSize": 10,
+	"source": "systemd-monitor",
+	"conditions": [],
+	"rules": [
+		{
+			"type": "temporary",
+			"reason": "KubeletStart",
+			"pattern": "Started Kubernetes kubelet."
+		},
+		{
+			"type": "temporary",
+			"reason": "DockerStart",
+			"pattern": "Starting Docker Application Container Engine..."
+		},
+		{
+			"type": "temporary",
+			"reason": "ContainerdStart",
+			"pattern": "Starting containerd container runtime..."
+		}
+	]
+}
--- a/test/build.sh
+++ b/test/build.sh
@@ -26,15 +26,33 @@ NPD_STAGING_PATH=${NPD_STAGING_PATH:-"gs://node-problem-detector-staging"}
 NPD_STAGING_REGISTRY=${NPD_STAGING_REGISTRY:-"gcr.io/node-problem-detector-staging"}
 PR_ENV_FILENAME=${PR_ENV_FILENAME:-"pr.env"}
 CI_ENV_FILENAME=${CI_ENV_FILENAME:-"ci.env"}
+CI_CUSTOM_FLAGS_ENV_FILENAME=${CI_CUSTOM_FLAGS_ENV_FILENAME:-"ci-custom-flags.env"}
 ROOT_PATH=$(git rev-parse --show-toplevel)
 GCS_URL_PREFIX="https://storage.googleapis.com/"
 
+
 function print-help() {
-  echo "Usage: build.sh [args...]"
-  echo "Available arguments:"
-  echo "  pr [pull_number]: Build node-problem-detector for presubmit jobs and push to staging."
-  echo "  ci:               Build node-problem-detector for CI jobs and push to staging."
-  echo "  get-ci-env:       Download environment variable file from staging for CI job."
+  echo "Usage: build.sh [flags] [command]"
+  echo
+  echo "Available flags:"
+  echo "  -p [PR_NUMBER]      Specify the pull request number when building for a presubmit job."
+  echo "  -f                  Use custom flags."
+  echo
+  echo "Available commands:"
+  echo "  help           Print this help message"
+  echo "  pr             Build node-problem-detector for presubmit jobs and push to staging. Flag -p is required."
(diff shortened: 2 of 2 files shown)


config/systemd-monitor.json
```diff
@@ -0,0 +1,28 @@
...
+	},
+	"logPath": "/var/log/journal",
+	"lookback": "",
+	"bufferSize": 10,
+	"source": "systemd-monitor",
+	"conditions": [],
+	"rules": [
+		{
+			"type": "temporary",
+			"reason": "KubeletRestart",
```
wangzhen127: @Random-Liu, should it be `KubeletStart` instead of `KubeletRestart`? WDYT?

Random-Liu:
@wangzhen127 Don't think I have time to review it today. Will try to get some time after I'm in Taiwan.

wangzhen127:
No problem. Have fun in Taiwan!

config/systemd-monitor.json
andyxning: Does Kubelet restart and start logs different messages? According to the log, it seems not. So,  i think `KubeletStart` should be reasonable. But users should diagnose which `KubeletStart` event represents a normal Kubelet start and which `KubeletStart` event represents a abnormal restart.

config/systemd-monitor.json
```diff
@@ -0,0 +1,28 @@
...
+			"pattern": "Started Kubernetes kubelet."
+		},
+		{
+			"type": "temporary",
+			"reason": "DockerRestart",
+			"pattern": "Starting Docker Application Container Engine..."
```
andyxning: BTW, Seems that this log content is only adapted to GKE, right? If we want to detecte a `DockerRestart` event, we need to change the `pattern`. Since we can not add comment to a json file, Maybe `GKE` prefixed event name is more clear. How about `GKEKubeletStart`/`GKEDockerRestart`.

andyxning: `s/Starting/Started`

config/systemd-monitor.json
```diff
@@ -0,0 +1,28 @@
...
+		},
+		{
+			"type": "temporary",
+			"reason": "ContainerdRestart",
+			"pattern": "Starting containerd container runtime..."
```
andyxning: `s/Starting/Started`

config/systemd-monitor.json
andyxning: In our Debian OS machine. when docker daemon is restarted, the log printed is ` Started Docker Application Container Engine.`. Seems the trailing dot are redundant.

andyxning: > Is it possible the container runtime somehow gets stuck at the 1st log line and never reach the log line with Started? Detecting Starting can count for that.

This is reasonable. 

systemd will print logs with pattern `Started UNIT_FILE_DESCRIPTION`. Will confirm this later.

andyxning: @wangzhen127 According to the source code of systemd https://github.com/systemd/systemd/blob/master/src/core/service.c#L4288-L4301. Systemd will start two lines of log. When a service unit is started `Starting %s ...` and `Started %s .` when the job is done successfully. 

In case you mentioned above, i think `Starting %s ...` should be used. WDYT?

andyxning:
I am not quite sure about the CI part changes. But, it seems ok.

When the `systemd-monitor.json` part is ready, we can merge this PR, imho.

config/systemd-monitor.json
wangzhen127: I agree.

One caveat is that for kubelet, I don't see the `Starting...` log line on GKE VMs. So I used `Started...`. Not sure if it is related to how it is started by systemd (using `systemctl start`).

wangzhen127:
> I am not quite sure about the CI part changes. But, it seems ok.

The CI part is to enable test infra to take customized flags for NPD. So we can have more combinations of test jobs. I have the test infra side PR ready to go after this PR is merged: https://github.com/kubernetes/test-infra/pull/13022

config/systemd-monitor.json
andyxning: > Not sure if it is related to how it is started by systemd (using systemctl start).

This should not be the truth. docker or containerd is also started with `systemctl start`. Can you double check this?

wangzhen127: I have double checked this.

andyxning: Well then, let's use `Started` for kubelet for now as actually this is just a hint for others end users. Users should adapt this pattern for their environment. We are in a heterogeneous world. :)

PR #293, opened 2019-06-14
Do not import plugins unnecessarily.

Random-Liu:
Follow up for https://github.com/kubernetes/node-problem-detector/pull/275.

This PR:
1) Get rid of unnecessary plugin imports in `cmd/options/options.go`;
2) Make NPD panic if deprecated flags are specified, but the corresponding plugin is not registered.

/cc @xueweiz

Signed-off-by: Lantao Liu <lantaol@google.com>


cmd/options/options_test.go
```diff
@@ -319,30 +322,36 @@ func TestSetConfigFromDeprecatedOptionsOrDie(t *testing.T) {
...
-				MonitorConfigPaths: types.ProblemDaemonConfigPathMap{
-					custompluginmonitor.CustomPluginMonitorName: &[]string{"config-b"},
-				},
+		},
+		{
+			name: "using deprecated options when SystemLogMonitor is not registered",
+			orig: NodeProblemDetectorOptions{
+				SystemLogMonitorConfigPaths: []string{"config-a"},
+				MonitorConfigPaths:          types.ProblemDaemonConfigPathMap{},
 			},
```
xueweiz: nit: probably change to:
```
types.ProblemDaemonConfigPathMap{
    customPluginMonitorName: &[]string{},
}
```

And similar for below test case,

xueweiz: Oh just realized that github does not highlight the selected code.
So to clarify:
I think we should change `MonitorConfigPaths:          types.ProblemDaemonConfigPathMap{},` into
```
MonitorConfigPaths: types.ProblemDaemonConfigPathMap{
    customPluginMonitorName: &[]string{},
},
```

Random-Liu: Not needed, because we are not using custom plugin monitor deprecated flag, it is fine to not register that plugin. :) This also helps us proves that.

Random-Liu: How about this?

Random-Liu:
Sent another commit to fix the build. `go build cmd/node_problem_detector.go` won't include `plugins.go` in the output binary.

xueweiz:
Sorry for the test failure. It is caused by my bug in #275 
Seems the [imports](https://github.com/kubernetes/node-problem-detector/blob/975dc718a5ecbda6ebd30faaeecc6472caa64d6a/cmd/plugins.go#L21) in package main are ignored.

cmd/options/options_test.go
xueweiz: Thanks!

PR #300, opened 2019-06-27
Report metrics from system-log-monitor

xueweiz:
Add a new field `metricsReporting` (`true` by default) in the `system-log-monitor` config file.
By setting the `metricsReporting` field to `true`, `system-log-monitor` will report metrics when temporary/permanent problems happen.

The behavior was specified in the NPD metrics design doc [here](https://docs.google.com/document/d/1SeaUz6kBavI283Dq8GBpoEUDrHA2a795xtw0OvjM568/edit#bookmark=id.fgbb40a0mr45).

For example, when I do:
`sudo sh -c "echo 'kernel: INFO: task docker:20744 blocked for more than 120 seconds.' >> /dev/kmsg"`

I will see:
```
$ curl http://localhost:20257/metrics
...
# HELP problem_counter Number of times a specific type of problem have occurred.
# TYPE problem_counter counter
problem_counter{reason="AUFSUmountHung"} 0
problem_counter{reason="DockerHung"} 2
problem_counter{reason="FilesystemIsReadOnly"} 0
problem_counter{reason="KernelOops"} 0
problem_counter{reason="OOMKilling"} 0
problem_counter{reason="TaskHung"} 2
problem_counter{reason="UnregisterNetDevice"} 0
# HELP problem_gauge Whether a specific type of problem is affecting the node or not.
# TYPE problem_gauge gauge
problem_gauge{reason="AUFSUmountHung",type="KernelDeadlock"} 0
problem_gauge{reason="DockerHung",type="KernelDeadlock"} 1
problem_gauge{reason="FilesystemIsReadOnly",type="ReadonlyFilesystem"} 0
```

This PR maintains backward compatibility on existing `system-log-monitor` configs:
```
$ make ./bin/node-problem-detector && ./bin/node-problem-detector --logtostderr --enable-k8s-exporter=false --config.system-log-monitor=./config/abrt-adaptor.json
make: 'bin/node-problem-detector' is up to date.
I0701 17:54:42.052347  140586 log_monitor.go:75] Finish parsing log monitor config file: {WatcherConfig:{Plugin:journald PluginConfig:map[source:abrt-notification] LogPath:/var/log/journal Lookback:5m Delay:} BufferSize:10 Source:abrt-adaptor DefaultConditions:[] Rules:[{Type:temporary Condition: Reason:CCPPCrash Pattern:Process \d+ \(\S+\) crashed in .*} {Type:temporary Condition: Reason:UncaughtException Pattern:Process \d+ \(\S+\) of user \d+ encountered an uncaught \S+ exception} {Type:temporary Condition: Reason:XorgCrash Pattern:Display server \S+ crash in \S+} {Type:temporary Condition: Reason:VMcore Pattern:System encountered a fatal error in \S+} {Type:temporary Condition: Reason:Kerneloops Pattern:System encountered a non-fatal error in \S+}] EnableMetricsReporting:0x1df3f2c}
I0701 17:54:42.052554  140586 log_watchers.go:40] Use log watcher of plugin "journald"
I0701 17:54:42.052638  140586 node_problem_detector.go:64] Prometheus exporter started.
I0701 17:54:42.052650  140586 log_monitor.go:85] Start log monitor
I0701 17:54:42.058273  140586 log_watcher.go:80] Start watching journald
I0701 17:54:42.058315  140586 problem_detector.go:67] Problem detector started
I0701 17:54:42.058349  140586 log_monitor.go:197] Initialize condition generated: []
```

This PR is part of #284  (the System Log Monitor part of below item):

- [ ] Implement feature in System Log Monitor and Custom Plugin Monitor, to allow them to report metrics.


xueweiz:
This is part of #284 :
- [x]  Implement feature in System Log Monitor and Custom Plugin Monitor, to allow them to report metrics.

/assign @wangzhen127

wangzhen127:
A few high level comments:
- For permanent problems, I think it should be gauge instead of counter, right?
- What about custom plugins? Do you plan to do it in the next PR?
- Have you tested this is backward compatible? If so, mention that in the PR description.
- Can you link the issue in the PR description?

xueweiz:
Thanks for the taking a look Zhen!

> For permanent problems, I think it should be gauge instead of counter, right?
True. Reporting permanent problems as a gauge metrics with value only as 0 or 1 is enough, and is more similar to the "node condition" model we uses today.
However, report them as a counter, which goes from 0 to 1 and then to 2...n is better in my opinion. Because:
1. We are not losing any information: we can simply categorize 0 as "node condition: false", and non-0 as "node condition: true".
2. We gain a bit more information about how many times a permanent problem happens. A permanent problem can theoretically happen more than once. Knowing whether they could happen more than once could also be helpful.

> What about custom plugins? Do you plan to do it in the next PR?
Yes. I will submit a follow up PR for custom plugins.

> Have you tested this is backward compatible? If so, mention that in the PR description.
Good point. Yes I have. Just updated the PR description.

> Can you link the issue in the PR description?
Done.

wangzhen127:
> For permanent problems, I think it should be gauge instead of counter, right?

>> True. Reporting permanent problems as a gauge metrics with value only as 0 or 1 is enough, and is more similar to the "node condition" model we uses today.
However, report them as a counter, which goes from 0 to 1 and then to 2...n is better in my opinion. Because:
>> 1. We are not losing any information: we can simply categorize 0 as "node condition: false", and non-0 as "node condition: true".
>> 2. We gain a bit more information about how many times a permanent problem happens. A permanent problem can theoretically happen more than once. Knowing whether they could happen more than once could also be helpful.

For custom plugins, the permanent problems can go away and NPD will reset the condition back to false. Counter is just for monotonically increasing things.

In addition, `permanent` already means the situation is very bad and needs immediate attention. Counting it does not make much sense, I think. If one wants to count how often a problem happens, that should be `temporary` problems.

xueweiz:
> For custom plugins, the permanent problems can go away and NPD will reset the condition back to false.

Oh interesting. I wasn't aware of this. Thanks!
I will change the code so that `permanent` problems are reported as gauge metrics.

xueweiz:
Just changed the code so that `permanent` problems are reported as gauges (0 or 1).
@wangzhen127  could you help take another look? Thanks!

pkg/util/metrics/problem_gauge.go
```diff
@@ -0,0 +1,83 @@
...
+
+	var err error
+	cg.keyReason, err = tag.NewKey("reason")
+	if err != nil {
+		return cg, err
+	}
+	cg.gauge = NewInt64Metric(
+		cg.metricName,
+		description,
+		"",
```
wangzhen127: Ditto: should be "1"?

pkg/systemlogmonitor/log_monitor.go
```diff
@@ -165,7 +200,11 @@ func (l *logMonitor) generateStatus(logs []*logtypes.Log, rule systemlogtypes.Ru
 				break
 			}
 		}
+		if rule.MetricName != "" {
+			l.gauges[rule.MetricName].Set(rule.Reason)
```
wangzhen127: We should also initialize the problem gauges in `initializeStatus()` (line 217).

pkg/util/metrics/problem_gauge.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+func (cg ConditionGauge) Set(reason string) error {
+	if cg.gauge == nil {
+		return fmt.Errorf("condition gauge %q is being set before initialized.", cg.metricName)
+	}
+
+	// If the condition reason has changed, we should clear the previous condition reason.
+	if cg.reason != "" && cg.reason != reason {
+		if err := cg.Clear(); err != nil {
```
wangzhen127: Why do we need to clear the previous condition reason?

pkg/util/metrics/problem_counter.go
```diff
@@ -0,0 +1,58 @@
...
+
+	var err error
+	pc.keyReason, err = tag.NewKey("reason")
+	if err != nil {
+		return pc, err
+	}
+	pc.counter = NewInt64Metric(
+		pc.metricName,
+		description,
+		"",
```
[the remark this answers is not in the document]
xueweiz: Oh you are right, it should be "1".


> How about defining a common type as a const in metrics package and invoke here?

Do you mean something like below?
```
const unitCount = "1"
const unitSecond = "second"
```

I feel it's a little redundant. Today, we have `"1"` and [`"second"`](https://github.com/kubernetes/node-problem-detector/blob/146dfd70b25090203162b1e9eee305e76b688dff/pkg/systemstatsmonitor/disk_collector.go#L52). Soon, we will have "kB", "kB/s" and others.

I slightly prefer to just keep them defined at where the metrics are defined, rather than in a central `metrics` package. I don't have a strong preference though. WDYT?

pkg/util/metrics/problem_gauge.go
```diff
@@ -0,0 +1,83 @@
...
+import (
+	"context"
+	"fmt"
+
+	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/tag"
+)
+
+type ConditionGauge struct {
```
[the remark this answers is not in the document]
xueweiz: SG :) Done.

pkg/util/metrics/problem_gauge.go
xueweiz: Right. Done.

pkg/systemlogmonitor/log_monitor.go
xueweiz: Sounds good. However, I think we can just initialize the metrics at the definition time.
i.e. When caller calls `NewProblemCounter()` or `NewProblemGauge()`, we automatically assign the first measurement for them (i.e. `0`).

That will result in below UX:
```
$ curl http://localhost:20257/metrics
...
# HELP kernel_deadlock Whether below pattern has been observed in log source "kernel-monitor" or not: "task umount\\\\.aufs:\\\\w+ blocked for more than \\\\w+ seconds\\\\."
# TYPE kernel_deadlock gauge
kernel_deadlock{reason=""} 0
# HELP kernel_oom_killing Number of times below pattern is observed in log source "kernel-monitor": "Kill process \\\\d+ (.+) score \\\\d+ or sacrifice child\\\\nKilled process \\\\d+ (.+) total-vm:\\\\d+kB, anon-rss:\\\\d+kB, file-rss:\\\\d+kB.*"
# TYPE kernel_oom_killing counter
kernel_oom_killing{reason=""} 0
# HELP kernel_oops Number of times below pattern is observed in log source "kernel-monitor": "BUG: unable to handle kernel NULL pointer dereference at .*"
# TYPE kernel_oops counter
kernel_oops{reason=""} 0
# HELP kernel_readonly_filesystem Whether below pattern has been observed in log source "kernel-monitor" or not: "Remounting filesystem read-only"
# TYPE kernel_readonly_filesystem gauge
kernel_readonly_filesystem{reason=""} 0
# HELP kernel_task_hung Number of times below pattern is observed in log source "kernel-monitor": "task \\\\S+:\\\\w+ blocked for more than \\\\w+ seconds\\\\."
# TYPE kernel_task_hung counter
kernel_task_hung{reason=""} 0
# HELP kernel_unregister_netdevice Number of times below pattern is observed in log source "kernel-monitor": "unregister_netdevice: waiting for \\\\w+ to become free. Usage count = \\\\d+"
# TYPE kernel_unregister_netdevice counter
kernel_unregister_netdevice{reason=""} 0
```

Does this looks like what you expected?

pkg/util/metrics/problem_gauge.go
xueweiz: Here is the tricky thing. A metrics with different labels are actually considered as two independent metrics.

Let's first see how node condition behaves:

1. At start, it looks like this:
```
{  
   "type":"KernelDeadlock",
   "status":"False",
   "transition":"...",
   "reason":"KernelHasNoDeadlock",
   "message":"kernel has no deadlock"
}
```
2. When a problem happens, it becomes this:
```
{  
   "type":"KernelDeadlock",
   "status":"True",
   "transition":"...",
   "reason":"DockerHung",
   "message":"task dockerd blocked for..."
}
```
3. If a new type of kernel deadlock happens, after condition manager reports to apiserver, it becomes this:
```
{  
   "type":"KernelDeadlock",
   "status":"True",
   "transition":"...",
   "reason":"ContainerdHung",
   "message":"task containerd blocked for..."
}
```
4. If someone then clears the condition, it will become this:
```
{  
   "type":"KernelDeadlock",
   "status":"False",
   "transition":"...",
   "reason":"KernelHasNoDeadlock",
   "message":"kernel has no deadlock"
}
```

However, for metrics, given the fact that below three metrics are independent, it becomes tricky:

- kernel_task_hung{reason=""} 0
- kernel_task_hung{reason="DockerHung"} 1
- kernel_task_hung{reason="ContainerdHung"} 1

Now, let's do the same workflow again, assuming that we do not clear the previous `reason`:
1. At start, it looks like this:
```
- kernel_task_hung{reason=""} 0
```
2. When a problem happens, it becomes this (two metrics):
```
- kernel_task_hung{reason=""} 0
- kernel_task_hung{reason="DockerHung"} 1
```

3. If a new type of kernel deadlock happens, it becomes this (three metrics):
```
- kernel_task_hung{reason=""} 0
- kernel_task_hung{reason="DockerHung"} 1
- kernel_task_hung{reason="ContainerdHung"} 1
```
4. If someone then clears the condition, it will become this:
```
- kernel_task_hung{reason=""} 0
- kernel_task_hung{reason="DockerHung"} 0
- kernel_task_hung{reason="ContainerdHung"} 0
```

Now let's do the same workflow again, assuming that we clear the previous `reason` (current PR implementation):
1. At start, it looks like this:
```
- kernel_task_hung{reason=""} 0
```
2. When a problem happens, it becomes this (two metrics):
```
- kernel_task_hung{reason=""} 0
- kernel_task_hung{reason="DockerHung"} 1
```

3. If a new type of kernel deadlock happens, it becomes this (three metrics):
```
- kernel_task_hung{reason=""} 0
- kernel_task_hung{reason="DockerHung"} 0
- kernel_task_hung{reason="ContainerdHung"} 1
```
4. If someone then clears the condition, it will become this:
```
- kernel_task_hung{reason=""} 0
- kernel_task_hung{reason="DockerHung"} 0
- kernel_task_hung{reason="ContainerdHung"} 0
```

Note how step 3 in above two cases are different, depending on whether we clear the previous reason or not:

- By clearing the previous `reason`, it is more similar to what we have with node condition: at any given point of time, only the most recent `reason` can be true (non-zero);
- By not clearing the previous `reason`, we obtain information about what are the `reason` that has happened over the history.

I chose to clear the `reason` simply to stay closer with node condition's implementation. I'm happy if we choose to not to clear the `reason` as well.

Do you have any preference on which one to choose from the above two options?

xueweiz:
I'm working on some unit tests now. Zhen, could you help review the rest of the changes again? Thanks!

pkg/util/metrics/problem_counter.go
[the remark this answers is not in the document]
wangzhen127: Sounds good.

xueweiz:
Hi Zhen, I just added some unit tests. Could you help take another look? Thanks!

config/kernel-monitor.json
```diff
@@ -4,6 +4,7 @@
 	"lookback": "5m",
 	"bufferSize": 10,
 	"source": "kernel-monitor",
+	"metricsReporting": true,
```
wangzhen127: I remembered that you used a different approach previously, i.e., allow setting the metric name. Why do we change it to a global bool?

pkg/systemlogmonitor/README.md
```diff
@@ -86,3 +86,26 @@ field in the configuration file is the log path. You can always configure
...
+By setting the boolean `metricsReporting` at top level, you can choose to enable or disable
+metrics reporting of System Log Monitor. If you omit the field, it will be set to `true` by
+default.
+
+Temporary problems will be reported as counter metrics, such as below example:
+
+```
+# HELP problem_counter Number of times a specific type of problem have occurred.
+# TYPE problem_counter counter
+problem_counter{reason="TaskHung"} 2
```
wangzhen127: All temporary metrics are now called `problem_counter` instead of their own real metric name? That seems not straightforward.

If metric name is provided, we can use that name.
If global bool is used, we can translate reason to metric name.

config/kernel-monitor.json
wangzhen127: Setting the metric name seems more flexible to me.
- It allows using arbitrary metric name
- It enables reporting only some metrics from the problem, but not all of them (e.g., only report metrics when metric name is provided).

xueweiz: Sorry my previous implementation was a mistake, and deviated from my original [design doc](https://docs.google.com/document/d/1SeaUz6kBavI283Dq8GBpoEUDrHA2a795xtw0OvjM568/edit#bookmark=id.fgbb40a0mr45).

So basically we can choose how to represent problems as metrics, in one of the below two forms:

- problem_counter{"name": "KernelDeadlock"}  0
- problem_kernel_deadlock  0 

Basically, the choice is whether to put the problem type in the metric name, or metric label. Both designs offer equal level of "query-bility". i.e. Any metrics query you can do with one design, you can always translate it easily to a query on the other design.

I think we should put them in metric label by default, see the reasoning [here](https://docs.google.com/document/d/1rcPBMJChlFXbrPpXHlE-gk6DtUQh6p85zqu5Aq3ySuc/edit#bookmark=id.w57y0telmrx).
tl;dr: Mainly because we want to be able to add new problem types easily. Adding a metrics endpoint typically requires some monitoring backend side review, and can be slow. Adding a metrics label value is very lightweight.


In the [design doc](https://docs.google.com/document/d/1SeaUz6kBavI283Dq8GBpoEUDrHA2a795xtw0OvjM568/edit#bookmark=id.fgbb40a0mr45), I also mentioned this as extension:

> And we plan to support some more customized behaviors using the config file. For example:
>  - Report the metric as gauge, rather than counter.
>  - Allow user to change the metric name.
>  - ...

This PR is implementing 1).
I don't think 2) is needed right now. When we might need it, I'm happy to submit a PR for it.

wangzhen127: Discussed offline. Will go with problem_counter approach.

pkg/util/metrics/problem_gauge.go
wangzhen127: I see your points now. Clearing the previous reason is fine. But when the condition is false, I think the reason should not be "", but should be something like "KernelHasNoDeadlock".

xueweiz: >  But when the condition is false, I think the reason should not be "", but should be something like "KernelHasNoDeadlock".

I agree. Sorry this is a bug in my first implementation.

This bug happened, because NPD set condition "KernelDeadlock" to "Status: false, Reason:KernelHasNoDeadlock" upon NPD start. And in my initial implementation, I'm setting the metrics based on it.

In the current implementation, at file `pkg/systemlogmonitor/log_monitor.go` function `initialProblemGauges`, there is this piece of code:
```
for _, condition := range initialConditions {
if condition.Status != types.True {
	continue
}
```

That way I can avoid setting a `problem_gauge` metrics with these nonsense reasons (e.g. "KernelHasNoDeadlock").

I also updated the README and PR description to reflect this change.

xueweiz:
I just rebased this PR, now that #303 is merged.

pkg/problemmetrics/problem_metrics.go
```diff
@@ -0,0 +1,110 @@
...
+
+func NewProblemMetricsManagerOrDie() *ProblemMetricsManager {
+	pmm := ProblemMetricsManager{}
+
+	var err error
+	pmm.problemCounter, err = metrics.NewInt64Metric(
+		"problem_counter",
+		"Number of times a specific type of problem have occurred.",
+		"1",
+		metrics.Sum,
```
wangzhen127: Should this be sum or count?

pkg/problemmetrics/problem_metrics.go
```diff
@@ -0,0 +1,110 @@
...
+		glog.Fatalf("Failed to create problem_gauge metric: %v", err)
+	}
+
+	pmm.conditionToReason = make(map[string]string)
+
+	return &pmm
+}
+
+// NewProblem reports a temporary problem that results in problem counter increment.
+func (pmm *ProblemMetricsManager) NewProblem(reason string) error {
```
wangzhen127: s/NewProblem/NewTemporaryProblem

pkg/problemmetrics/problem_metrics.go
```diff
@@ -0,0 +1,110 @@
...
+	if pmm.problemCounter == nil {
+		return errors.New("problem counter is being incremented before initialized.")
+	}
+
+	return pmm.problemCounter.Record(map[string]string{"reason": reason}, 1)
+}
+
+// SetCondition reports a permanent problem that result in a problem gauge value set.
+func (pmm *ProblemMetricsManager) SetCondition(condition string, reason string, value bool) error {
```
wangzhen127: Should this function be:
```
NewPermanentProblem(type string, reason string, value bool) error {
```

pkg/problemmetrics/problem_metrics.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		glog.Fatalf("Failed to create problem_counter metric: %v", err)
+	}
+
+	pmm.problemGauge, err = metrics.NewInt64Metric(
+		"problem_gauge",
+		"Whether a specific type of problem is affecting the node or not.",
+		"1",
+		metrics.LastValue,
+		[]string{"condition", "reason"})
```
wangzhen127: The `condition` is a k8s specific concept. I think for the problem metrics manager, it is better not to use this term. Call it `type` or something similar? (`type` is consistent with the config)

pkg/problemmetrics/problem_metrics.go
```diff
@@ -0,0 +1,110 @@
...
+func init() {
+	GlobalProblemMetricsManager = NewProblemMetricsManagerOrDie()
+}
+
+// ProblemMetricsManager manages problem-converted metrics.
+// ProblemMetricsManager is thread-safe.
+type ProblemMetricsManager struct {
+	problemCounter         metrics.Int64MetricInterface
+	problemGauge           metrics.Int64MetricInterface
+	conditionToReason      map[string]string
```
wangzhen127: Call this `PermanentProblemTypeToReason` or something similar? Also see other comments below in this file.

pkg/problemmetrics/problem_metrics.go
```diff
@@ -0,0 +1,110 @@
...
+func (pmm *ProblemMetricsManager) SetCondition(condition string, reason string, value bool) error {
+	if pmm.problemGauge == nil {
+		return errors.New("problem gauge is being set before initialized.")
+	}
+
+	pmm.conditionToReasonMutex.Lock()
+	defer pmm.conditionToReasonMutex.Unlock()
+
+	lastReason := pmm.conditionToReason[condition]
+	if lastReason != "" {
```
wangzhen127: If you are checking if `condition` is in the map, it is better to use `ok`:
```
if lastReason, ok := pmm.conditionToReason[condition], ok {
    ...
}
```
instead of assuming `lastReason` is not empty string.

pkg/systemlogmonitor/log_monitor.go
```diff
@@ -196,6 +212,19 @@ func initialConditions(defaults []types.Condition) []types.Condition {
 	return conditions
 }
 
+func initialProblemGauges(initialConditions []types.Condition) {
+	for _, condition := range initialConditions {
+		if condition.Status != types.True {
```
wangzhen127: In `initialConditions`, the initial ones are all set to false, so this function becomes noops, right?

I think the purpose of this function is to create problem_gauges that have conditions being false, right (the so called default conditions)?

pkg/problemmetrics/problem_metrics.go
xueweiz: I slightly prefer Sum for following reasons:
1. Sum's operations is a super set of Count's operations.
2. They will both show up as `# TYPE counter` in Prometheus metrics, so they won't confuse the user.
3. If we use Sum, we retain the flexibility to implement below behaviors (I'm not saying we want to implement these behaviors. I just think that **if** we want them for some reason, we should be able to do them):

    - Preset a data point for the metrics during initialization, with value as 0.
All metrics are created empty, with no data point. `Sum(0)` will introduce a 0-valued data point, while `Count(0)` will introduce a 1-valued data point. There is no way to introduce a 0-valued Sum data point.
    - Make the data point stateful across NPD restarts/crashes.
Today, when NPD crashes, all metrics are lost, and will start from 0 after NPD starts up.
Say **if** we want to allow NPD to store the counters at disk, and restore them after crash. And if the counter has a value of 1000, it would require calling `Count(x)` 1000 times, while require calling `Sum(1000)` one time.
Using Count make implementing this much harder.

pkg/problemmetrics/problem_metrics.go
xueweiz: Sounds good. Thanks!

pkg/problemmetrics/problem_metrics.go
xueweiz: Thanks! Done.

pkg/problemmetrics/problem_metrics.go
xueweiz: Sounds good. Thanks!

pkg/problemmetrics/problem_metrics.go
xueweiz: Sounds good.

pkg/problemmetrics/problem_metrics.go
xueweiz: Ah sorry. Don't know what was I thinking.

pkg/systemlogmonitor/log_monitor.go
xueweiz: > In initialConditions, the initial ones are all set to false, so this function becomes noops, right?

Correct. However, this function is still useful **if** some condition has initial value set to True.

> I think the purpose of this function is to create problem_gauges that have conditions being false

That was the case for k8s conditions. However, I don't think the same UI looks good with Prometheus. Mostly because a metric with different labels are consider as multiple metrics in Prometheus, and the UI looks ugly.

For example, if I keep the code, and skip when the `condition.Status != True`, then this is what I get (which look pretty OK):
```
# HELP problem_gauge Whether a specific type of problem is affecting the node or not.
# TYPE problem_gauge gauge
problem_gauge{reason="DockerHung",type="KernelDeadlock"} 1
```

And if I remove the skip logic, then the output looks like this:
```
# HELP problem_gauge Whether a specific type of problem is affecting the node or not.
# TYPE problem_gauge gauge
problem_gauge{reason="DockerHung",type="KernelDeadlock"} 1
problem_gauge{reason="FilesystemIsNotReadOnly",type="ReadonlyFilesystem"} 0
problem_gauge{reason="KernelHasNoDeadlock",type="KernelDeadlock"} 0
```

Basically, the initial states `problem_gauge{reason="KernelHasNoDeadlock",type="KernelDeadlock"} 0`  and `problem_gauge{reason="FilesystemIsNotReadOnly",type="ReadonlyFilesystem"} 0` will show up there all the time. That seems a bit ugly and misleading.

xueweiz:
Hi @wangzhen127 , thanks for the review! I just fixed the problems. Could you help take another look? Thanks!

pkg/problemmetrics/problem_metrics.go
wangzhen127: Thanks for the explanation!

pkg/systemlogmonitor/log_monitor.go
wangzhen127: Hmm, I think we need to clarify what does `0` and `1` mean for the problem gauge.

With your current implementation:
- For `DockerHung`, `0` means no such condition, `1` means it has such condition.
- However, for `KernelHasNoDeadlock`, `0` now means it has such condition (the condition is that "there is no deadlock". And the value will never be `1`.

The above looks inconsistent. IMO, the value `0` and `1` should just mean `false` and `true`. So the initial state for `KernelHasNoDeadlock` should be:
```
problem_gauge{reason="KernelHasNoDeadlock",type="KernelDeadlock"} 1
```

Later when we have `DockerHung`, it should become:
```
problem_gauge{reason="DockerHung",type="KernelDeadlock"} 1
problem_gauge{reason="KernelHasNoDeadlock",type="KernelDeadlock"} 0
```

xueweiz: 1. 
> However, for KernelHasNoDeadlock, 0 now means it has such condition (the condition is that "there is no deadlock". And the value will never be 1.

> The above looks inconsistent. IMO, the value 0 and 1 should just mean false and true. So the initial state for KernelHasNoDeadlock should be:

I guess there is a misunderstanding. `KernelHasNoDeadlock` is actually not the permanent problem (i.e. not the `condition`), it is the `reason` field of the `KernelDeadlock` condition.

So `0` or `1` does not indicate whether `KernelHasNoDeadlock` is happening or not; it indicates whether `KernelDeadlock` is happening or not.
`reason=xxx` field explain why is `KernelDeadlock` happening (e.g. `DockerHung`), or why is it not happening (e.g. `KernelHasNoDeadlock`, in which case I think is very confusing).

2. 

First I want to say that both `problem_gauge{reason="KernelHasNoDeadlock",type="KernelDeadlock"} 0` and `problem_gauge{reason="KernelHasNoDeadlock",type="KernelDeadlock"} 1` is very confusing and misleading.

What we are trying to say is, "the node is not suffering from KernelDeadlock problems", and the reason of this is "the kernel has no deadlock".

I think the best UI is to not show the "KernelHasNoDeadlock" field at all.

3.

> I think we need to clarify what does 0 and 1 mean for the problem gauge.

In my opinion, for problem gauge, `0` and `1` should indicate "whether a specific type of problem is affecting the node or not" (the quote is the description of the `problem_gauge` Prometheus metric).

So when the value is `0`, it should mean "this type of problem is NOT affecting the node"; when the value is `1`, it should mean "this type of problem is affecting the node".

4.

I think showing `problem_gauge{reason="KernelHasNoDeadlock",type="KernelDeadlock"}` only confuses the user, and has no real benefit. i.e. If a permanent problem is not happening, the user don't care why it's not happening. The `reason` is always gonna be a very dumb thing to say, like "Kernel is not deadlocking", "OOM is not happening". I think it only confuses the user.

The right user experience should be, they simply build a Prometheus query, and see whether "problem_gauge{type="KernelDeadlock"}" is 0 (using `addition` to aggregate all sub-metrics like `problem_gauge{type="KernelDeadlock", reason="DockerHung"}` or `problem_gauge{type="KernelDeadlock", reason="ContainerdHung"}` ).

If it's 0, then the problem is not affecting the node; if it's 1, then the problem is affecting the node.

And if it's 1, the 1-valued sub-metric should have a `reason` field describing what caused the problem.

WDYT? Is this UX reasonable?

wangzhen127: Your points make sense. We do not need `KernelHasNoDeadlock`.

I think what I tried to get is to have some initial value for the problems (can be 0). The metrics that I have seen so far have values starting from the beginning, like disk io, cpu, memory, etc. So you can see a line starting from the beginning of the life of the node and lasting till the end of the life of the node.

For problem gauges, it seems to me that metrics start to have values only when some problems occur. I am not sure if that's going to be an issue or not. For example, for most of the time, there is no metric for problem_gauge and problem_counter when no issue occurs. This makes me feel inconsistent with other metrics, where we always have some values when a node is alive.

What in my mind is more like:

When node is booted, we have
```
problem_gauge{reason="DockerHung",type="KernelDeadlock"} 0
problem_gauge{reason="AUFSUmountHung",type="KernelDeadlock"} 0
```

Then when we have AUFSUmountHung, we get
```
problem_gauge{reason="DockerHung",type="KernelDeadlock"} 0
problem_gauge{reason="AUFSUmountHung",type="KernelDeadlock"} 1
```

Then we have DockerHung, we get
```
problem_gauge{reason="DockerHung",type="KernelDeadlock"} 1
problem_gauge{reason="AUFSUmountHung",type="KernelDeadlock"} 0
```

Similarly for temporary problems, the initial values are like
```
problem_counter{reason="TaskHung"} 0
```

xueweiz: Ah I see. I like that. I will implement this behavior today. Thanks!

xueweiz: Done.

xueweiz:
Hi @wangzhen127 , I just changed the code to match the [expected behavior](https://github.com/kubernetes/node-problem-detector/pull/300#discussion_r301894362), and added test for the new `initializeProblemMetricsOrDie()` function.
Could you help take another look? Thanks!

pkg/systemlogmonitor/log_monitor.go
```diff
@@ -73,13 +73,35 @@ func NewLogMonitorOrDie(configPath string) types.Monitor {
 		glog.Fatalf("Failed to validate matching rules %+v: %v", l.config.Rules, err)
 	}
 	glog.Infof("Finish parsing log monitor config file: %+v", l.config)
+
 	l.watcher = logwatchers.GetLogWatcherOrDie(l.config.WatcherConfig)
 	l.buffer = NewLogBuffer(l.config.BufferSize)
 	// A 1000 size channel should be big enough.
 	l.output = make(chan *types.Status, 1000)
+
+	initializeProblemMetricsOrDie(l.config.Rules)
```
wangzhen127: Should we call this in the monitor loop, alongside with `l.initializeStatus()`?

pkg/systemlogmonitor/log_monitor.go
```diff
@@ -73,13 +73,35 @@ func NewLogMonitorOrDie(configPath string) types.Monitor {
...
 	return l
 }
 
+// initializeProblemMetricsOrDie creates problem metrics for all problems and set the value to 0,
+// panic if error occurs.
+func initializeProblemMetricsOrDie(rules []systemlogtypes.Rule) {
+	for _, rule := range rules {
```
wangzhen127: We also need to check if it enables metrics reporting first, right?

pkg/systemlogmonitor/log_monitor.go
```diff
@@ -196,6 +233,19 @@ func initialConditions(defaults []types.Condition) []types.Condition {
 	return conditions
 }
 
+func initialProblemGauges(initialConditions []types.Condition) {
```
wangzhen127: Shall this be removed as we have `initializeProblemMetricsOrDie` now?

wangzhen127:
And as part of the testing, can you update the build script to use new flags?
https://github.com/kubernetes/node-problem-detector/blob/master/test/build.sh#L109

config/kernel-monitor.json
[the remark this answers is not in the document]
xueweiz: SG. Done.

pkg/problemmetrics/problem_metrics.go
```diff
@@ -0,0 +1,109 @@
...
+		"1",
+		metrics.Sum,
+		[]string{"reason"})
+	if err != nil {
+		glog.Fatalf("Failed to create problem_counter metric: %v", err)
+	}
+
+	pmm.problemGauge, err = metrics.NewInt64Metric(
+		"problem_gauge",
+		"Whether a specific type of problem is affecting the node or not.",
```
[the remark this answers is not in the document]
xueweiz: I think we added `problem_gauge` simply for one reason:
**When we know a (permanent) problem is no longer affecting the node, we will clear it.**

Other than this, I don't see any use of `problem_gauge`.

So I think the "**or not**" part is the key in the current description:
> Whether a specific type of problem is affecting the node or not

I actually think the description is OK. Basically:
1. when we see a `problem_gauge` is `1`, we know it's affecting the node; 
1. when we see a `problem_gauge` is `0`, we know it's not affecting the node;
1. for any problem that does not show up on `problem_gauge`, we do not know whether it's affecting the node or node.

So in the case of temporary problems, we only know whether they happened or not. We do not know whether they are affecting the node or not. So I think it's consistent: they do not show up in `problem_gauge`, and we don't know whether they are affecting the node or not.

WDYT?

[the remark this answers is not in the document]
wangzhen127: Good points. SGTM. Thanks for the explanation!

pkg/problemmetrics/problem_metrics.go
```diff
@@ -0,0 +1,109 @@
...
+// NewPermanentProblem reports a permanent problem that result in a problem gauge value set.
+func (pmm *ProblemMetricsManager) NewPermanentProblem(problemType string, reason string, value bool) error {
+	if pmm.problemGauge == nil {
+		return errors.New("problem gauge is being set before initialized.")
+	}
+
+	pmm.permanentProblemTypeToReasonMutex.Lock()
+	defer pmm.permanentProblemTypeToReasonMutex.Unlock()
+
+	if lastReason, ok := pmm.permanentProblemTypeToReason[problemType]; ok {
```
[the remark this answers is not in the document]
xueweiz: SG. Done.

pkg/systemlogmonitor/log_monitor.go
```diff
@@ -73,13 +73,35 @@ func NewLogMonitorOrDie(configPath string) types.Monitor {
...
+// panic if error occurs.
+func initializeProblemMetricsOrDie(rules []systemlogtypes.Rule) {
+	for _, rule := range rules {
+		if rule.Type == types.Temp {
+			err := problemmetrics.GlobalProblemMetricsManager.NewTemporaryProblems(rule.Reason, 0)
+			if err != nil {
+				glog.Fatalf("Failed to initialize temporary problem metrics for %q: %v", rule.Reason, err)
+			}
+		} else {
+			err := problemmetrics.GlobalProblemMetricsManager.NewPermanentProblem(rule.Condition, rule.Reason, false)
```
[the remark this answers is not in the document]
xueweiz: SG. I like it. Done.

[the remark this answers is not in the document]
xueweiz: In that case, I will rename `NewTemporaryProblems()` and `NewPermanentProblem()` to `IncrementProblemCounter()` and `SetProblemGauge()`.
Because new we are now reporting to problem counter for permanent problems as well.

pkg/systemlogmonitor/log_monitor.go
```diff
@@ -165,7 +193,15 @@ func (l *logMonitor) generateStatus(logs []*logtypes.Log, rule systemlogtypes.Ru
 				break
 			}
 		}
+		if *l.config.EnableMetricsReporting {
+			err := problemmetrics.GlobalProblemMetricsManager.NewPermanentProblem(rule.Condition, rule.Reason, true)
```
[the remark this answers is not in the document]
xueweiz: SG. Done.

In that case, I will rename `NewTemporaryProblems()` and `NewPermanentProblem()` to `IncrementProblemCounter()` and `SetProblemGauge()`.
Because new we are now reporting to problem counter for permanent problems as well.

pkg/systemlogmonitor/log_monitor.go
xueweiz: I prefer not. Mainly because I prefer any function that ends in `OrDie` (i.e. may panic when en-counting error) happens before `Start()`.

And if declaring/initializing a metric failed, I actually prefer NPD to just panic. Because that's most likely a configuration problem. And I prefer the failure to be very loud.

pkg/systemlogmonitor/log_monitor.go
xueweiz: Good catch. Done.

pkg/systemlogmonitor/log_monitor.go
xueweiz: Sounds fair. Let's remove it.

xueweiz: This function is only useful for [this](https://github.com/kubernetes/node-problem-detector/issues/304#issuecomment-508870052).

When I work on #304 , I will take these into consideration.

xueweiz:
> And as part of the testing, can you update the build script to use new flags?
The build script is already using `kernel-monitor.json`, `docker-monitor.json` and `systemd-monitor.json`. And these three config files all already have the new flag enabled (i.e. `"metricsReporting": true`).

I think that's good enough, right?

Prometheus exporter is already enabled by default, at port 20257.
So in the testing, the metrics is already being exported locally.

xueweiz:
Hi @wangzhen127 , I just address the comments. Could you help take a look again? Huge thanks :)

pkg/problemmetrics/problem_metrics.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		glog.Fatalf("Failed to create problem_gauge metric: %v", err)
+	}
+
+	pmm.permanentProblemTypeToReason = make(map[string]string)
+
+	return &pmm
+}
+
+// NewTemporaryProblem increments the value of a problem counter.
```
wangzhen127: s/NewTemporaryProblem/IncrementProblemCounter

pkg/problemmetrics/problem_metrics.go
```diff
@@ -0,0 +1,113 @@
...
+func init() {
+	GlobalProblemMetricsManager = NewProblemMetricsManagerOrDie()
+}
+
+// ProblemMetricsManager manages problem-converted metrics.
+// ProblemMetricsManager is thread-safe.
+type ProblemMetricsManager struct {
+	problemCounter                    metrics.Int64MetricInterface
+	problemGauge                      metrics.Int64MetricInterface
+	permanentProblemTypeToReason      map[string]string
```
wangzhen127: Since we don't call temporary or permanent problems, maybe call this `problemTypeToReason`.

pkg/problemmetrics/problem_metrics.go
```diff
@@ -0,0 +1,113 @@
...
+func (pmm *ProblemMetricsManager) SetProblemGauge(problemType string, reason string, value bool) error {
+	if pmm.problemGauge == nil {
+		return errors.New("problem gauge is being set before initialized.")
+	}
+
+	pmm.permanentProblemTypeToReasonMutex.Lock()
+	defer pmm.permanentProblemTypeToReasonMutex.Unlock()
+
+	// We clear the last reason, because the expected behavior is that at any point of time,
+	// for each type of permanent problem, there should be at most one reason got set to 1.
```
wangzhen127: Explain that this is to be consistent with k8s condition.

pkg/systemlogmonitor/log_monitor.go
```diff
@@ -165,7 +188,21 @@ func (l *logMonitor) generateStatus(logs []*logtypes.Log, rule systemlogtypes.Ru
 				break
 			}
 		}
+		if *l.config.EnableMetricsReporting {
```
wangzhen127: Put this into the if clause on line 176.

pkg/systemlogmonitor/log_monitor.go
```diff
@@ -165,7 +188,21 @@ func (l *logMonitor) generateStatus(logs []*logtypes.Log, rule systemlogtypes.Ru
...
+			err := problemmetrics.GlobalProblemMetricsManager.SetProblemGauge(rule.Condition, rule.Reason, true)
+			if err != nil {
+				glog.Errorf("Failed to update problem gauge metrics for problem %q, reason %q: %v",
+					rule.Condition, rule.Reason, err)
+			}
+		}
+	}
+	if *l.config.EnableMetricsReporting {
+		err := problemmetrics.GlobalProblemMetricsManager.IncrementProblemCounter(rule.Reason, 1)
```
wangzhen127: This needs to put back into line 167 and line 184. For condition, we only generate conditions when the if check on line 176 is true.

pkg/systemlogmonitor/log_monitor.go
wangzhen127: ACK

pkg/problemmetrics/problem_metrics.go
xueweiz: Sorry.

pkg/problemmetrics/problem_metrics.go
xueweiz: Good catch. Done.

pkg/problemmetrics/problem_metrics.go
xueweiz: Done.

pkg/systemlogmonitor/log_monitor.go
xueweiz: Done.

pkg/systemlogmonitor/log_monitor.go
xueweiz: Done.

xueweiz:
Huge thanks Zhen, for all the reviews on this PR!

PR #315, opened 2019-07-22
Report metrics from custom-plugin-monitor

xueweiz:
Add a new field `metricsReporting` (`true` by default) in the `custom-plugin-monitor` config file.
By setting the `metricsReporting` field to `true`, `custom-plugin-monitor` will report metrics when temporary/permanent problems happen.

The behavior was specified in the NPD metrics design doc [here](https://docs.google.com/document/d/1SeaUz6kBavI283Dq8GBpoEUDrHA2a795xtw0OvjM568/edit#bookmark=id.fgbb40a0mr45).

For example, when I make the custom plugin script to report problem (by manually editing the script to force it return error), I will see:
```
$ curl http://localhost:20257/metrics
# HELP problem_counter Number of times a specific type of problem have occurred.
# TYPE problem_counter counter
problem_counter{reason="NTPIsDown"} 5
# HELP problem_gauge Whether a specific type of problem is affecting the node or not.
# TYPE problem_gauge gauge
problem_gauge{reason="NTPIsDown",type="NTPProblem"} 1
```

This PR maintains backward compatibility on existing `custom-plugin-monitor` configs.

This PR is part of #284:

- [ x ] Implement feature in System Log Monitor and Custom Plugin Monitor, to allow them to report metrics.

Most of the heavy lifting has been done in #300 .

I tried to update the README for custom-plugin-monitor, but seems that's not necessary:
1. The README for custom-plugin-monitor simply points to its design doc.
1. In the design doc, all common configuration flags (e.g. `rules`, `conditions`) were omitted.

Since the `metricsReporting` flag is now a common config flag used by both `system-log-monitor` and `custom-plugin-monitor`, I don't think it's necessary to add it to the custom-plugin-monitor design doc.

### DIFF
--- a/pkg/custompluginmonitor/custom_plugin_monitor.go
+++ b/pkg/custompluginmonitor/custom_plugin_monitor.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/node-problem-detector/pkg/custompluginmonitor/plugin"
 	cpmtypes "k8s.io/node-problem-detector/pkg/custompluginmonitor/types"
 	"k8s.io/node-problem-detector/pkg/problemdaemon"
+	"k8s.io/node-problem-detector/pkg/problemmetrics"
 	"k8s.io/node-problem-detector/pkg/types"
 	"k8s.io/node-problem-detector/pkg/util"
 	"k8s.io/node-problem-detector/pkg/util/tomb"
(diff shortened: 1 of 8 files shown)


xueweiz:
/cc @andyxning 
Hi Ning, could you also help take a look of this PR? Thanks! Since you are the original author of this plugin, I want to hear your opinion.

xueweiz:
My bad. I'll fix the test.
Also, since Lantao is back now:
/assign @Random-Liu

xueweiz:
@Random-Liu Hi Lantao, sorry for the delay. I just fixed the tests. Could you help take a look? Thanks!

pkg/custompluginmonitor/custom_plugin_monitor.go
```diff
@@ -202,6 +252,19 @@ func (c *customPluginMonitor) generateStatus(result cpmtypes.Result) *types.Stat
...
+						if err != nil {
+							glog.Errorf("Failed to update problem counter metrics for %q: %v", result.Rule.Reason, err)
+						}
+						err = problemmetrics.GlobalProblemMetricsManager.SetProblemGauge(result.Rule.Condition, result.Rule.Reason, true)
+						if err != nil {
+							glog.Errorf("Failed to update problem gauge metrics for problem %q, reason %q: %v",
+								result.Rule.Condition, result.Rule.Reason, err)
+						}
+					}
 				}
```
Random-Liu: Generate counter and gauge at the end of the function based on the final events and condition.

The logic is too complex, the code makes it even harder to maintain.

Random-Liu: And let's do the same for system log monitor.

xueweiz: Ah this is a very good idea! Thanks for noticing that1
Done.

pkg/custompluginmonitor/custom_plugin_monitor.go
```diff
@@ -196,22 +225,50 @@ func (c *customPluginMonitor) generateStatus(result cpmtypes.Result) *types.Stat
...
+		// Increment problem counter only for active problems which just got detected.
+		for _, event := range activeProblemEvents {
+			err := problemmetrics.GlobalProblemMetricsManager.IncrementProblemCounter(
+				event.Reason, 1)
+			if err != nil {
+				glog.Errorf("Failed to update problem counter metrics for %q: %v",
+					event.Reason, err)
+			}
+		}
+		for _, condition := range metricChangingConditions {
```
Random-Liu: Can't we just go through condition list, and set problem gauge for all conditions correspondingly?

xueweiz: Oh just realized that in custom-plugin-monitor, the generateStatus() is only called periodically. Then I think it's fine. Changing the code.

(in system-log-monitor, the generateStatus() is called for each matched Rule, which is much more frequent. So I'd prefer to keep the `changedConditions` logic in there.

xueweiz: Done. Thanks!

Random-Liu: SG. The logic in system-log-monitor is simple, so I'm also fine with adding smarter logic there. :)

PR #320, opened 2019-07-30
Don't update condition if status stays False/Unknown for custom plugin

### LINKED ISSUE #202: Reason change once and it stay in wrong state with custom plugins

### LINKED ISSUE #306: LastHeartbeatTime is not update when plugin does not have a condition
We have a plugin that runs every 5 minutes. Previously it had a `conditions`

```json
  "conditions": [
    {
      "type": "NetworkConnectivityProblem",
      "reason": "NetworkConnectivityNormal",
      "message": "Network connectivity is normal"
    }
  ],
```

which we removed because it causes:
 -  `LastTransitionTime` to be updated on each deploy https://github.com/kubernetes/node-problem-detector/issues/304
 - `LastHeartbeatTime` to be updated every minute even if it did not run

... but now it does not update `LastHeartbeatTime` at all, even though we see in the logs that it runs (succeeds) every 5 minutes. So we have to chose between 2 broken states ... please add a fix or allow some config like `heartBeatMakesSense: true` that updates the heartbeat only when the plugin runs 🤷‍♂ 

FYI rest of the plugin config
```json
{
  "plugin": "custom",
  "pluginConfig": {
    "invoke_interval": "5m",
    "timeout": "60s",
    "max_output_length": 80,
    "concurrency": 1
  },
  "source": "node-network-connectivity",
  "rules": [
    {
      "type": "permanent",
      "condition": "NetworkConnectivityProblem",
      "reason": "NetworkConnectivityFail",
      "path": "./config/plugin/network_connectivity.rb",
      "timeout": "60s"
    }
  ]
}
```
/cc @eatwithforks @xueweiz @wangzhen127

wangzhen127:
Fixes #202

When condition status stays True, we can update condition if reason or message changes. But when condition status stays False/Unknown, we should just ignore and do not update condition.

Here is the example:
1. NPD starts up, custom condition is set to False and reason is set to default reason.
2. When custom plugin runs, status stays False, but `result.Rule.Reason` is always different from the existing reason, which is the default reason set in step 1. In this case, we should just ignore and do not update condition.

Also added some validation to check custom plugin config.
Fixes #306

### DIFF
--- a/pkg/custompluginmonitor/custom_plugin_monitor.go
+++ b/pkg/custompluginmonitor/custom_plugin_monitor.go
@@ -156,77 +156,78 @@ func (c *customPluginMonitor) generateStatus(result cpmtypes.Result) *types.Stat
 			})
 		}
 	} else {
-		// For permanent error changes the condition
+		// For permanent error that changes the condition
 		for i := range c.conditions {
 			condition := &c.conditions[i]
 			if condition.Type == result.Rule.Condition {
-				status := toConditionStatus(result.ExitStatus)
-				// change 1: Condition status change from True to False/Unknown
-				if condition.Status == types.True && status != types.True {
-					condition.Transition = timestamp
-					var defaultConditionReason string
-					var defaultConditionMessage string
-					for j := range c.config.DefaultConditions {
-						defaultCondition := &c.config.DefaultConditions[j]
-						if defaultCondition.Type == result.Rule.Condition {
-							defaultConditionReason = defaultCondition.Reason
-							defaultConditionMessage = defaultCondition.Message
-							break
-						}
+				// The condition reason specified in the rule and the result message
+				// represent the problem happened. We need to know the default condition
+				// from the config, so that we can set the new condition reason/message
+				// back when such problem goes away.
+				var defaultConditionReason string
+				var defaultConditionMessage string
+				for j := range c.config.DefaultConditions {
+					defaultCondition := &c.config.DefaultConditions[j]
+					if defaultCondition.Type == result.Rule.Condition {
+						defaultConditionReason = defaultCondition.Reason
+						defaultConditionMessage = defaultCondition.Message
+						break
 					}
+				}
 
-					inactiveProblemEvents = append(inactiveProblemEvents, util.GenerateConditionChangeEvent(
-						condition.Type,
-						status,
-						defaultConditionReason,
-						timestamp,
-					))
-
-					condition.Status = status
-					condition.Message = defaultConditionMessage
-					condition.Reason = defaultConditionReason
+				needToUpdateCondition := true
+				var newReason string
+				var newMessage string
+				status := toConditionStatus(result.ExitStatus)
+				if condition.Status == types.True && status != types.True {
+					// Scenario 1: Condition status changes from True to False/Unknown
+					newReason = defaultConditionReason
+					if newMessage == "" {
+						newMessage = defaultConditionMessage
+					} else {
+						newMessage = result.Message
+					}
 				} else if condition.Status != types.True && status == types.True {
-					// change 2: Condition status change from False/Unknown to True
-					condition.Transition = timestamp
-					condition.Message = result.Message
-					activeProblemEvents = append(activeProblemEvents, util.GenerateConditionChangeEvent(
-						condition.Type,
-						status,
-						result.Rule.Reason,
-						timestamp,
-					))
-
-					condition.Status = status
-					condition.Reason = result.Rule.Reason
+					// Scenario 2: Condition status changes from False/Unknown to True
+					newReason = result.Rule.Reason
+					newMessage = result.Message
 				} else if condition.Status != status {
-					// change 3: Condition status change from False to Unknown or vice versa
-					condition.Transition = timestamp
-					condition.Message = result.Message
-					inactiveProblemEvents = append(inactiveProblemEvents, util.GenerateConditionChangeEvent(
-						condition.Type,
-						status,
-						result.Rule.Reason,
-						timestamp,
-					))
-
-					condition.Status = status
-					condition.Reason = result.Rule.Reason
-				} else if condition.Status == status &&
+					// Scenario 3: Condition status changes from False to Unknown or vice versa
+					newReason = defaultConditionReason
+					if newMessage == "" {
+						newMessage = defaultConditionMessage
+					} else {
+						newMessage = result.Message
+					}
+				} else if condition.Status == types.True && status == types.True &&
 					(condition.Reason != result.Rule.Reason ||
 						(*c.config.PluginGlobalConfig.EnableMessageChangeBasedConditionUpdate && condition.Message != result.Message)) {
-					// change 4: Condition status do not change.
+					// Scenario 4: Condition status does not change and it stays true.
 					// condition reason changes or
 					// condition message changes when message based condition update is enabled.
+					newReason = result.Rule.Reason
+					newMessage = result.Message
+				} else {
+					// Scenario 5: Condition status does not change and it stays False/Unknown.
+					// This should just be the default reason or message (as a consequence
+					// of scenario 1 and scenario 3 above).
+					needToUpdateCondition = false
+				}
+
+				if needToUpdateCondition {
 					condition.Transition = timestamp
-					condition.Reason = result.Rule.Reason
-					condition.Message = result.Message
+					condition.Status = status
+					condition.Reason = newReason
+					condition.Message = newMessage
+
 					updateEvent := util.GenerateConditionChangeEvent(
 						condition.Type,
 						status,
-						condition.Reason,
+						newReason,
 						timestamp,
 					)
-					if condition.Status == types.True {
+
+					if status == types.True {
 						activeProblemEvents = append(activeProblemEvents, updateEvent)
 					} else {
 						inactiveProblemEvents = append(inactiveProblemEvents, updateEvent)
--- a/pkg/custompluginmonitor/types/config.go
+++ b/pkg/custompluginmonitor/types/config.go
@@ -141,5 +141,22 @@ func (cpc CustomPluginConfig) Validate() error {
 		}
 	}
 
+	for _, rule := range cpc.Rules {
+		if rule.Type != types.Perm {
+			continue
+		}
+		conditionType := rule.Condition
+		defaultConditionExists := false
+		for _, cond := range cpc.DefaultConditions {
+			if conditionType == cond.Type {
+				defaultConditionExists = true
+				break
+			}
+		}
+		if !defaultConditionExists {
+			return fmt.Errorf("Permanent problem %s does not have preset default condition.", conditionType)
+		}
+	}
+
 	return nil
 }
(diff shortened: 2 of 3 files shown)


pkg/custompluginmonitor/custom_plugin_monitor.go
```diff
@@ -160,76 +160,68 @@ func (c *customPluginMonitor) generateStatus(result cpmtypes.Result) *types.Stat
...
-						result.Rule.Reason,
-						timestamp,
-					))
-
-					condition.Status = status
-					condition.Reason = result.Rule.Reason
-				} else if condition.Status == status &&
+					// Scenario 3: Condition status changes from False to Unknown or vice versa
+					newReason = defaultConditionReason
+					newMessage = defaultConditionMessage
```
wangzhen127: @Random-Liu @andyxning 
This is different from previous implementation. I think it makes more sense to set to default reason and message here. And unknow status is somewhat undefined. So I don't think there will be any compatibility issue.

WDYT?

pkg/custompluginmonitor/types/config.go
```diff
@@ -141,5 +141,22 @@ func (cpc CustomPluginConfig) Validate() error {
...
+		conditionType := rule.Condition
+		defaultConditionExists := false
+		for _, cond := range cpc.DefaultConditions {
+			if conditionType == cond.Type {
+				defaultConditionExists = true
+				break
+			}
+		}
+		if !defaultConditionExists {
+			return fmt.Errorf("Permanent problem %s does not have preset default condition.", conditionType)
```
grosser: awesome!

pkg/custompluginmonitor/types/config.go
```diff
@@ -141,5 +141,22 @@ func (cpc CustomPluginConfig) Validate() error {
 		}
 	}
 
+	for _, rule := range cpc.Rules {
```
grosser: maybe a `validateDefaultConditionExists` method ?

pkg/custompluginmonitor/custom_plugin_monitor.go
```diff
@@ -160,73 +160,62 @@ func (c *customPluginMonitor) generateStatus(result cpmtypes.Result) *types.Stat
...
-					var defaultConditionReason string
-					var defaultConditionMessage string
-					for j := range c.config.DefaultConditions {
-						defaultCondition := &c.config.DefaultConditions[j]
-						if defaultCondition.Type == result.Rule.Condition {
-							defaultConditionReason = defaultCondition.Reason
-							defaultConditionMessage = defaultCondition.Message
-							break
-						}
+				var defaultConditionReason string
```
grosser: some comment about what is going on in that block ... or even extract a method
`// find default condition`

pkg/custompluginmonitor/custom_plugin_monitor.go
```diff
@@ -160,73 +160,62 @@ func (c *customPluginMonitor) generateStatus(result cpmtypes.Result) *types.Stat
...
-
-					condition.Status = status
-					condition.Message = defaultConditionMessage
-					condition.Reason = defaultConditionReason
+				needToUpdateCondition := true
+				var newReason string
+				var newMessage string
+				status := toConditionStatus(result.ExitStatus)
+				if condition.Status == types.True && status != types.True {
+					// Scenario 1: Condition status changes from True to False/Unknown
```
grosser: some more enduser centered docs would be nice like
```
// Scenario 1: status recovered (True -> False/Unknown), switch it back to the default
```

grosser:
logic looks correct to me, I have the feeling there is a more logical ordering for the if/else, like first going over all the `status == True / != True`, but should work either way 👍

grosser:
might be better to turn the new validation into it's own PR so that it can be merged more quickly (also needs a test ?)

pkg/custompluginmonitor/custom_plugin_monitor.go
wangzhen127: Added some comment

pkg/custompluginmonitor/types/config.go
wangzhen127: I think either way is fine. I will just keep it here as I feel this is still simple.

pkg/custompluginmonitor/custom_plugin_monitor.go
wangzhen127: This logic is very complex. We have been thinking about improving it. This CL is more like a band-aid. So I will just keep the doc and probably improve the whole logic later.

grosser:
this bot is annoying ... there is commit status for that :/

Issue #296, opened 2019-06-22
Add e2e test for NPD metric-only mode


xueweiz:
Today, NPD has two test pipelines: [CI](https://github.com/kubernetes/test-infra/blob/83ebd1d386d3c6c67c2b28bfd3f7ecf8292f2814/config/jobs/kubernetes/node-problem-detector/node-problem-detector-ci.yaml) and [presubmits](https://github.com/kubernetes/test-infra/blob/83ebd1d386d3c6c67c2b28bfd3f7ecf8292f2814/config/jobs/kubernetes/node-problem-detector/node-problem-detector-presubmits.yaml).

In both cases, they are essentially running unit tests (`make test`) and k8s e2e test (`kubernetes_e2e.py ... --ginkgo.focus=NodeProblemDetector`).

We should add more NPD-focused tests to ensure it does not regress. And many of them are best done from e2e tests, e.g.:

- Trigger/simulate failures (`kill -9 [docker]`, `echo c > /proc/sysrq-trigger`, `stress --vm 2 --vm-bytes 128M --timeout 120s`, `echo "OOM" > /dev/kmsg`, etc), and see if NPD can correctly behave under the situation.

- Performance test. e.g. write a busy program spamming "OOM" to `/dev/kmsg`, and see if NPD's CPU usage is capped. (We could use cgroup to limit NPD's usage to make the test pass)

- Soak test. e.g. Let NPD run for ~5 days, then see if it has an expanded memory footprint.

Obviously, these tests are best done in e2e tests, rather than unit test. Also, I think we should do the tests on standalone machines, rather than in a k8s cluster. Because the test should focus on node-level behavior, rather than cluster-level.

Action items:

- [ ] Add basic logistic e2e test to ensure NPD reports some expected Prometheus metrics and Stackdriver metrics: https://github.com/kubernetes/node-problem-detector/pull/323, https://github.com/kubernetes/node-problem-detector/pull/352
- [ ] Add performance e2e test to ensure NPD can handles medium/large log volume:
- [ ] Setup CI jobs in Prow to run NPD e2e tests on different distros (cos-73-lts, cos-77-lts, ubuntu-1804-lts): https://github.com/kubernetes/node-problem-detector/pull/349, #353, https://github.com/kubernetes/test-infra/pull/14372, https://github.com/kubernetes/test-infra/pull/14375, https://github.com/kubernetes/test-infra/pull/14376, https://github.com/kubernetes/test-infra/pull/14408
[...]

PR #324, opened 2019-08-10
vendor changes for e2e tests

xueweiz:
This is the vendor change part for #323 , which is part of #296 

I separate out the vendor changes to make review easier. Since vendor changes are easier to review and get merged. We can then focus on the actual code changes at #323 .


wangzhen127:
I am now fine with having vendor code along with the actual code change in the same PR, as long as they are separate commits. So I will review #323 in favor of this PR.

xueweiz:
Sounds good. :) Let me close this PR. We can reopen it if you want.

(The reason that I prefer to separate them, is because if I click "Files changed" for the PR, my browser will start to freeze. I have to filter the page by commits soon, otherwise the webpage will freeze.

Issue #329, opened 2019-08-14
Wrong type for host_uptime metric


xueweiz:
`host_uptime` metrics records the uptime of the OS in seconds, and is currently stored as an int64. We should store it at float64 to match the standard other projects use. e.g.:
1. prometheus node_exporter [example](https://github.com/prometheus/node_exporter/blob/070e4b2e17750941e38ebb4037d67da3a05a6788/collector/supervisord.go#L146)
2. Stackdriver monitoring agent [example](https://cloud.google.com/monitoring/api/metrics_agent#agent-agent)

xueweiz:
We eventually decided to keep `host_uptime` as `int64`:

> because the metric collection granularity is about once-per-minute, having an uptime measurement with sub-second granularity doesn't really help much at here.

PR #330, opened 2019-08-14
Use float64 for host_uptime metric

xueweiz:
This is for #329

### DIFF
--- a/pkg/systemstatsmonitor/host_collector.go
+++ b/pkg/systemstatsmonitor/host_collector.go
@@ -27,7 +27,7 @@ import (
 
 type hostCollector struct {
 	tags   map[string]string
-	uptime *metrics.Int64Metric
+	uptime *metrics.Float64Metric
 }
 
 func NewHostCollectorOrDie(hostConfig *ssmtypes.HostStatsConfig) *hostCollector {
@@ -46,7 +46,7 @@ func NewHostCollectorOrDie(hostConfig *ssmtypes.HostStatsConfig) *hostCollector 
 	hc.tags["os_version"] = osVersion
 
 	if hostConfig.MetricsConfigs["host/uptime"].DisplayName != "" {
-		hc.uptime, err = metrics.NewInt64Metric(
+		hc.uptime, err = metrics.NewFloat64Metric(
 			hostConfig.MetricsConfigs["host/uptime"].DisplayName,
 			"The uptime of the operating system",
 			"second",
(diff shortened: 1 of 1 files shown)


andyxning:
Since the uptime for JVM is int64 whereas the uptime for redis is double in https://cloud.google.com/monitoring/api/metrics_agent#agent-agent.

However, i think it is ok to use int because when we use `uptime` command line on Linux the output is on seconds precision. Int should be intuitive.

xueweiz:
Thanks for looking into this Ning!

> the uptime for JVM is int64

Do you mean the [`getRuntimeMXBean().getUptime()`](https://docs.oracle.com/javase/1.5.0/docs/api/java/lang/management/RuntimeMXBean.html) function? I verified that it returns a `long` typed data. But, the data in milliseconds, not seconds. So it's still a finer granularity.

When I tried to figure out whether to report it as float64 or int64 today, I was mainly checking existing system monitoring agents. I'm mostly familiar with two of them: 1. prometheus/node_exporter 2. Stackdriver Monitoring agent (which is essentially a collectd agent)

Both of them report system/agent uptime in seconds, as a double-typed variable. And that's what convinced me that the "standard" way should be double-typed.

I can check a few more agents and see what type of data do they report if you'd like :)

Also, I wonder do you know any downside of reporting uptime in float64 rather than int64?

Thanks!

xueweiz:
Oh! I understand what you mean by "the uptime for JVM is int64" now. You were talking about the metric "The uptime of the Java virtual machine. Sampled every 60 seconds" at this page: https://cloud.google.com/monitoring/api/metrics_agent#agent-agent

That's a good point. I see various types being used in that page. Now I'm not 100% convinced what type should it be :)
I will discuss with Zhen and Lantao tomorrow and see what they think.
I wonder what's your opinion on this? Do you have a preference? Thanks!

andyxning:
> I wonder do you know any downside of reporting uptime in float64 rather than int64?

I am just wondering whether int is more convenient in comparing or computing when configuring an alarm in alertmanager. An int value in milliseconds sounds good to me. :)

xueweiz:
I see. I'm also a bit leaning towards to use `int` with "second" now.
My primary reason is: because the metric collection granularity is about once-per-minute, having an uptime measurement with sub-second granularity doesn't really help much at here.

andyxning:
> because the metric collection granularity is about once-per-minute, having an uptime measurement with sub-second granularity doesn't really help much at here.

Agreed.

xueweiz:
Closing this PR. Let's use `int64`.
Thanks for the review @andyxning !

PR #335, opened 2019-08-19
Add Stackdriver exporter

xueweiz:
Tested via:
```
TARBALL=npd.tar.gz make build-tar

PROJECT=xueweiz-experimental
INSTANCE=npd-test
gcloud compute --project=$PROJECT scp npd.tar.gz $INSTANCE:/tmp/npd.tar.gz
gcloud compute --project=$PROJECT scp ./test/e2e-install.sh $INSTANCE:/tmp/e2e-install.sh
gcloud compute --project=$PROJECT ssh $INSTANCE -- sudo bash /tmp/e2e-install.sh -t /tmp/npd.tar.gz install
gcloud compute --project=$PROJECT ssh $INSTANCE -- sudo bash
```

The Stackdriver exporter can be enabled via passing the `--exporter.stackdriver` option to NPD.
The Stackdriver exporter is included in NPD by default. Set environment variable `BUILD_TAGS=disable_stackdriver_exporter` when compiling NPD to disable it at compile time.

This is for #346


pkg/exporters/stackdriver/config/config.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+var (
+	defaultExportPeriodString         = (60 * time.Second).String()
+	defaultEndpoint                   = "monitoring.googleapis.com:443"
+	defaultMetadataFetchTimeoutString = (600 * time.Second).String()
+)
+
+type StackdriverExporterConfig struct {
+	ExportPeriodString          string            `json:"exportPeriod"`
```
wangzhen127: I think if you use yaml instead of json, it can unmarshal `10s` to `time.Duration` automatically. Then you do not need a `XYZString` field.

pkg/exporters/stackdriver/metadata/type.go
```diff
@@ -0,0 +1,67 @@
...
+*/
+
+package metadata
+
+import (
+	"cloud.google.com/go/compute/metadata"
+	"github.com/golang/glog"
+)
+
+type Metadata struct {
```
wangzhen127: Does this only apply to GCP? Is so, name it more specifically. This also applies to the metadata field in StackdriverExporterConfig.

config/stackdriver-exporter.json
```diff
@@ -0,0 +1,3 @@
+{
+	"exportPeriod": "60s"
```
wangzhen127: Make an exporter folder under config

cmd/options/options.go
```diff
@@ -118,6 +128,15 @@ func (npdo *NodeProblemDetectorOptions) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&npdo.PrometheusServerAddress, "prometheus-address",
 		"127.0.0.1", "The address to bind the Prometheus scrape endpoint.")
 
+	for _, exporterName := range exporters.GetExporterNames() {
+		fs.StringSliceVar(
+			npdo.ExporterConfigPaths[exporterName],
+			"exporter."+string(exporterName),
+			[]string{},
+			fmt.Sprintf("Comma separated configurations for %v exporter. %v",
```
wangzhen127: For exporters, we just need one config file per exporter, right? Or should we enforce that? Do you expect we would have multiple stackdriver exporters?

pkg/exporters/stackdriver/stackdriver_exporter.go
```diff
@@ -0,0 +1,153 @@
...
+
+var NPDMetricToSDMetric = map[metrics.MetricID]string{
+	metrics.HostUptimeID:      "compute.googleapis.com/guest/system/uptime",
+	metrics.ProblemCounterID:  "compute.googleapis.com/guest/system/problem_count",
+	metrics.DiskAvgQueueLenID: "compute.googleapis.com/guest/disk/queue_length",
+	metrics.DiskIOTimeID:      "compute.googleapis.com/guest/disk/io_time",
+	metrics.DiskWeightedIOID:  "compute.googleapis.com/guest/disk/weighted_io_time",
+}
+
+var metadataFetchInterval = 10 * time.Second
```
wangzhen127: You have fetch timeout in the config, why not make this also configurable? Or why not make both un-configurable?

pkg/exporters/stackdriver/stackdriver_exporter.go
```diff
@@ -0,0 +1,153 @@
...
+		return fallbackMetricType
+	}
+	stackdriverMetricType, ok := NPDMetricToSDMetric[metricID]
+	if !ok {
+		return fallbackMetricType
+	}
+	return stackdriverMetricType
+}
+
+func (se *stackdriverExporter) setupViewExporterOrDie() {
```
wangzhen127: What is view exporter? Is that a thing? Or call it setupOpenCensusViewExporter (indicated from the fatalf message below?

Just curious, are there other kinds of exporters for stackdriver?

pkg/systemstatsmonitor/host_collector.go
```diff
@@ -26,13 +26,12 @@ import (
 )
 
 type hostCollector struct {
-	tags       map[string]string
-	uptime     *metrics.Int64Metric
-	lastUptime int64
+	tags   map[string]string
```
wangzhen127: Are the metrics changes related to stackdriver? If not, do you mind making them a separate commit or PR, so that it is easier to review?

Same question applies to other metrics related file changes.

cmd/options/options.go
xueweiz: Good point. Changed the code to only access one config file per exporter.

test/e2e-install.sh
```diff
@@ -54,6 +54,9 @@ function install-npd() {
   # Below remount is to work around COS's noexec mount on /home.
   mount -o remount,exec "${BIN_DIR}"
 
+  echo "Shutting down NPD service if it exist."
```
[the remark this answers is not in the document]
xueweiz: Not really useful. Removed this.

pkg/exporters/stackdriver/config/config.go
xueweiz: Thanks for the tips!
Since all other config files are in json format, I think it's safer to keep it like that. Do you think it's worth it to switch to yaml?

pkg/exporters/stackdriver/metadata/type.go
xueweiz: SG. Changed the package name to `gce`.
So when we use this type, it will look like `gce.Metadata`.

config/stackdriver-exporter.json
xueweiz: Done.

pkg/exporters/stackdriver/stackdriver_exporter.go
xueweiz: I think controlling the timeout is more important, so I only made timeout configurable.
Configuring the fetch interval won't be that much useful. But I agree that they should be treated equal.
Added this to the config.

pkg/exporters/stackdriver/stackdriver_exporter.go
xueweiz: [`view`](https://opencensus.io/stats/view/) is an OpenCensus concept, it essentially is the representation of a `metric` in OpenCensus.

[`exporter`](https://opencensus.io/exporters/supported-exporters/go/stackdriver/#creating-the-exporter) is also an OpenCensus concept, it essentially represents the method for sending metrics to the correct destination.

There is no such thing as "view exporter" though :P
In this made up word of mine, `view` is the subject that's being exported :P

Let me rename it to `setupOpenCensusViewExporter`.

pkg/systemstatsmonitor/host_collector.go
xueweiz: Yes, this refactoring is related to the Stackdriver PR. Here is the incentive:

1. The OpenCensus Stackdriver exporter determines the metric endpoint [1] from view name [2] by calling the [`GetMetricType()` option](https://godoc.org/contrib.go.opencensus.io/exporter/stackdriver#Options).

1. So we must keep a mapping from view name to metric endpoint name.

1. Before this refactoring, each individual problem daemon defines their own view name at run time (specified by the config files specified for them). It's hard to keep a global record for the above mapping.

This refactoring is simply to track that mapping.

I separated it into a separate commit.

[1] "compute.googleapis.com/guest/xxx", i.e. the name recognized by Stackdriver Monitoring service.
[2] "disk/io_time", i.e. the `viewName` field we used when calling `NewFloat64Metric()`

cmd/nodeproblemdetector/node_problem_detector.go
```diff
@@ -54,21 +56,25 @@ func main() {
...
 
 	// Initialize exporters.
-	exporters := []types.Exporter{}
+	initializedExporters := []types.Exporter{}
 	if ke := k8sexporter.NewExporterOrDie(npdo); ke != nil {
-		exporters = append(exporters, ke)
+		initializedExporters = append(initializedExporters, ke)
 		glog.Info("K8s exporter started.")
 	}
 	if pe := prometheusexporter.NewExporterOrDie(npdo); pe != nil {
```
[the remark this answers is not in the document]
xueweiz: I will try this tomorrow. Will update soon.

xueweiz:
Hi Zhen, thanks for the review!
I just addressed to most of the items (except for making existing exporters into plugins, which I will look into tomorrow). Could you help take another look? Thanks!

pkg/exporters/stackdriver/config/config.go
wangzhen127: I don't think that's a problem myself. Maybe ping @Random-Liu to check if there is any history or it does not matter?

xueweiz: Done. Currently only storing a string formatted `ExportPeriod`, and parses it at run time.

Filed #344 for yaml support, which will be done in a separate PR.

cmd/nodeproblemdetector/node_problem_detector.go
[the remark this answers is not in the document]
xueweiz: I need a little more investigation. Will track the effort at #345 .

cmd/options/options.go
```diff
@@ -94,30 +97,38 @@ func NewNodeProblemDetectorOptions() *NodeProblemDetectorOptions {
...
+		"The address to bind the node problem detector server.")
+
+	fs.IntVar(&npdo.PrometheusServerPort, "prometheus-port", 20257,
+		"The port to bind the Prometheus scrape endpoint. Prometheus exporter is enabled by default at port 20257. Use 0 to disable.")
+	fs.StringVar(&npdo.PrometheusServerAddress, "prometheus-address", "127.0.0.1",
+		"The address to bind the Prometheus scrape endpoint.")
+
+	for _, exporterName := range exporters.GetExporterNames() {
+		exporterHandler := exporters.GetExporterHandlerOrDie(exporterName)
+		exporterHandler.Options.SetFlags(fs)
```
wangzhen127: Can you move the content of `SetFlags` here? It is harder to read the code to understand what flag is added in different places.

config/exporter/stackdriver-exporter.json
```diff
@@ -0,0 +1,10 @@
+{
+	"apiEndpoint": "staging-monitoring.sandbox.googleapis.com:443",
+	"exportPeriod": "60s",
+	"gceMetadata": {
+		"projectID": "xueweiz-experimental",
```
wangzhen127: Better not use personal project. What's the purpose of this config file? Is it just for demonstration? Will it be used in production?

pkg/exporters/stackdriver/config/config.go
```diff
@@ -0,0 +1,55 @@
...
+	defaultExportPeriod          = (60 * time.Second).String()
+	defaultEndpoint              = "monitoring.googleapis.com:443"
+	defaultMetadataFetchTimeout  = (600 * time.Second).String()
+	defaultMetadataFetchInterval = (10 * time.Second).String()
+)
+
+type StackdriverExporterConfig struct {
+	ExportPeriod                string       `json:"exportPeriod"`
+	APIEndpoint                 string       `json:"apiEndpoint"`
+	Metadata                    gce.Metadata `json:"gceMetadata"`
```
wangzhen127: s/Metadata/GCEMetadata

pkg/exporters/stackdriver/config/config.go
```diff
@@ -0,0 +1,55 @@
...
+	MetadataFetchTimeout        string       `json:"metadataFetchTimeout"`
+	MetadataFetchInterval       string       `json:"metadataFetchInterval"`
+	PanicOnMetadataFetchFailure bool         `json:"panicOnMetadataFetchFailure"`
```
wangzhen127: Shall we put the default values of the above 3 config options in the config file, so that users know this can be configured?

pkg/exporters/stackdriver/stackdriver_exporter.go
```diff
@@ -0,0 +1,177 @@
...
+		glog.Fatalf("Failed to parse ExportPeriod %q: %v", se.config.ExportPeriod, err)
+	}
+
+	view.SetReportingPeriod(exportPeriod)
+	view.RegisterExporter(viewExporter)
+}
+
+func (se *stackdriverExporter) populateMetadataOrDie() {
+	if se.config.Metadata.IsValid() {
+		return
```
wangzhen127: When it is invalid, why should we just return? Should we die here?

pkg/exporters/stackdriver/gce/type.go
```diff
@@ -0,0 +1,67 @@
...
+)
+
+type Metadata struct {
+	ProjectID    string `json:"projectID"`
+	Zone         string `json:"zone"`
+	InstanceID   string `json:"instanceID"`
+	InstanceName string `json:"instanceName"`
+}
+
+func (md *Metadata) IsValid() bool {
```
wangzhen127: Better return error to indicate which metadata is missing.

pkg/exporters/stackdriver/stackdriver_exporter.go
```diff
@@ -0,0 +1,177 @@
...
+	metrics.DiskAvgQueueLenID: "compute.googleapis.com/guest/disk/queue_length",
+	metrics.DiskIOTimeID:      "compute.googleapis.com/guest/disk/io_time",
+	metrics.DiskWeightedIOID:  "compute.googleapis.com/guest/disk/weighted_io_time",
+}
+
+func getMetricType(view *view.View) string {
+	viewName := view.Measure.Name()
+	// When there is no pre-defined Stackdriver Metric Type, fallback to custom metrics.
+	// e.g. custom.googleapis.com/npd/host/uptime
+	fallbackMetricType := "custom.googleapis.com/npd/" + viewName
```
wangzhen127: If custom metrics can incur extra cost, shall we make this configurable and default to drop them instead of using custom metrics?

pkg/util/metrics/metric_id.go
```diff
@@ -0,0 +1,70 @@
...
+	HostUptimeID      MetricID = "host/uptime"
+)
+
+var MetricMap MetricMapping
+
+func init() {
+	MetricMap.mapMutex.Lock()
+	MetricMap.metricIDToViewNameMap = make(map[MetricID]string)
+	MetricMap.viewNameToMetricIDMap = make(map[string]MetricID)
+	MetricMap.mapMutex.Unlock()
```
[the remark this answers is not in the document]
xueweiz: Thanks! Done.

pkg/problemmetrics/problem_metrics.go
```diff
@@ -49,7 +49,8 @@ func NewProblemMetricsManagerOrDie() *ProblemMetricsManager {
 
 	var err error
 	pmm.problemCounter, err = metrics.NewInt64Metric(
-		"problem_counter",
+		metrics.ProblemCounterID,
+		string(metrics.ProblemCounterID),
```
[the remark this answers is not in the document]
xueweiz: The `metrics.ProblemCounterID` was defined as a const with type `metrics.MetricID`, so it actually needs a type conversion. If not, compiler will complain like this:

```
xueweiz@xueweiz:~/.gvm/pkgsets/go1.11/global/src/k8s.io/node-problem-detector$ make test
GO111MODULE=on go list -mod vendor -tags """ journald" ./... | \
	grep -v "./vendor/*" | \
	GO111MODULE=on xargs go vet -mod vendor -tags """ journald"
# k8s.io/node-problem-detector/pkg/problemmetrics
pkg/problemmetrics/problem_metrics.go:51:50: cannot use metrics.ProblemCounterID (type metrics.MetricID) as type string in argument to metrics.NewInt64Metric
# k8s.io/node-problem-detector/pkg/problemmetrics [k8s.io/node-problem-detector/pkg/problemmetrics.test]
pkg/problemmetrics/problem_metrics.go:51:50: cannot use metrics.ProblemCounterID (type metrics.MetricID) as type string in argument to metrics.NewInt64Metric
make: *** [Makefile:79: vet] Error 123
```

pkg/util/metrics/metric_id.go
```diff
@@ -0,0 +1,70 @@
+/*
```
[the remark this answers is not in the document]
xueweiz: SG :) Thanks!

pkg/util/metrics/metric_id.go
```diff
@@ -0,0 +1,70 @@
...
+
+func (mm *MetricMapping) AddMapping(metricID MetricID, viewName string) {
+	mm.mapMutex.Lock()
+	defer mm.mapMutex.Unlock()
+
+	mm.metricIDToViewNameMap[metricID] = viewName
+	mm.viewNameToMetricIDMap[viewName] = metricID
+}
+
+func (mm *MetricMapping) MetricIDToViewName(metricID MetricID) (string, bool) {
```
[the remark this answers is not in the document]
xueweiz: Good point. No one was calling it. I added it because my intention was to provide a two-way mapping. Since no one uses it, let's remove it.

Makefile
```diff
@@ -41,7 +41,7 @@ PKG:=k8s.io/node-problem-detector
 PKG_SOURCES:=$(shell find pkg cmd -name '*.go')
 
 # TARBALL is the name of release tar. Include binary version by default.
-TARBALL:=node-problem-detector-$(VERSION).tar.gz
+TARBALL?=node-problem-detector-$(VERSION).tar.gz
```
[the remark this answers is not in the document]
xueweiz: This makes manual testing much easier.

To test it on a GCE VM, I run below commands:
```
TARBALL=npd.tar.gz make build-tar

PROJECT=xueweiz-experimental
INSTANCE=npd-test
gcloud compute --project=$PROJECT scp npd.tar.gz $INSTANCE:/tmp/npd.tar.gz
gcloud compute --project=$PROJECT scp ./test/e2e-install.sh $INSTANCE:/tmp/e2e-install.sh
gcloud compute --project=$PROJECT ssh $INSTANCE -- sudo bash /tmp/e2e-install.sh -t /tmp/npd.tar.gz install
```

I define the name of the tarball by setting the TARBALL environment variable. If we cannot do so, the tarball's name will be different each time, and I have to use some fancy logic to find the tarball.

pkg/exporters/stackdriver/config/config.go
xueweiz: Done.

pkg/exporters/stackdriver/config/config.go
xueweiz: SG :) done.

pkg/exporters/stackdriver/stackdriver_exporter.go
xueweiz: :P sorry for the confusing naming. This `if` clause actually triggers when the metadata is valid.

Basically we want this:
1. If the user defined a set of valid metadata in the config file, we want NPD to respect it and use it.
2. If the config file don't have any metadata, or they are not valid, we want NPD to [query GCE metadata server](https://cloud.google.com/compute/docs/storing-retrieving-metadata#querying), and collect them at runtime.

This `if` clause is for use case #1.

pkg/exporters/stackdriver/gce/type.go
xueweiz: I don't really think returning the list of missing fields would be useful.

This is because if any field is missing, we will just call `PopulateFromGCE` and fill it. If `PopulateFromGCE` failed, it will generate an error message then.

I changed the function to `HasMissingField()`, maybe that can help make the logic a bit clearer?

pkg/exporters/stackdriver/stackdriver_exporter.go
xueweiz: SG. Done.

cmd/options/options.go
```diff
@@ -85,7 +86,9 @@ type NodeProblemDetectorOptions struct {
 }
 
 func NewNodeProblemDetectorOptions() *NodeProblemDetectorOptions {
-	npdo := &NodeProblemDetectorOptions{MonitorConfigPaths: types.ProblemDaemonConfigPathMap{}}
+	npdo := &NodeProblemDetectorOptions{
+		MonitorConfigPaths: types.ProblemDaemonConfigPathMap{}}
```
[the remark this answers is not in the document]
xueweiz: Done.

cmd/options/options.go
```diff
@@ -94,30 +97,38 @@ func NewNodeProblemDetectorOptions() *NodeProblemDetectorOptions {
...
+		"The interval between the checks on the readiness of kube-apiserver. This is ignored if --enable-k8s-exporter is false.")
+	fs.IntVar(&npdo.ServerPort, "port", 20256,
+		"The port to bind the node problem detector server. Use 0 to disable.")
+	fs.StringVar(&npdo.ServerAddress, "address", "127.0.0.1",
+		"The address to bind the node problem detector server.")
+
+	fs.IntVar(&npdo.PrometheusServerPort, "prometheus-port", 20257,
+		"The port to bind the Prometheus scrape endpoint. Prometheus exporter is enabled by default at port 20257. Use 0 to disable.")
+	fs.StringVar(&npdo.PrometheusServerAddress, "prometheus-address", "127.0.0.1",
+		"The address to bind the Prometheus scrape endpoint.")
```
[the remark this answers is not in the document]
xueweiz: Sorry, I was grouping them by their "plugin". Reverted.

cmd/options/options.go
xueweiz: Actually I created the `Options.SetFlags()` abstraction to allow each exporter plugin to define their own flags flexibly.

This is for #345 , and later I will send a PR to create `ExporterHandler` for k8s_exporter and prometheus_exporter. And in their `SetFlags()` function, existing options we have today will be added:
```
	fs.BoolVar(&npdo.EnableK8sExporter, "enable-k8s-exporter", true, "Enables reporting to Kubernetes API server.")
	fs.StringVar(&npdo.ApiServerOverride, "apiserver-override", "",
		"Custom URI used to connect to Kubernetes ApiServer. This is ignored if --enable-k8s-exporter is false.")
	fs.DurationVar(&npdo.APIServerWaitTimeout, "apiserver-wait-timeout", time.Duration(5)*time.Minute,
		"The timeout on waiting for kube-apiserver to be ready. This is ignored if --enable-k8s-exporter is false.")
	fs.DurationVar(&npdo.APIServerWaitInterval, "apiserver-wait-interval", time.Duration(5)*time.Second,
		"The interval between the checks on the readiness of kube-apiserver. This is ignored if --enable-k8s-exporter is false.")
	fs.IntVar(&npdo.ServerPort, "port", 20256,
		"The port to bind the node problem detector server. Use 0 to disable.")
	fs.StringVar(&npdo.ServerAddress, "address", "127.0.0.1",
		"The address to bind the node problem detector server.")
```
and 
```
	fs.IntVar(&npdo.PrometheusServerPort, "prometheus-port", 20257,
		"The port to bind the Prometheus scrape endpoint. Prometheus exporter is enabled by default at port 20257. Use 0 to disable.")
	fs.StringVar(&npdo.PrometheusServerAddress, "prometheus-address", "127.0.0.1",
		"The address to bind the Prometheus scrape endpoint.")
```

config/exporter/stackdriver-exporter.json
xueweiz: Removed. Sorry the personal project part is added when I was testing it. It's not useful anymore now.

The config file is intended to be used in production.
(Although an empty config file can also be used in production, because the current one is basically just explicitly specifying all the default options.)

xueweiz:
Hi Zhen, I just resolved the above comments and changed README. Can you help take another look? Thanks!

wangzhen127:
Can we move the last commit "Allow problem daemon plugins to define their command line options" to a different PR?

xueweiz:
SG, filed #348 for the problem daemon plugin interface change. Thanks!

xueweiz:
Thanks a lot Zhen!

PR #339, opened 2019-08-29
avoid log channel closed caused endless loop

wenjun93:
if log watcher meets error and closes logCh, log monitor will enter endless for loop

### DIFF
--- a/pkg/systemlogmonitor/log_monitor.go
+++ b/pkg/systemlogmonitor/log_monitor.go
@@ -129,7 +129,11 @@ func (l *logMonitor) monitorLoop() {
 	l.initializeStatus()
 	for {
 		select {
-		case log := <-l.logCh:
+		case log, ok := <-l.logCh:
+			if !ok {
+				glog.Errorf("Log channel closed")
+				return
+			}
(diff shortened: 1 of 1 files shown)


pkg/systemlogmonitor/log_monitor.go
```diff
@@ -129,7 +129,11 @@ func (l *logMonitor) monitorLoop() {
 	l.initializeStatus()
 	for {
 		select {
-		case log := <-l.logCh:
+		case log, ok := <-l.logCh:
+			if !ok {
+				glog.Errorf("Log channel closed")
```
Random-Liu: We should add more information in the log.

I'll send another change to address this. I want to fix something else anyway.

Random-Liu:
@wenjun93 Thanks for the fix!

Random-Liu:
@wenjun93 Can you sign CLA?

Random-Liu:
@wangzhen127 we may want to cherrypick this into release/1.0

wenjun93:
> @wenjun93 Can you sign CLA?

@Random-Liu I changed my email in the past, now corrected

Issue #345, opened 2019-09-10
consider making k8s_exporter and prometheus_exporter into plugins


xueweiz:
Making existing exporters into plugins has a few benefits:

1. It allows users to disable them at compile time, to reduce the memory usage of NPD.

1. It unifies the interface (command-line or config file) and code path for all exporters in NPD.

However a few questions needs to be answered:

1. Do we keep the CLI interface? Or do we just take in one config file for each exporter without additional command-line options?

I need a bit more investigation into this. Also, I don't this as a hard requirements right now. As I don't imagine many NPD users would like to disable k8s_exporter or prometheus_exporter at compile time.

PR #347, opened 2019-09-13
Make all existing exporters into plugins

xueweiz:
This is for #345 

A little background:
We want to make all problem daemons (system-log, system-stats, custom-plugin) and all exporters (k8s, prometheus, stackdriver) into "plugins". By "plugin", what I really mean is this:

1. NPD main code does not directly import these libraries, and does not use these libraries directly.

1. Each plugin should be able to be disabled at compile time.

The challenge comes from the existing exporters (k8s, prometheus). They have been existing for a while, and is configured via command-line flags (e.g. `--apiserver-override`, `--apiserver-wait-timeout`). So to refactor them into plugins, we had to allow plugins to define their own command-line options.

We have already implemented mechanism to allow plugins to define their own command-line interfaces in #335. This PR is to pick it up for the existing exporters (k8s and prometheus).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,3 +29,7 @@ script:
  - BUILD_TAGS="disable_system_stats_monitor" make test
  - make clean && BUILD_TAGS="disable_stackdriver_exporter" make
  - BUILD_TAGS="disable_stackdriver_exporter" make test
+ - make clean && BUILD_TAGS="disable_k8s_exporter" make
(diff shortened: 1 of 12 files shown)


xueweiz:
/assign @Random-Liu 
Hi Lantao, could you help take a look at this PR as well? I just rebased it on top of master (to accommodate #356 ).

(Again no rush :) this is just a refactoring to make things more consistent. Not blocking anything.

Issue #346, opened 2019-09-12
Add Stackdriver Exporter


xueweiz:
It's nice to have an exporter that can export the metrics to Stackdriver Monitoring if needed. The exporter will have an option to be disabled at compile time. So for the people who do not use Stackdriver at all, they disable it to avoid any side effect.

xueweiz:
This is done via #335 . Closing.

PR #348, opened 2019-09-13
Allow problem daemon plugins to define their command line options

xueweiz:
This is to keep the interface for problem daemon plugins and exporter plugins in sync.

A little background:
We want to make all problem daemons (system-log, system-stats, custom-plugin) and all exporters (k8s, prometheus, stackdriver) into "plugins". By "plugin", what I really mean is this:

1. NPD main code does not directly import these libraries, and does not use these libraries directly.

1. Each plugin should be able to be disabled at compile time.

The challenge comes from the existing exporters (k8s, prometheus). They have been existing for a while, and is configured via command-line flags (e.g. `--apiserver-override`, `--apiserver-wait-timeout`). So to refactor them into plugins, we had to allow plugins to define their own command-line options.

This flexibility (defining command-line options) is implemented in #335 as part of #346 . So now we face another question:

**Do we also change the code for problem daemon plugins so that they follow the same pattern? (i.e. allowing each plugin to define their own command-line interfaces)**

- The benefit we gain is a more unified code base.
- The down side is that problem daemons already has a pretty unified command-line interfaces (`--config.xxx`), and does not really need this flexibility. So it's not that crucial.

### DIFF
--- a/cmd/nodeproblemdetector/node_problem_detector.go
+++ b/cmd/nodeproblemdetector/node_problem_detector.go
@@ -35,7 +35,7 @@ import (
 )
 
 func main() {
-	npdo := options.NewNodeProblemDetectorOptions()
+	npdo := &options.NodeProblemDetectorOptions{}
 	npdo.AddFlags(pflag.CommandLine)
 
 	pflag.Parse()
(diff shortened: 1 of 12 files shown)


wangzhen127:
/assign @Random-Liu 
I don't have a strong opinion on this. Deferring to @Random-Liu to take a look and see if other OSS projects did similar things.

wangzhen127:
Actually, this is automatically done if we merge https://github.com/kubernetes/node-problem-detector/pull/347, right?

xueweiz:
Hi @Random-Liu , could you help take a look for this PR when you have time? Thanks :)
(No rush, this is not blocking anything.

xueweiz:
I don't think Zhen and Lantao would have time to review this in these days.
I'm going to close this PR for now. If this becomes important in the future, we can then bump it up.

PR #352, opened 2019-09-18
Set SSH timeout to 5 minutes

xueweiz:
This is for https://github.com/kubernetes/node-problem-detector/issues/296#issuecomment-532887983

### DIFF
--- a/test/e2e/lib/gce/instance.go
+++ b/test/e2e/lib/gce/instance.go
@@ -96,7 +96,8 @@ func CreateInstance(instance Instance, imageName string, imageProject string) (I
 	}
 
 	instanceRunning := false
-	for i := 0; i < 30 && !instanceRunning; i++ {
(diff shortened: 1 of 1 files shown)


test/e2e/lib/gce/instance.go
```diff
@@ -96,7 +96,7 @@ func CreateInstance(instance Instance, imageName string, imageProject string) (I
 	}
 
 	instanceRunning := false
-	for i := 0; i < 30 && !instanceRunning; i++ {
+	for i := 0; i < 15 && !instanceRunning; i++ {
```
wangzhen127: Can you add a comment before this line on the timeout (15*20=300sec=5min)?

xueweiz: Sorry. Just changed it. WDYT?

PR #356, opened 2019-09-23
Make heartbeatPeriod const into a flag

mm4tt:
This is an important parameter when it comes to kubernetes scalability (it controlls the node/status patch QPS) and it deserves to be a flag instead of a constant. 

/sig scalability

### DIFF
--- a/cmd/options/options.go
+++ b/cmd/options/options.go
@@ -57,6 +57,8 @@ type NodeProblemDetectorOptions struct {
 	// APIServerWaitInterval is the interval between the checks on the
 	// readiness of kube-apiserver.
 	APIServerWaitInterval time.Duration
+	// K8sExporterHeartbeatPeriod is the period at which the k8s exporter does forcibly sync with apiserver.
+	K8sExporterHeartbeatPeriod time.Duration
 
 	// prometheusExporter options
 	// PrometheusServerPort is the port to bind the Prometheus scrape endpoint. Use 0 to disable.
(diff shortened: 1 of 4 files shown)


pkg/exporters/k8sexporter/condition/manager.go
```diff
@@ -36,10 +37,17 @@ const (
...
+)
+
+var (
 	// heartbeatPeriod is the period at which condition manager does forcibly sync with apiserver.
-	heartbeatPeriod = 1 * time.Minute
+	heartbeatPeriod time.Duration
 )
 
+func init() {
+	flag.DurationVar(&heartbeatPeriod, "heartbeat-period", 1 * time.Minute, "heartbeatPeriod is the period at which condition manager does forcibly sync with apiserver")
```
wangzhen127: Can you add an option to NodeProblemDetectorOptions and create the flag in:
https://github.com/kubernetes/node-problem-detector/blob/master/cmd/options/options.go

mm4tt: Done.

cmd/options/options.go
```diff
@@ -83,6 +83,9 @@ type NodeProblemDetectorOptions struct {
 
 	// NodeName is the node name used to communicate with Kubernetes ApiServer.
 	NodeName string
+
+	// K8sExporterHeartbeatPeriod is the period at which the k8s exporter does forcibly sync with apiserver.
+	K8sExporterHeartbeatPeriod time.Duration
```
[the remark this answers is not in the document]
mm4tt: Sure, done.

PR #358, opened 2019-09-24
Feature support per rule interval

gmemcc:
Hello ! I have just added per rule interval support for CustemPluginMonitor, PTAL.

### DIFF
--- a/pkg/custompluginmonitor/custom_plugin_monitor.go
+++ b/pkg/custompluginmonitor/custom_plugin_monitor.go
@@ -120,7 +120,7 @@ func (c *customPluginMonitor) Stop() {
 	c.tomb.Stop()
 }
 
-// monitorLoop is the main loop of log monitor.
+// monitorLoop is the main loop of custom plugin monitor.
 func (c *customPluginMonitor) monitorLoop() {
 	c.initializeStatus()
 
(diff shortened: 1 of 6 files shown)


wangzhen127:
Can you explain why the plugin invoke interval is not enough? Why do you need rule invoke interval?

gmemcc:
> Can you explain why the plugin invoke interval is not enough? Why do you need rule invoke interval?

I think it would be better to let some heavy rules be invoked at a lower frequency.

andyxning:
IIRC, we have talked about adding rule level interval before. Will check this later.

wangzhen127:
/assign @andyxning 
/assign @Random-Liu 
/ok-to-test

Do you mean you want to use different invoke intervals for different rules of the same plugin?

xueweiz:
This is the real failure:
> unable to create gce instance with running docker daemon for image coreos-beta-1911-2-0-v20181024.  could not create instance tmp-node-e2e-fdba7006-coreos-beta-1911-2-0-v20181024: API error: googleapi: Error 400: Invalid value for field 'resource.disks[0].initializeParams.sourceImage': 'projects/coreos-cloud/global/images/coreos-beta-1911-2-0-v20181024'. The referenced image resource cannot be found., invalid

Seems the image `coreos-beta-1911-2-0-v20181024` has been removed from `coreos-cloud`. 

Same issue is also tracked at https://github.com/kubernetes/test-infra/issues/14465, and https://github.com/kubernetes/test-infra/pull/14466 should have it fixed.

/retest

PR #365, opened 2019-09-26
support goenv / gvm by auto-switching to the correct go version

grosser:
Ideally this should be verified to match the dockerfile


xueweiz:
Hi @grosser , can you point me to some documentation to this `.go-version` file? Specifically, I want to learn about: 1. Who is consuming this file. 2. What is the interface of this specification.

This would help me figure out whether this file should be checked into the repository, or be whitelisted under `.gitignore`. Thanks!

grosser:
- afaik only "go version managers" consume/write this file (maybe some CI providers too)
 - interface is "put go version into this file"
 - should be checked in so new users use the correct version when checking out the repo

... with gos new behavior of putting the go version into the go.mod https://github.com/kubernetes/node-problem-detector/blob/master/go.mod#L3 this might be no longer needed and go version managers can just read the go.mod

so I'll just close since nobody else has wanted this so far 🤷‍♂

xueweiz:
Ah I see, interesting :)

I think using the go version in `go.mod` is good, hopefully `gvm` is planning on supporting this. I agree that we probably can close this PR.

PR #369, opened 2019-10-03
Support "nf_conntrack", check for 90% full

arekkusu:
- Script was checking for "ip_conntrack_..." which was replaced by "nf_conntrack_..." on newer system. Now support both.

-  Return failure ("not ok") when table is more than 90% full. Note sure what value is best here but I think that is better than when the table is full. Otherwise we might end up with a value close to the max or bouncing around.

- Replaced cat by "$(< file )" to avoid calling external command

### DIFF
--- a/config/plugin/network_problem.sh
+++ b/config/plugin/network_problem.sh
@@ -1,23 +1,36 @@
 #!/bin/bash
 
-# This plugin checks for common network issues. Currently, it only checks
-# if the conntrack table is full. 
+# This plugin checks for common network issues.
+# Currently only checks if conntrack table is more than 90% used.
 
-OK=0
-NONOK=1
-UNKNOWN=2
+readonly OK=0
+readonly NONOK=1
+readonly UNKNOWN=2
 
-[ -f /proc/sys/net/ipv4/netfilter/ip_conntrack_max ] || exit $UNKNOWN
-[ -f /proc/sys/net/ipv4/netfilter/ip_conntrack_count ] || exit $UNKNOWN
+# "nf_conntrack" replaces "ip_conntrack" - support both
+readonly NF_CT_COUNT_PATH='/proc/sys/net/netfilter/nf_conntrack_count'
+readonly NF_CT_MAX_PATH='/proc/sys/net/netfilter/nf_conntrack_max'
+readonly IP_CT_COUNT_PATH='/proc/sys/net/ipv4/netfilter/ip_conntrack_count'
+readonly IP_CT_MAX_PATH='/proc/sys/net/ipv4/netfilter/ip_conntrack_max'
 
-conntrack_max=$(cat /proc/sys/net/ipv4/netfilter/ip_conntrack_max)
-conntrack_count=$(cat /proc/sys/net/ipv4/netfilter/ip_conntrack_count)
-
-if (( conntrack_count >= conntrack_max )); then
-    echo "Conntrack table full"
-    exit $NONOK
+if [[ -f $NF_CT_COUNT_PATH ]] && [[ -f $NF_CT_MAX_PATH ]]; then
+  readonly CT_COUNT_PATH=$NF_CT_COUNT_PATH
+  readonly CT_MAX_PATH=$NF_CT_MAX_PATH
+elif [[ -f $IP_CT_COUNT_PATH ]] && [[ -f $IP_CT_MAX_PATH ]]; then
+  readonly CT_COUNT_PATH=$IP_CT_COUNT_PATH
+  readonly CT_MAX_PATH=$IP_CT_MAX_PATH
+else
+  exit $UNKNOWN
 fi
 
-echo "Conntrack table available"
-exit $OK
+readonly conntrack_count=$(< $CT_COUNT_PATH) || exit $UNKNOWN
+readonly conntrack_max=$(< $CT_MAX_PATH) || exit $UNKNOWN
+readonly conntrack_usage_msg="${conntrack_count} out of ${conntrack_max}"
 
+if (( conntrack_count > conntrack_max * 9 /10 )); then
+  echo "Conntrack table usage over 90%: ${conntrack_usage_msg}"
+  exit $NONOK
+else
+  echo "Conntrack table usage: ${conntrack_usage_msg}"
+  exit $OK
+fi


config/plugin/network_problem.sh
```diff
@@ -1,23 +1,35 @@
 #!/bin/bash
 
 # This plugin checks for common network issues. Currently, it only checks
-# if the conntrack table is full. 
+# if the conntrack table is full.
```
xueweiz: We should also change this comment:
"# if the conntrack table is more than 90% full."

config/plugin/network_problem.sh
```diff
@@ -1,23 +1,35 @@
...
+else
+    exit $UNKNOWN
+fi
+
+conntrack_count=$(< $CONNTRACK_COUNT) || exit $UNKNOWN
+conntrack_max=$(< $CONNTRACK_MAX) || exit $UNKNOWN
 
-if (( conntrack_count >= conntrack_max )); then
-    echo "Conntrack table full"
+if (( conntrack_count > conntrack_max /10 * 9)); then
```
xueweiz: nit: Bash only does integer arithmetic. To avoid/reduce precision loss problems, it's best to change this to ` *9 /10`.

config/plugin/network_problem.sh
```diff
@@ -1,23 +1,35 @@
...
-[ -f /proc/sys/net/ipv4/netfilter/ip_conntrack_count ] || exit $UNKNOWN
+# "nf_conntrack" replaces "ip_conntrack" - support both
+NF_CONNTRACK_COUNT='/proc/sys/net/netfilter/nf_conntrack_count'
+NF_CONNTRACK_MAX='/proc/sys/net/netfilter/nf_conntrack_max'
+IP_CONNTRACK_COUNT='/proc/sys/net/ipv4/netfilter/ip_conntrack_count'
+IP_CONNTRACK_MAX='/proc/sys/net/ipv4/netfilter/ip_conntrack_max'
 
-conntrack_max=$(cat /proc/sys/net/ipv4/netfilter/ip_conntrack_max)
-conntrack_count=$(cat /proc/sys/net/ipv4/netfilter/ip_conntrack_count)
+if [ -f $NF_CONNTRACK_COUNT ] && [ -f $NF_CONNTRACK_MAX ]; then
```
xueweiz: nit: Could you help change the `[ -f xxx ]`s here and below into `[[ -f xxx ]]`? Thanks! :)
This is because `[[ ]]` is [more prefered](https://google.github.io/styleguide/shell.xml#Test,_%5B_and_%5B%5B) than `[ ]`. If you could help use this chance to help improve the code quality, I'm sure everyone would appreciate it :)

config/plugin/network_problem.sh
```diff
@@ -1,23 +1,35 @@
...
+# if the conntrack table is full.
 
 OK=0
 NONOK=1
 UNKNOWN=2
 
-[ -f /proc/sys/net/ipv4/netfilter/ip_conntrack_max ] || exit $UNKNOWN
-[ -f /proc/sys/net/ipv4/netfilter/ip_conntrack_count ] || exit $UNKNOWN
+# "nf_conntrack" replaces "ip_conntrack" - support both
+NF_CONNTRACK_COUNT='/proc/sys/net/netfilter/nf_conntrack_count'
```
xueweiz: nit: Could you add `readonly ` before the [constants](https://google.github.io/styleguide/shell.xml#Constants_and_Environment_Variable_Names) (`NF_CONNTRACK_COUNT`, `NF_CONNTRACK_MAX`...) and the [read-only variables](https://google.github.io/styleguide/shell.xml#Read-only_Variables) (`conntrack_count`, `conntrack_max`)?

Thank you!

config/plugin/network_problem.sh
```diff
@@ -1,23 +1,35 @@
...
+    CONNTRACK_COUNT=$NF_CONNTRACK_COUNT
+    CONNTRACK_MAX=$NF_CONNTRACK_MAX
+elif [ -f $IP_CONNTRACK_COUNT ] && [ -f $IP_CONNTRACK_MAX ]; then
+    CONNTRACK_COUNT=$IP_CONNTRACK_COUNT
+    CONNTRACK_MAX=$IP_CONNTRACK_MAX
+else
+    exit $UNKNOWN
+fi
+
+conntrack_count=$(< $CONNTRACK_COUNT) || exit $UNKNOWN
```
xueweiz: nit: We have `conntrack_count` and `CONNTRACK_COUNT`, one for the value and another one for the path. do you mind appending `_PATH` to the path constants? Thanks so much!

config/plugin/network_problem.sh
```diff
@@ -1,23 +1,35 @@
...
+conntrack_max=$(< $CONNTRACK_MAX) || exit $UNKNOWN
 
-if (( conntrack_count >= conntrack_max )); then
-    echo "Conntrack table full"
+if (( conntrack_count > conntrack_max /10 * 9)); then
+    echo "Conntrack table is more than 90% full"
```
xueweiz: Maybe
`echo "Conntrack table is more than 90% used: ${conntrack_count} out of ${conntrack_max}"`

What do you think?

config/plugin/network_problem.sh
xueweiz: Thanks for replacing `cat` with `< ` :)

config/plugin/network_problem.sh
arekkusu: Thanks, update the comment.

config/plugin/network_problem.sh
arekkusu: Very good point ! Updated as proposed. Thanks.

config/plugin/network_problem.sh
arekkusu: Updated to use `[[ ]]`, I also switched to 2 space indent to follow Google Shell Style Guide.

I think same should be done for `check_ntp.sh`.

config/plugin/network_problem.sh
arekkusu: I added `readonly` where necessary. 
Feels like a lot of `readonly` for a small script (readability wise) but I understand the reasoning :)

config/plugin/network_problem.sh
arekkusu: Not a nit, you're right about the naming.
I added `PATH` and shortened `CONNTRACK` to `CT`

config/plugin/network_problem.sh
arekkusu: Update to included this additional information for both `OK` and `NONOK` exit.
In addition I choose to output to `STDERR` for `NONOK`

config/plugin/network_problem.sh
arekkusu: :)

arekkusu:
Really appreciate your review and input. Have a look at the update and let me know what you think.

config/plugin/network_problem.sh
xueweiz: Thanks! Yes I agree, we should update `check_ntp.sh` as well as a small improvement.

config/plugin/network_problem.sh
xueweiz: Sounds good to me! Thanks! :)

config/plugin/network_problem.sh
xueweiz: Sorry for the trouble. Could you help change the print back to `STDOUT`? :P

This is a small detail in `custom-plugin-monitor`'s interface, where only `STDOUT` is collected, and `STDERR` is ignored. See the [design](https://docs.google.com/document/d/1jK_5YloSYtboj-DtfjmYKxfNnUxCAvohLnsH5aGCAYQ/edit#) and the [implementation](https://github.com/kubernetes/node-problem-detector/blob/8704ec0c42b27db5eca8e57c3dbfc172ba82765d/pkg/custompluginmonitor/plugin/plugin.go#L131) for this detail.

xueweiz:
Hi @arekkusu , thanks a lot for making the changes! Just one more small item for the `STDOUT` thing :) once that's fixed I'll approve and merge the PR.
Thanks!

xueweiz:
And by the way, you can (optionally) squash all the 2-3 commits into one. We typically do this to have a simpler/cleaner commit history. But I understand that a lot of people choose to not do that to keep a history for the code review.
Your choice :) I'm fine with either way.

arekkusu:
Ok regarding outputting to `STDOUT`. Updated and squashed the commits.

BTW do see any changes needed to network-problem-monitor.json`?
I guess `ConntrackFull` is no longer fully accurate ( `ConntrackPressure` or `ConntrackHighUsage`). I am not sure it's something we want / need to change.

https://github.com/kubernetes/node-problem-detector/blob/599ca532e81833debb4791c0410464507a80c869/config/network-problem-monitor.json#L1-L20

arekkusu:
I think the (draft?) design doc doesn't clearly indicate the UNKNOWN return status.

Implementation considers anything else than 0 or 1 as unknown

https://github.com/kubernetes/node-problem-detector/blob/8704ec0c42b27db5eca8e57c3dbfc172ba82765d/pkg/custompluginmonitor/plugin/plugin.go#L153-L161

This test does refer to exit code 2 as unkown

https://github.com/kubernetes/node-problem-detector/blob/8704ec0c42b27db5eca8e57c3dbfc172ba82765d/pkg/custompluginmonitor/plugin/test-data/unknown.sh#L1-L5

Let me know if you think it's better to use 3 as exit code for UNKNOWN ...

xueweiz:
> I guess `ConntrackFull` is no longer fully accurate (`ConntrackPressure` or `ConntrackHighUsage`). I am not sure it's something we want / need to change.

I agree with you, I think `ConntrackPressure` seems like a good option. However I don't know whether it's worth it to change it...Since users may have already doing detecting & remedy based on the `ConntrackFull` event that NPD reports. Changing it is more visible to the users. I don't have a strong opinion in either way. I think your PR is fine as is.
+cc @andyxning , who reviewed the original PR #152 .

> Let me know if you think it's better to use 3 as exit code for UNKNOWN ...

I think keeping it as is (which is `2`) is OK :)
(I'm quite curious why was this not clearly defined lol

/lgtm
/approve

PR #375, opened 2019-10-25
Properly close channel when monitors exit.

Random-Liu:
This is a follow up of https://github.com/kubernetes/node-problem-detector/pull/339.
Will rebase after https://github.com/kubernetes/node-problem-detector/pull/339 is merged.

This PR:
1) Handles channel close for log counter, or else we may get nil log entry and cause a panic;
2) Handles channel close for kmsg parser, or else we may have another busy loop;
3) Handles channel close for custom plugin monitor, this is less an issue because the plugin will never close its channel, but let's add this for completion;
4) Closes channels returned to the problem detector, so that the goroutine in the problem detector can exit if the corresponding problem daemon exits.

1 and 2 are a more important bug fix that we may want to cherry-pick, 3 and 4 are not that important.

Signed-off-by: Lantao Liu <lantaol@google.com>

### DIFF
--- a/cmd/logcounter/log_counter.go
+++ b/cmd/logcounter/log_counter.go
@@ -44,7 +44,11 @@ func main() {
 		fmt.Print(err)
 		os.Exit(int(types.Unknown))
 	}
-	actual := counter.Count()
+	actual, err := counter.Count()
+	if err != nil {
+		fmt.Print(err)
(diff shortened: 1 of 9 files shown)


Random-Liu:
@wangzhen127 rebased.

PR #377, opened 2019-10-28
add a case is ID="centos"  

CuZn13:
When reading "/etc/os-release" file , ID="centos", report an error “Unsupported ID in /etc/os-release:......”


CuZn13:
CLA completed

CuZn13:
CLA completed

PR #379, opened 2019-10-28
add a case is ID="centos"

CuZn13:
When reading "/etc/os-release" file , ID="centos", report an error “Unsupported ID in /etc/os-release:......”


xueweiz:
Hi @CuZn13 ,

Could you also add an unit test for parsing the "/etc/os-release" of CentOS?
You can see an example [here](https://github.com/kubernetes/node-problem-detector/blob/ad76b93208380392df5615e6a0eaca7db1b377d1/pkg/util/helpers_test.go#L158).

Thanks!

xueweiz:
Thanks for adding the unit test! It's very helpful for someone like me who don't have a CentOS machine at hand to see how is its `/etc/os-release` looks like :)

/lgtm
/cc @wangzhen127 
/cc @Random-Liu

pkg/util/testdata/os-release-centos
```diff
@@ -0,0 +1,15 @@
...
+PRETTY_NAME="CentOS Linux 7 (Core)"
+ANSI_COLOR="0;31"
+CPE_NAME="cpe:/o:centos:centos:7"
+HOME_URL="https://www.centos.org/"
+BUG_REPORT_URL="https://bugs.centos.org/"
+
+CENTOS_MANTISBT_PROJECT="CentOS-7"
+CENTOS_MANTISBT_PROJECT_VERSION="7"
+REDHAT_SUPPORT_PRODUCT="centos"
+REDHAT_SUPPORT_PRODUCT_VERSION="7"
```
xueweiz: nit: the formatter is complaining that this file does not have a trailing newline character.
So if you could add a newline character at the end, [it would be great](https://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline) :)

CuZn13: ok

xueweiz:
Hi @CuZn13 ,

I think the commit "When the state changes, it may be necessary to output customized mess…" is unintentional, right?

If you are working on some experiments, it's probably easier to create a new branch (`git checkout -b exp`), and try things out there. Your current branch (which is `master`) is tied to this PR. It's best to keep it clean :)

Also, I suggest you to merge the two "add an unit test for parsing the "/etc/os-release" of CentOS" commit into one. You can do that via running `git rebase -i HEAD^^^^`, then edit the history there.

CuZn13:
Sorry, I was careless. I'll fix it later

PR #384, opened 2019-11-08
add link for libsystemd-journal-dev install

zhangxiaoyu-zidif:


xueweiz:
I'd actually prefer to not add this link :) because the link points to a Ubuntu Trusty (14.04) repository. But the developer could be using any Ubuntu version, and this link could mislead them.

What do you think?

Also, I think providing the package name as is OK for now. If you want it to be more clear, I think a good way would be to shutdown more gracefully.

One way to do that would be to add another rule in the makefile, (e.g. "check dependency"). The rule will check whether these requirements are met, and fail with a clear & actionable error message ;)

zhangxiaoyu-zidif:
`check dependency` is fine. I prefer it also.

PR #385, opened 2019-11-13
update the testing example in README.md

bhcleek:
Fix the testing example in README.md so that it works as described to
show an event given that config/kernel-monitor-filelog.json expects a
string wrapped with square brackets to come right after `kernel: `.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,7 +202,7 @@ For more scenarios, see [here](https://github.com/kubernetes/heapster/blob/maste
 
 ## Try It Out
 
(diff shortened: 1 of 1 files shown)


andyxning:
@bhcleek 

You just add a timestamp `2019-11-13T19:28:16.288365+00:00` before `kernel: `, right? The config files are just examples, you need to adjust the rule patterns according to your `/dev/kmsg` log format since different Linux distributions have different formats. Even with the same Linux distribution, people can adjust the log format. 

What we should do maybe add a statement like: `The config files are just examples, you need to adjust the rule patterns according to your log records.`

bhcleek:
You're right as far as an accurate description of my changes go. Apparently I missed an important bit here: it was actually described in the commit message, but I failed to include the critical change. I've added amended the commit to correct that. The example now matches the message format specified in config/kernel-monitor-filelog.json

xueweiz:
Sorry, the test failure is a flake, tracked in #403 
/retest

Issue #366, opened 2019-09-26
make build fails with ENABLE_JOURNALD=0


grosser:
```
ENABLE_JOURNALD=0 make build
CGO_ENABLED=0 GOOS=linux GO111MODULE=on go build \
		-mod vendor \
		-o bin/node-problem-detector \
		-ldflags '-X k8s.io/node-problem-detector/pkg/version.version=v0.7.1-20-gc55fa33' \
		"" \
		./cmd/nodeproblemdetector
can't load package: package k8s.io/node-problem-detector: no Go files in /Users/mgrosser/Code/tools/node-problem-detector
make: *** [bin/node-problem-detector] Error 1
```

xueweiz:
Thanks for filing the bug. Uploaded #390 to fix it.
/assign @xueweiz

xueweiz:
/close
#390 should have this fixed :)

PR #390, opened 2019-11-26
Fix build tags manipulation in Makefile

xueweiz:
This is for #366 

In the current Makefile, when ENABLE_JOURNAL is set to 0, BUILD_TAGS will be set to `""`. And result in below command:
```
CGO_ENABLED=0 GOOS=linux GO111MODULE=on go build \
		-mod vendor \
		-o bin/node-problem-detector \
		-ldflags '-X k8s.io/node-problem-detector/pkg/version.version=v0.7.1-20-gc55fa33' \
		"" \
		./cmd/nodeproblemdetector
```

The `""` confuses golang into compiling the current directory :) which then causes the problem:
```
can't load package: package k8s.io/node-problem-detector: no Go files in /Users/mgrosser/Code/tools/node-problem-detector
make: *** [bin/node-problem-detector] Error 1
```

Also added a test case for the `ENABLE_JOURNAL=0` use case.

Also fixed a few rules for `log-counter` to deal with `ENABLE_JOURNAL=0`:
`log-counter` currently [only supports journald](https://github.com/kubernetes/node-problem-detector/blob/ef30a1fdd173198d1ce6475d6ed1f698759cee72/cmd/logcounter/log_counter.go#L42). So when `ENABLE_JOURNAL=0` is specified, the binary won't compile, and `docker build` won't find the binary to be installed into the docker image.

The fix basically skips the compilation and installation for `log-counter` when `ENABLE_JOURNAL=0`.
The change does not has any effect when `ENABLE_JOURNAL=1`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,12 +84,16 @@ version:
 	@echo $(VERSION)
 
 ./bin/log-counter: $(PKG_SOURCES)
+ifeq ($(ENABLE_JOURNALD), 1)
 	CGO_ENABLED=$(CGO_ENABLED) GOOS=linux GO111MODULE=on go build \
 		-mod vendor \
 		-o bin/log-counter \
 		-ldflags '-X $(PKG)/pkg/version.version=$(VERSION)' \
 		-tags "$(BUILD_TAGS)" \
 		cmd/logcounter/log_counter.go
+else
+	echo "Warning: log-counter requires journald, skipping."
+endif
 
 ./bin/node-problem-detector: $(PKG_SOURCES)
 	CGO_ENABLED=$(CGO_ENABLED) GOOS=linux GO111MODULE=on go build \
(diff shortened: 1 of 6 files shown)


grosser:
thx for looking into this!

Makefile
```diff
@@ -87,29 +84,38 @@ version:
...
+ifeq ($(ENABLE_JOURNALD), 1)
 	CGO_ENABLED=$(CGO_ENABLED) GOOS=linux GO111MODULE=on go build \
 		-mod vendor \
 		-o bin/log-counter \
 		-ldflags '-X $(PKG)/pkg/version.version=$(VERSION)' \
-		$(BUILD_TAGS) \
+		-tags "$(BUILD_TAGS)" \
 		cmd/logcounter/log_counter.go
+else
+	echo "log-counter requires journald, skipping."
```
wangzhen127: Nit: Print this as a warning:

echo "Warning: ..."

Makefile
```diff
@@ -87,29 +84,38 @@ version:
...
 		-ldflags '-X $(PKG)/pkg/version.version=$(VERSION)' \
-		$(BUILD_TAGS) \
+		-tags "$(BUILD_TAGS)" \
 		./cmd/nodeproblemdetector
 
 Dockerfile: Dockerfile.in
 	sed -e 's|@BASEIMAGE@|$(BASEIMAGE)|g' $< >$@
+ifneq ($(ENABLE_JOURNALD), 1)
+	sed -i '/Below command depends on ENABLE_JOURNAL=1/,+2d' $@
+	echo "log-counter requires journald, skipping."
```
wangzhen127: ditto

Makefile
xueweiz: Good point. Thanks!

Makefile
xueweiz: Thanks!

xueweiz:
Thanks for the review Zhen! :)

Issue #364, opened 2019-09-26
Cleanup go.sum


grosser:
there are so many duplicated dependencies in there, which 
 - makes installation take forever
 - could add bugs if multiple versions of the same library are used
 - increases the binary size

for example:

```
k8s.io/apimachinery v0.0.0-20180126010752-19e3f5aa3adc h1:+YMPfU/2f9LMz0Sy68+JodreTTgo9PNSpkLw3PakNbs=
k8s.io/apimachinery v0.0.0-20180126010752-19e3f5aa3adc/go.mod h1:ccL7Eh7zubPUSh9A3USN90/OzHNSVN6zxzde07TDCL0=
k8s.io/apimachinery v0.0.0-20181015213631-60666be32c5d/go.mod h1:ccL7Eh7zubPUSh9A3USN90/OzHNSVN6zxzde07TDCL0=
k8s.io/apimachinery v0.0.0-20190404173353-6a84e37a896d/go.mod h1:ccL7Eh7zubPUSh9A3USN90/OzHNSVN6zxzde07TDCL0=
k8s.io/apimachinery v0.0.0-20190816221834-a9f1d8a9c101 h1:QtHYUjIdgXTtJVdYQhWIQZZoXa32aF3O9BNX2up2plE=
k8s.io/apimachinery v0.0.0-20190816221834-a9f1d8a9c101/go.mod h1:ccL7Eh7zubPUSh9A3USN90/OzHNSVN6zxzde07TDCL0=
```

I'd propose some CI step that makes sure it is only using 1 version per dependency ... does that make sense ?

xueweiz:
Hi @grosser ,
From my understanding, unused entries in go.sum does not affect anything:
1. The only thing that's really affecting installation or compilation is what we stored in the /vendor directory (since we are using the "-mod vendor" option).
2. The golden rule determining the code in the /vendor directory is stored at go.mod.

I don't think it would slow down installation or increases binary size. And I don't think multiple versions in go.sum hurts anything (other than a few Bytes of storage on the repository :) ).

The closest tool I can think of to trim them, is to run `GO111MODULE=on go mod tidy`.
According to the [documentation](https://tip.golang.org/cmd/go/#hdr-Add_missing_and_remove_unused_modules), it should remove unnecessary entries in go.mod and go.sum. I ran it once against the current state, and filed #391 .

---
[...]

PR #391, opened 2019-11-26
Trim go.mod

xueweiz:
This is done via running `GO111MODULE=on go mod tidy`.

This is for #364 .


grosser:
that needs to be part of CI or it will just come back ... `GO111MODULE=on go mod tidy && git diff --exit-status`

xueweiz:
Hi @grosser , unfortunately "go mod tidy" always return 0 :)

And also I don't think it's a good idea to do this as part of CI, mainly due to the complicity of the build tags. I commented the reasoning [here](https://github.com/kubernetes/node-problem-detector/issues/364#issuecomment-558809179). Please let me know what you think under #364 . Thanks! :)

grosser:
👍

PR #392, opened 2019-11-29
Improve systemctl check, style + cleanup

arekkusu:
- Use `systemctl is-active` to check if service is running
  - Cleaner that `grep` on `systemctl status` output
  - Return success means service is running/active
  - Return failure means not running which could be due to
    stopped/failed service or that service does not exist

- Use `command -v` instead of `which`
  Ref: https://github.com/koalaman/shellcheck/wiki/SC2230

- Follow Google "Shell Style Guide": indent, use "readonly"

- Minor: Rephrase comment, avoid all caps

### DIFF
--- a/config/plugin/check_ntp.sh
+++ b/config/plugin/check_ntp.sh
@@ -1,23 +1,27 @@
 #!/bin/bash
 
-# NOTE: THIS NTP SERVICE CHECK SCRIPT ASSUME THAT NTP SERVICE IS RUNNING UNDER SYSTEMD.
-#       THIS IS JUST AN EXAMPLE. YOU CAN WRITE YOUR OWN NODE PROBLEM PLUGIN ON DEMAND.
+# This plugin checks if the ntp service is running under systemd.
+# NOTE: This is only an example for systemd services.
 
-OK=0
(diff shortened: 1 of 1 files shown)


arekkusu:
Follow up to https://github.com/kubernetes/node-problem-detector/pull/369 - update second plugin

Mostly some cleanup (see commit notes). Note this does not differences between non-running service and non-existent one. Could probably add more logic to cover those and output more precise message but probably not essential. Also I think this is meant as simple example.

I am wondering if change to `custom-plugin-monitor.json` is needed:

https://github.com/kubernetes/node-problem-detector/blob/8704ec0c42b27db5eca8e57c3dbfc172ba82765d/config/custom-plugin-monitor.json#L16

xueweiz:
Thanks for helping improving the code health! :)
/ok-to-test
/lgtm
/approve

xueweiz:
> I am wondering if change to custom-plugin-monitor.json is needed:

I think it's OK to leave it as is? I agree "NTP" should probably in capital case. I'll defer that to @andyxning to decide :) He knows more about this. I'm fine with either way.

PR #393, opened 2019-11-30
fix make test early failure

jiayuc:
### Motivation:
`make test` would fail right now because `TestMain()` in `test/e2e/metriconly/e2e_npd_test.go` is attempting to create gcloud compute service. See the error reported below by running `make test`:
image

### Changes:
This check should be moved into `TestNPD(t *testing.T)`, where this test can be skipped according to `make test rule`:

```makefile
 test: vet fmt
	GO111MODULE=on go test -mod vendor -timeout=1m -v -race -short -tags "$(BUILD_TAGS)" ./...
```

### Test Plan:
- `make test` should pass

### DIFF
--- a/test/e2e/metriconly/e2e_npd_test.go
+++ b/test/e2e/metriconly/e2e_npd_test.go
@@ -59,6 +59,12 @@ func TestNPD(t *testing.T) {
 		t.Skip("skipping test in short mode.")
 	}
 
+	var err error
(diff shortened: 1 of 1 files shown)


jiayuc:
I signed it.

xueweiz:
Thanks for filing this PR @jiayuc !

PR #395, opened 2019-12-05
Using time.Since(t) instead of t.Sub(time.Now())

yuzhiquan:
Using time.Since(t) instead of t.Sub(time.Now())


yuzhiquan:
@xueweiz hi, xueweiz, i had send a email for newrequest, if you have receive email, may be you could reply it, thanks！

PR #401, opened 2019-12-07
Rename problem metrics

### LINKED ISSUE #326: Prometheus metrics naming
Maybe you can follow the prometheus naming metrics and label style-guide for the exported metrics?
https://prometheus.io/docs/practices/naming/
https://github.com/kubernetes/node-problem-detector/blob/e280e2075ada86d2bde947bb2ff0776575026734/pkg/problemmetrics/problem_metrics.go#L41-L42

xueweiz:
This PR implements the proposal at #326 .
/fix #326

Renamed below two metrics:
`problem_counter` -> `problem_count_total`
`problem_gauge` -> `problem_state`

This is a user-visible breaking change due the metric renaming:
1. Anyone depending on Prometheus metrics will need to change their dashboard/alerting to the new metric name.
2. Anyone depending on the reported Kubernetes events/conditions will NOT be affected.
3. Anyone depending on the Stackdriver Monitoring metrics will NOT be affected.

### DIFF
--- a/pkg/problemmetrics/problem_metrics_test.go
+++ b/pkg/problemmetrics/problem_metrics_test.go
@@ -43,7 +43,7 @@ func TestNewProblem(t *testing.T) {
 			counts:  []int64{1},
 			expectedMetrics: []metrics.Int64MetricRepresentation{
 				{
-					Name:   "problem_counter",
+					Name:   "problem_count_total",
 					Labels: map[string]string{"reason": "foo"},
 					Value:  1,
 				},
@@ -55,7 +55,7 @@ func TestNewProblem(t *testing.T) {
 			counts:  []int64{1, 1},
 			expectedMetrics: []metrics.Int64MetricRepresentation{
 				{
-					Name:   "problem_counter",
+					Name:   "problem_count_total",
 					Labels: map[string]string{"reason": "foo"},
 					Value:  2,
 				},
@@ -67,12 +67,12 @@ func TestNewProblem(t *testing.T) {
 			counts:  []int64{1, 1, 1},
 			expectedMetrics: []metrics.Int64MetricRepresentation{
 				{
-					Name:   "problem_counter",
+					Name:   "problem_count_total",
 					Labels: map[string]string{"reason": "foo"},
 					Value:  2,
 				},
 				{
-					Name:   "problem_counter",
+					Name:   "problem_count_total",
 					Labels: map[string]string{"reason": "bar"},
 					Value:  1,
 				},
@@ -84,12 +84,12 @@ func TestNewProblem(t *testing.T) {
 			counts:  []int64{0, 0},
 			expectedMetrics: []metrics.Int64MetricRepresentation{
 				{
-					Name:   "problem_counter",
+					Name:   "problem_count_total",
 					Labels: map[string]string{"reason": "foo"},
 					Value:  0,
 				},
 				{
-					Name:   "problem_counter",
+					Name:   "problem_count_total",
 					Labels: map[string]string{"reason": "bar"},
 					Value:  0,
 				},
@@ -101,12 +101,12 @@ func TestNewProblem(t *testing.T) {
 			counts:  []int64{0, 0, 1, 1, 1},
 			expectedMetrics: []metrics.Int64MetricRepresentation{
 				{
-					Name:   "problem_counter",
+					Name:   "problem_count_total",
 					Labels: map[string]string{"reason": "foo"},
 					Value:  2,
 				},
 				{
-					Name:   "problem_counter",
+					Name:   "problem_count_total",
 					Labels: map[string]string{"reason": "bar"},
 					Value:  1,
 				},
@@ -152,7 +152,7 @@ func TestSetProblemGauge(t *testing.T) {
 			},
 			expectedMetrics: []metrics.Int64MetricRepresentation{
 				{
-					Name:   "problem_gauge",
+					Name:   "problem_state",
 					Labels: map[string]string{"type": "ProblemTypeA", "reason": "ReasonFoo"},
 					Value:  1,
 				},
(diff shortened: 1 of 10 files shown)


wangzhen127:
/assign @Random-Liu 
What's the policy of making user facing breaking changes in NPD?

pkg/problemmetrics/problem_metrics_test.go
```diff
@@ -43,7 +43,7 @@ func TestNewProblem(t *testing.T) {
 			counts:  []int64{1},
 			expectedMetrics: []metrics.Int64MetricRepresentation{
 				{
-					Name:   "problem_counter",
+					Name:   "problem_count_total",
```
yeya24: count and total may seem quite similar? What about problem_occurrence_total?

pkg/systemlogmonitor/README.md
```diff
@@ -96,20 +96,20 @@ default.
...
-problem_counter{reason="DockerHung"} 1
-# HELP problem_gauge Whether a specific type of problem is affecting the node or not.
-# TYPE problem_gauge gauge
-problem_gauge{condition="KernelDeadlock",reason="DockerHung"} 1
+# HELP problem_count_total Number of times a specific type of problem have occurred.
+# TYPE problem_count_total counter
+problem_count_total{reason="DockerHung"} 1
+# HELP problem_state Whether a specific type of problem is affecting the node or not.
+# TYPE problem_state gauge
+problem_state{condition="KernelDeadlock",reason="DockerHung"} 1
```
yeya24: Why there is a label `condition`? Should it be `type`?

yeya24:
@xueweiz  Could I ask that why choosing open census, not Prometheus or opentelemetry?

pkg/problemmetrics/problem_metrics_test.go
xueweiz: I have no preference about this :) `problem_total` or `problem_count_total` or `problem_occurrence_total` all looks fine to me :)

@HaveFun83 proposed `problem_count_total` [here](https://github.com/kubernetes/node-problem-detector/issues/326#issuecomment-521135338).

pkg/systemlogmonitor/README.md
xueweiz: Good catch! Yes this is a typo. Fixed :)

pkg/problemmetrics/problem_metrics_test.go
yeya24: That's OK. Let's choose problem_count_total since problem_total is kind of ambiguous.

yeya24:
> > why choosing open census, not Prometheus or opentelemetry?
> 
> We chose OpenCensus when we [proposed](https://github.com/kubernetes/node-problem-detector#links) the Metrics Mode for NPD, see the [Metrics Mode Proposal](https://docs.google.com/document/d/1SeaUz6kBavI283Dq8GBpoEUDrHA2a795xtw0OvjM568/edit#heading=h.xw2rl6ej7zhr) for details.
> 
> We didn't choose Prometheus because Prometheus does not have a flexible exporting ability. So if you want to export the metrics to an external monitoring system (e.g. Stackdriver Monitoring, CloudWatch), Prometheus library won't help you.
> 
> OpenTelemetry would be a good alternative. However, it is still in [pre-release](https://opentelemetry.io/project-status/) status right now. We should not rely on an unstable technology, and we cannot wait for OpenTelemetry to become production ready. So we cannot use it. When it becomes production ready, we could consider switching to it (we'd like to keep the change invisible to the NPD users, hopefully).

@xueweiz  Thanks for the detailed answer. Learned a lot!

blakebarnett:
Would it be possible to also add a label that includes the node name? This would help in correlating problems a lot quicker without needing to create a custom scrape config for prometheus, etc.

xueweiz:
Hi @balajismaniam ,

> Would it be possible to also add a label that includes the node name?

Technically, this is pretty easy to do. However, I don't think this follows the best Prometheus monitoring pattern. Specifically, I'm referring to the ["jobs and instances" Prometheus concepts](https://prometheus.io/docs/concepts/jobs_instances/).

These labels (e.g. node name) specifies the "monitored resource". i.e. what is the machine / pod / container that is being monitored. And eventually, these ID must be attached to the metrics to be useful for the end users (SREs, oncalls). The question is, which IDs should be populated in the raw metrics, and which IDs should be attached by Prometheus during scrapping?

In my opinion, any label that is part of the scrapped target ID (i.e. machine ID in NPD's case) should be automatically attached by Prometheus; and any label that is not part of the scrapped target ID (i.e. container ID, pod ID, NPD does not have per pod/container scrapped target) should be populated in the raw metrics.

This fully leverages the flexibility on Prometheus, and allows people to specify the labels in their preferred format (e.g. node name? IP? port?). This might be especially useful for people running 2 NPDs on different ports (yes this is a pretty strange use case :) and if you'd like, I can explain why would someone do that). WDYT?

---

Also, there are some similar use cases. For example, [cadvisor metrics](https://github.com/google/cadvisor/blob/master/docs/storage/prometheus.md). You can [see](https://github.com/google/cadvisor/blob/c470c610615c5662cce68d19a10b53435f153e28/metrics/testdata/prometheus_metrics#L151) that only container ID, image name, etc gets reported in the raw metrics; all node level ID are not reported, and relies on Prometheus to populate them during scrapping.

blakebarnett:
I think in this case the `instance` is the node (because it's running as a DaemonSet and is privileged), so there's no harm in adding it as that is usually considered "best practice" anyway. 

I agree about more specific details being inappropriate, for example I would _LOVE_ to be able to correlate OOMKills with specific pods or containers by mapping that with the cgroup, but adding labels for that would result in a huge cardinality problem, so I'm trying to find a better solution for that on top of these metrics.

blakebarnett:
Nevermind, I just realized that I was assuming some of our metrics had this but really it's just one of our generic scrape configs. :)

PR #402, opened 2019-12-11
Ignore first collected disk stats to prevent metric distortion

gmemcc:
Disk collector depends on the last collected metrics,  which are defaulted to zero. This can cause  distortion on first time collection.

### DIFF
--- a/pkg/systemstatsmonitor/disk_collector.go
+++ b/pkg/systemstatsmonitor/disk_collector.go
@@ -115,15 +115,6 @@ func (dc *diskCollector) collect() {
 		dc.historyIOTime[deviceName] = ioCountersStat.IoTime
 		dc.historyWeightedIO[deviceName] = ioCountersStat.WeightedIO
 
-		if !historyExist {
-			// Ignore first collected stats.
-			return
-		}
-		avgQueueLen := float64(0.0)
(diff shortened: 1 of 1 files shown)


xueweiz:
Hi @gmemcc , thanks for proposing this!

Just to confirm, which metric(s) are initialized to 0 and causing the distortion on first time collection? Could you share a Prometheus metrics output (`curl localhost:20257/metrics`) or a screenshot with me? Thanks!

(From my understanding, only the queue length should have this problem. But I want to confirm.

/ok-to-test

gmemcc:
@xueweiz Thanks for your review and comments, you are right, it's my misunderstanding.  only disk_avg_queue_len  is distorted because its metric type is 'gauge'

here's my metrics report with the config file  `config/system-stats-monitor.json`, sdb is idle during the test:

on startup:
```
disk_avg_queue_len{device_name="sdb"} 69.453342040168
disk_io_time{device_name="sdb"} 4.1983876e+07
disk_weighted_io{device_name="sdb"} 2.9159205e+09
```

1 minutes later:
```
disk_avg_queue_len{device_name="sdb"} 0
disk_io_time{device_name="sdb"} 4.1983876e+07
disk_weighted_io{device_name="sdb"} 2.9159205e+09
```

It doesn't matter although the io time is very large.

PR #404, opened 2020-01-03
Fix disk metrics unit and queue_length calculation

xueweiz:
I just realized that the unit for the disk metrics is wrong. The metrics are in milliseconds, not seconds. Since Prometheus does not print units under query, I didn't catch this sooner. And for the same reason, this should not affect any NPD users using Prometheus exporter.

Another problem is that the way I calculated queue_length was actually wrong. The queue_length field is supposed to match the `aqu-sz` field (FKA `avgqu-sz`) provided by [iostat](https://manpages.debian.org/testing/sysstat/iostat.1.en.html):
> The average queue length of the requests that were issued to the device

This is how `iostat` [calculates](https://github.com/sysstat/sysstat/blob/e011d97f0a0c7c42a687d02386ad60853ee7dc42/iostat.c#L1095) `aqu-sz`: `S_VALUE(ioj->rq_ticks, ioi->rq_ticks, itv) / 1000.0)`
1. `ioj` and `ioi` are [two sample statistics](https://github.com/sysstat/sysstat/blob/e011d97f0a0c7c42a687d02386ad60853ee7dc42/iostat.c#L957) `iostat` collected from `/proc/diskstats`;
1. `rq_ticks` field is the [14th field](https://github.com/sysstat/sysstat/blob/e011d97f0a0c7c42a687d02386ad60853ee7dc42/iostat.c#L627) collected from a line in `/proc/diskstats`, which [means](https://www.kernel.org/doc/Documentation/iostats.txt) "weighted # of milliseconds spent doing I/Os";
1. `itv` [means](https://github.com/sysstat/sysstat/blob/e011d97f0a0c7c42a687d02386ad60853ee7dc42/iostat.c#L951) "Interval of time".
1. The `/ 1000.0` in the end is to translate between `s` and `ms`.

So basically we want to get the delta of weighted IO time, and divide it by the time elapsed.

In NPD I used to implement it as the delta of weighted IO time divided by the delta of IO time. That was wrong. This PR fixes it.

### DIFF
--- a/pkg/systemstatsmonitor/disk_collector.go
+++ b/pkg/systemstatsmonitor/disk_collector.go
@@ -40,6 +40,7 @@ type diskCollector struct {
 
 	historyIOTime     map[string]uint64
 	historyWeightedIO map[string]uint64
+	lastSampleTime    time.Time
 }
 
 func NewDiskCollectorOrDie(diskConfig *ssmtypes.DiskStatsConfig) *diskCollector {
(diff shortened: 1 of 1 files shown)


xueweiz:
/assign @wangzhen127 
Zhen, could you help review this bug fix? Thanks!

PR #418, opened 2020-03-06
Add namespace option for events

muff1nman:
By default node-problem-detector writes events to the default namespace. This PR adds an option to allow events to be written to a different namespace.

### DIFF
--- a/cmd/options/options.go
+++ b/cmd/options/options.go
@@ -49,6 +49,8 @@ type NodeProblemDetectorOptions struct {
 	// k8sExporter options
 	// EnableK8sExporter is the flag determining whether to report to Kubernetes.
 	EnableK8sExporter bool
+	// EventNamespace is the namespace events are written to
+	EventNamespace string
 	// ApiServerOverride is the custom URI used to connect to Kubernetes ApiServer.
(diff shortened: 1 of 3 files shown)


xueweiz:
Looking at the unit test, seems there is some problem:

```
GO111MODULE=on go list -mod vendor -tags " journald" ./... | \
	grep -v "./vendor/*" | \
	GO111MODULE=on xargs go vet -mod vendor -tags " journald"
# k8s.io/node-problem-detector/pkg/exporters/k8sexporter/problemclient
vet: pkg/exporters/k8sexporter/problemclient/problem_client_test.go:45:33: too few arguments in call to getNodeRef
```

Seems the refactoring is breaking some other code paths. Could you please fix them? Thanks! You can run `make test` to verify whether the problem has been fixed.

xueweiz:
/lgtm
/approve

@muff1nman the patch looks good to me. However, it will only be merged after you signed the Linux Foundation CLA. [Here](https://github.com/kubernetes/node-problem-detector/pull/418#issuecomment-595521407) are the necessary steps.

muff1nman:
I signed it.

PR #421, opened 2020-03-25
Install util-linux to have lsblk binary

### LINKED ISSUE #420: official image does not contain lsblk binary
we run npd 0.8.0 from: k8s.gcr.io/node-problem-detector:v0.8.0

but every some seconds we have the following error log entries:

```
node-problem-detector-q99zr node-problem-detector E0325 10:30:02.459512       1 disk_collector.go:145] Error calling lsblk
```

Checking inside the container confirms this.
I am happy to raise a PR to fix this.

majst01:
This addresses #420

### DIFF
--- a/Dockerfile.in
+++ b/Dockerfile.in
@@ -15,7 +15,7 @@
 FROM @BASEIMAGE@
 MAINTAINER Random Liu <lantaol@google.com>
 
-RUN clean-install libsystemd0 bash
+RUN clean-install util-linux libsystemd0 bash
 
 # Avoid symlink of /etc/localtime.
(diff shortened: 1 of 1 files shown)


majst01:
@xueweiz any estimates when there is a new release of NPD ?

xueweiz:
@Random-Liu Do we have plan to release NPD any time soon?

---
By the way, if the release is blocking you from picking up the fix, you could consider build a docker image on top of the released NPD image, and install lsblk in it. Please let me know if this still cannot unblock you.

majst01:
Yes of course i will do if a release of NPD will take several weeks, but if its in the foreseeable future i will wait.

xueweiz:
I guess we might wait for a while before doing the next NPD release, because there are not many major features/fixes recently.

PR #422, opened 2020-04-08
update system-log-monitor and image version

blackwith:
Update system-log-monitor which is deprecated to --config.system-log-monitor and update image to 0.8.1


blackwith:
@Random-Liu @dchen1107 possible to get an ok to test?

blackwith:
@Random-Liu @andyxning @xueweiz is it possible to get an /ok-to-test ?

PR #426, opened 2020-05-11
Add health-check-monitor

abansal4032:
The PR contains the [health-monitor](https://github.com/kubernetes/kubernetes/blob/master/cluster/gce/gci/health-monitor.sh) implementation as a NPD monitor.
This is implemented as a new binary : health-checker, which can be enabled using config/container-runtime-health-checker.json and config/kubelet-health-checker.json to check and repair container runtime and kubelet services respectively.

### DIFF
--- a/cmd/healthchecker/options/options.go
+++ b/cmd/healthchecker/options/options.go
@@ -1,0 +1,102 @@
+/*
+Copyright 2020 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package options
+
+import (
+	"flag"
+	"fmt"
+	"time"
+
+	"github.com/spf13/pflag"
+
+	"k8s.io/node-problem-detector/pkg/healthchecker/types"
+)
+
+// NewHealthCheckerOptions returns an empty health check options struct.
+func NewHealthCheckerOptions() *HealthCheckerOptions {
+	return &HealthCheckerOptions{}
+}
+
+// HealthCheckerOptions are the options used to configure the health checker.
+type HealthCheckerOptions struct {
+	Component          string
+	SystemdService     string
+	EnableRepair       bool
+	CriCtlPath         string
+	CriSocketPath      string
+	CoolDownTime       time.Duration
+	HealthCheckTimeout time.Duration
+}
+
+// AddFlags adds health checker command line options to pflag.
+func (hco *HealthCheckerOptions) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&hco.Component, "component", types.KubeletComponent,
+		"The component to check health for. Supports kubelet, docker and cri")
+	fs.StringVar(&hco.SystemdService, "systemd-service", "",
+		"The underlying systemd service responsible for the component. Set to the corresponding component for docker and kubelet, containerd for cri.")
+	fs.BoolVar(&hco.EnableRepair, "enable-repair", true, "Flag to enable/disable repair attempt for the component.")
+	fs.StringVar(&hco.CriCtlPath, "crictl-path", types.DefaultCriCtl,
+		"The path to the crictl binary. This is used to check health of cri component.")
+	fs.StringVar(&hco.CriSocketPath, "cri-socket-path", types.DefaultCriSocketPath,
+		"The path to the cri socket. Used with crictl to specify the socket path.")
+	fs.DurationVar(&hco.CoolDownTime, "cooldown-time", types.DefaultCoolDownTime,
+		"The duration to wait for the service to be up before attempting repair.")
+	fs.DurationVar(&hco.HealthCheckTimeout, "health-check-timeout", types.DefaultHealthCheckTimeout,
+		"The time to wait before marking the component as unhealthy.")
+}
+
+// IsValid validates health checker command line options.
+// Returns error if invalid, nil otherwise.
+func (hco *HealthCheckerOptions) IsValid() error {
+	// Make sure the component specified is valid.
+	if hco.Component != types.KubeletComponent && hco.Component != types.DockerComponent && hco.Component != types.CRIComponent {
+		return fmt.Errorf("the component specified is not supported. Supported components are : <kubelet/docker/cri>")
+	}
+	// Make sure the systemd service is specified if repair is enabled.
+	if hco.EnableRepair && hco.SystemdService == "" {
+		return fmt.Errorf("systemd-service cannot be empty when repair is enabled")
+	}
+	// Skip checking further if the component is not cri.
+	if hco.Component != types.CRIComponent {
+		return nil
+	}
+	// Make sure the crictl path is not empty for cri component.
+	if hco.Component == types.CRIComponent && hco.CriCtlPath == "" {
+		return fmt.Errorf("the crictl-path cannot be empty for cri component")
+	}
+	// Make sure the cri socker path is not empty for cri component.
+	if hco.Component == types.CRIComponent && hco.CriSocketPath == "" {
+		return fmt.Errorf("the cri-socket-path cannot be empty for cri component")
+	}
+	return nil
+}
+
+// SetDefaults sets the defaults values for the dependent flags.
+func (hco *HealthCheckerOptions) SetDefaults() {
+	if hco.SystemdService != "" {
+		return
+	}
+	if hco.Component != types.CRIComponent {
+		hco.SystemdService = hco.Component
+		return
+	}
+	hco.SystemdService = types.ContainerdService
+}
+
+func init() {
+	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
+}
--- a/cmd/healthchecker/options/options_test.go
+++ b/cmd/healthchecker/options/options_test.go
@@ -1,0 +1,76 @@
+/*
+Copyright 2020 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package options
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"k8s.io/node-problem-detector/pkg/healthchecker/types"
+)
+
+func TestIsValid(t *testing.T) {
+	testCases := []struct {
+		name        string
+		hco         HealthCheckerOptions
+		expectError bool
+	}{
+		{
+			name: "valid component",
+			hco: HealthCheckerOptions{
+				Component: types.KubeletComponent,
+			},
+			expectError: false,
+		},
+		{
+			name: "invalid component",
+			hco: HealthCheckerOptions{
+				Component: "wrongComponent",
+			},
+			expectError: true,
+		},
+		{
+			name: "empty crictl-path with cri",
+			hco: HealthCheckerOptions{
+				Component:    types.CRIComponent,
+				CriCtlPath:   "",
+				EnableRepair: false,
+			},
+			expectError: true,
+		},
+		{
+			name: "empty systemd-service and repair enabled",
+			hco: HealthCheckerOptions{
+				Component:      types.KubeletComponent,
+				EnableRepair:   true,
+				SystemdService: "",
+			},
+			expectError: true,
+		},
+	}
+
+	for _, test := range testCases {
+		t.Run(test.name, func(t *testing.T) {
+			if test.expectError {
+				assert.Error(t, test.hco.IsValid(), "HealthChecker option %+v is invalid. Expected IsValid to return error.", test.hco)
+			} else {
+				assert.NoError(t, test.hco.IsValid(), "HealthChecker option %+v is valid. Expected IsValid to return nil.", test.hco)
+			}
+		})
+	}
+}
--- a/config/health-checker-kubelet.json
+++ b/config/health-checker-kubelet.json
@@ -1,0 +1,33 @@
+{
+  "plugin": "custom",
+  "pluginConfig": {
+    "invoke_interval": "10s",
+    "timeout": "3m",
+    "max_output_length": 80,
+    "concurrency": 1
+  },
+  "source": "health-checker",
+  "metricsReporting": true,
+  "conditions": [
+    {
+      "type": "KubeletUnhealthy",
+      "reason": "KubeletIsHealthy",
+      "message": "kubelet on the node is functioning properly"
+    }
+  ],
+  "rules": [
+    {
+      "type": "permanent",
+      "condition": "KubeletUnhealthy",
+      "reason": "KubeletUnhealthy",
+      "path": "/home/kubernetes/bin/health-checker",
+      "args": [
+        "--component=kubelet",
+        "--enable-repair=false",
+        "--cooldown-time=1m",
+        "--health-check-timeout=10s"
+      ],
+      "timeout": "3m"
+    }
+  ]
+}
--- a/pkg/healthchecker/health_checker.go
+++ b/pkg/healthchecker/health_checker.go
@@ -1,0 +1,163 @@
+/*
+Copyright 2020 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package healthchecker
+
+import (
+	"context"
+	"errors"
+	"net/http"
+	"os/exec"
+	"strings"
+	"time"
+
+	"github.com/golang/glog"
+
+	"k8s.io/node-problem-detector/cmd/healthchecker/options"
+	"k8s.io/node-problem-detector/pkg/healthchecker/types"
+)
+
+type healthChecker struct {
+	enableRepair    bool
+	healthCheckFunc func() bool
+	// The repair is "best-effort" and ignores the error from the underlying actions.
+	// The bash commands to kill the process will fail if the service is down and hence ignore.
+	repairFunc         func()
+	uptimeFunc         func() (time.Duration, error)
+	crictlPath         string
+	healthCheckTimeout time.Duration
+	coolDownTime       time.Duration
+}
+
+// NewHealthChecker returns a new health checker configured with the given options.
+func NewHealthChecker(hco *options.HealthCheckerOptions) (types.HealthChecker, error) {
+	hc := &healthChecker{
+		enableRepair:       hco.EnableRepair,
+		crictlPath:         hco.CriCtlPath,
+		healthCheckTimeout: hco.HealthCheckTimeout,
+		coolDownTime:       hco.CoolDownTime,
+	}
+	hc.healthCheckFunc = getHealthCheckFunc(hco)
+	hc.repairFunc = getRepairFunc(hco)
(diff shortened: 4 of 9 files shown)


config/container-runtime-health-checker.json
```diff
@@ -0,0 +1,32 @@
+{
```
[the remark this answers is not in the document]
abansal4032: Done.

config/container-runtime-health-checker.json
```diff
@@ -0,0 +1,32 @@
...
+  "plugin": "custom",
+  "pluginConfig": {
+    "invoke_interval": "5m",
```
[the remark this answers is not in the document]
abansal4032: Yes. Updated to 10s.

config/container-runtime-health-checker.json
```diff
@@ -0,0 +1,32 @@
...
+  "conditions": [
+    {
+      "type": "ContainerRuntimeUnhealthy",
+      "reason": "ContainerRuntimeIsHealthy",
+      "message": "Container runtime on the node is functioning properly"
+    }
+  ],
+  "rules": [
+    {
+      "type": "temporary",
```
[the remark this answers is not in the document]
abansal4032: Changed this to add a permanent condition here.

pkg/healthchecker/health_checker.go
```diff
@@ -0,0 +1,143 @@
...
+// Returns true if healthy, false otherwise.
+func (hc *healthChecker) CheckHealth() bool {
+	// Poll till the timeout for the component to be up.
+	if wait.PollImmediate(types.PollInterval, hc.timeout, func() (bool, error) {
+		healthy := hc.healthCheckFunc()
+		if healthy {
+			return true, nil
+		}
+		return false, nil
+	}) != nil {
```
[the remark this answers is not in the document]
abansal4032: Done.

pkg/healthchecker/health_checker.go
```diff
@@ -0,0 +1,143 @@
...
+		return nil, fmt.Errorf("invalid component : %+v specified in health check options", hco.Component)
+	}
+	return hc, nil
+}
+
+// CheckHealth checks for the health of the component and tries to repair if enabled.
+// Returns true if healthy, false otherwise.
+func (hc *healthChecker) CheckHealth() bool {
+	// Poll till the timeout for the component to be up.
+	if wait.PollImmediate(types.PollInterval, hc.timeout, func() (bool, error) {
```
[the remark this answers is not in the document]
abansal4032: 1) Initial wait time : NPD starts only after API server is up and hence the first invocation of the plugin guarantees kubelet and docker to be up.
2) Cool down time : Changed wait-time to cooldown-time to wait after repair is attempted.
3) Health check timeout : Added a new flag defaulting to 10s.

pkg/healthchecker/health_checker.go
```diff
@@ -0,0 +1,147 @@
...
+		return nil, fmt.Errorf("invalid component : %+v specified in health check options", hco.Component)
+	}
+	return hc, nil
+}
+
+// CheckHealth checks for the health of the component and tries to repair if enabled.
+// Returns true if healthy, false otherwise.
+func (hc *healthChecker) CheckHealth() bool {
+	// Poll till the health check timeout for the component to be up.
+	if err := wait.PollImmediate(types.PollInterval, hc.healthCheckTimeout, func() (bool, error) {
```
wangzhen127, line 68: I think we just need the timeout part here, and do not need the retry, because the health check func is blocking IIRC. Is there a dedicated timeout util function? If we want to just utilize `PollImmediate` for the timeout logic, we can set the interval to be same as timeout value.

pkg/healthchecker/health_checker.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+		return false
+	}
+	return true
+}
+
+// kubeletHealthCheck checks the health of kubelet by hitting to the health endpoint.
+// Returns true if healthy, false otherwise.
+func kubeletHealthCheck() bool {
+	httpClient := http.Client{Timeout: types.HttpTimeout}
```
wangzhen127, line 91: s/HttpTimeout/KubeletTimeout

pkg/healthchecker/types/types.go
```diff
@@ -0,0 +1,36 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package types
+
+import "time"
+
+const (
+	HttpTimeout                = 10 * time.Second
```
wangzhen127, line 22: Those are consts, not types?

pkg/healthchecker/health_checker.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+	}
+	return func() {
+		execCommand("systemctl", "kill", "--kill-who=main", types.ContainerdRuntime)
+	}
+}
+
+// execCommand executes the bash command and returns the error.
+func execCommand(command string, args ...string) error {
+	cmd := exec.Command(command, args...)
```
wangzhen127, line 140: Is sudo needed?

cmd/healthchecker/options/options.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	// Skip checking for container-runtime if the component is not container-runtime
+	if hco.Component != types.ContainerRuntimeComponent {
+		return
+	}
+	// Make sure the container-runtime is valid.
+	if hco.ContainerRuntime != types.DockerRuntime && hco.ContainerRuntime != types.ContainerdRuntime {
+		panic("The container-runtime specified is not supported. Supported runtimes are : <docker/containerd>")
+	}
+}
```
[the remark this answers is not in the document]
abansal4032: The path has a default value which is used in case it is empty.

config/health-checker-container-runtime.json
```diff
@@ -0,0 +1,33 @@
...
+      "type": "permanent",
+      "condition": "ContainerRuntimeUnhealthy",
+      "reason": "DockerUnhealthy",
+      "path": "/home/kubernetes/bin/health-checker",
+      "args": [
+        "--component=container-runtime",
+        "--enable-repair=false",
+        "--container-runtime=docker",
+        "--cooldown-time=2m"
+      ],
```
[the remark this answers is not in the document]
abansal4032: Done.

config/health-checker-kubelet.json
```diff
@@ -0,0 +1,32 @@
...
+    {
+      "type": "permanent",
+      "condition": "KubeletUnhealthy",
+      "reason": "KubeletUnhealthy",
+      "path": "/home/kubernetes/bin/health-checker",
+      "args": [
+        "--component=kubelet",
+        "--enable-repair=false",
+        "--cooldown-time=2m"
+      ],
```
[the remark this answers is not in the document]
abansal4032: Done.

cmd/healthchecker/options/options.go
```diff
@@ -0,0 +1,76 @@
...
+		"The component to check health for. Supports kubelet and container-runtime. Defaults to kubelet.")
+	fs.StringVar(&hco.ContainerRuntime, "container-runtime", types.DockerRuntime,
+		"The underlying container runtime name on the node. This is ignored if component is not container-runtime. Supports docker and containerd. Defaults to docker.")
+	fs.BoolVar(&hco.EnableRepair, "enable-repair", true, "Flag to enable/disable repair attempt for the component. Defaults to true.")
+	fs.StringVar(&hco.CriCtlPath, "crictl-path", types.DefaultCriCtl,
+		"The path to the crictl binary. This is used to check health of container-runtimes other than docker. Defaults to /usr/bin/crictl")
+	fs.DurationVar(&hco.CoolDownTime, "cooldown-time", types.DefaultCooldownTime,
+		"The time to wait after repairing the component. This helps in maintaining a buffer between repair and next invocation. Defaults to 2m.")
+	fs.DurationVar(&hco.HealthCheckTimeout, "health-check-timeout", types.DefaultHealthCheckTimeout,
+		"The time to wait before marking the component as unhealthy. Defaults to 10s.")
```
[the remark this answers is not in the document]
abansal4032: fs.PrintDefaults() prints the defaults with usages. Removed the hardcodings here.

config/health-checker-kubelet.json
```diff
@@ -0,0 +1,32 @@
...
+  "rules": [
+    {
+      "type": "permanent",
+      "condition": "KubeletUnhealthy",
+      "reason": "KubeletUnhealthy",
+      "path": "/home/kubernetes/bin/health-checker",
+      "args": [
+        "--component=kubelet",
+        "--enable-repair=false",
+        "--cooldown-time=2m"
```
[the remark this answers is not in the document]
abansal4032: Done.

pkg/healthchecker/health_checker.go
abansal4032: Did not see a dedicated timeout function. Using the PollImmediate with timeout = pollinterval.

pkg/healthchecker/health_checker.go
abansal4032: Now that we are using PollImmediate and assuming the healthcheck function is blocking, we can make this call without timeout and let the caller handle the same.
Removing the timeout here.

pkg/healthchecker/types/types.go
abansal4032: Yeah these are just consts. Since the usage is across packages, this felt like a correct place to have them. Similar pattern is followed here : https://github.com/kubernetes/node-problem-detector/blob/master/pkg/custompluginmonitor/types/types.go#L26

pkg/healthchecker/health_checker.go
abansal4032: I don't think so. It is running fine and able to execute commands on the test cluster. Maybe because NPD itself is running as sudo on the node?
Similar usages here : https://github.com/kubernetes/node-problem-detector/blob/master/pkg/systemstatsmonitor/disk_collector.go#L293

cmd/healthchecker/options/options.go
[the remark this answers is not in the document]
abansal4032: No, it will use the empty string. I have added a test to detect that. But in general, if that flag is set there is no way of making sure that path exists in the test.

wangzhen127:
/lgtm
/hold

Check with @Random-Liu to see if he wants to take a look (I remember he said he would take a look this week?). Feel free to remove hold if he does not need to review.

cmd/healthchecker/options/options.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+// NewHealthCheckerOptions returns an empty health check options struct.
+func NewHealthCheckerOptions() *HealthCheckerOptions {
+	return &HealthCheckerOptions{}
+}
+
+// HealthCheckerOptions are the options used to configure the health checker.
+type HealthCheckerOptions struct {
+	Component          string
```
Random-Liu, line 35: Can we define a type here `type ComponentType string`?

cmd/healthchecker/options/options.go
```diff
@@ -0,0 +1,80 @@
...
+		"The time to wait after repairing the component. This helps in maintaining a buffer between repair and next invocation.")
+	fs.DurationVar(&hco.HealthCheckTimeout, "health-check-timeout", types.DefaultHealthCheckTimeout,
+		"The time to wait before marking the component as unhealthy.")
+}
+
+// ValidOrDie validates health checker command line options.
+func (hco *HealthCheckerOptions) ValidOrDie() {
+	// Make sure the component specified is valid.
+	if hco.Component != types.KubeletComponent && hco.Component != types.ContainerRuntimeComponent {
+		panic("The component specified is not supported. Supported components are : <docker/container-runtime>")
```
Random-Liu, line 62: Supported components don't have `kubelet`.

Random-Liu, line 62: Should we return error instead of panic? So that we can return an Unknown status outside.

cmd/healthchecker/options/options.go
```diff
@@ -0,0 +1,80 @@
...
+func NewHealthCheckerOptions() *HealthCheckerOptions {
+	return &HealthCheckerOptions{}
+}
+
+// HealthCheckerOptions are the options used to configure the health checker.
+type HealthCheckerOptions struct {
+	Component          string
+	ContainerRuntime   string
+	EnableRepair       bool
+	CriCtlPath         string
```
Random-Liu, line 38: How is the CRI socket path defined?

cmd/healthchecker/options/options.go
```diff
@@ -0,0 +1,80 @@
...
+
+// NewHealthCheckerOptions returns an empty health check options struct.
+func NewHealthCheckerOptions() *HealthCheckerOptions {
+	return &HealthCheckerOptions{}
+}
+
+// HealthCheckerOptions are the options used to configure the health checker.
+type HealthCheckerOptions struct {
+	Component          string
+	ContainerRuntime   string
```
Random-Liu, line 36: I hope we can make the systemd service name configurable, so that this can be used to support cri-o as well.

Probably:
* Component type: kubelet, docker, cri.
* SystemdService: for `kubelet`, `docker`, default to the component name; for `cri` this is required if `EnableRepair` is enabled.

And the repair will just restart the systemd service.

pkg/healthchecker/health_checker.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+	return hc, nil
+}
+
+// CheckHealth checks for the health of the component and tries to repair if enabled.
+// Returns true if healthy, false otherwise.
+func (hc *healthChecker) CheckHealth() bool {
+	// Poll till the health check timeout for the component to be up.
+	if err := wait.PollImmediate(hc.healthCheckTimeout, hc.healthCheckTimeout, func() (bool, error) {
+		healthy := hc.healthCheckFunc()
```
Random-Liu, line 69: What if the `healthCheckFunc` function itself stuck? We need a timeout to cancel that.

This is very important, because that is the most frequent way the health check would fail.

pkg/healthchecker/health_checker.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+		return false, nil
+	}); err != nil {
+		// The service is unhealthy.
+		// Attempt repair based on flag.
+		if hc.enableRepair {
+			glog.Infof("health-checker: component is unhealthy, proceeding to repair")
+			hc.repairFunc()
+			// stall for cool down period after repairing
+			time.Sleep(hc.coolDownTime)
```
Random-Liu, line 81: Hm, I feel like the status should be reported before the cooldown.

Can we use the systemd service startup time to implement cooldown? If the service hasn't run for 2 min yet, we don't repair it.

For example:
`systemctl show docker --property=ActiveEnterTimestamp`

pkg/healthchecker/health_checker.go
```diff
@@ -0,0 +1,146 @@
...
+		return func() bool {
+			if err := execCommand("docker", "ps"); err != nil {
+				return false
+			}
+			return true
+		}
+	}
+	// Use "crictl pods" for containerd health check.
+	return func() bool {
+		if err := execCommand(crictlPath, "pods"); err != nil {
```
Random-Liu, line 117: Where is the CRI socket path configured? Are we going to assume it is configured on the node?

cmd/healthchecker/options/options.go
abansal4032: Was a typo there. Updated the flags.

abansal4032: Done.

cmd/healthchecker/options/options.go
abansal4032: Added the flag for CRI socket path.

pkg/healthchecker/health_checker.go
abansal4032: Included the timeout logic in the healthCheckFunc. Removed the polling in the wrapper function.

pkg/healthchecker/health_checker.go
abansal4032: Done. Since we are attempting the repair only if the component has been up for cool down period, the status report will not wait for cool down period.

pkg/healthchecker/health_checker.go
abansal4032: Added a flag to specify this.

cmd/healthchecker/options/options.go
abansal4032: Restructured the logic to include the suggested changes.

pkg/healthchecker/types/types.go
```diff
@@ -0,0 +1,37 @@
...
+package types
+
+import "time"
+
+const (
+	DefaultCoolDownTime        = 2 * time.Minute
+	DefaultHealthCheckTimeout  = 10 * time.Second
+	DefaultCmdTimeout          = 10 * time.Second
+	DefaultCriCtl              = "/usr/bin/crictl"
+	DefaultCricSocketPath      = "unix:///var/run/containerd/containerd.sock"
```
Random-Liu, line 26: s/DefaultCricSocketPath/DefaultCriSocketPath

cmd/healthchecker/options/options.go
```diff
@@ -0,0 +1,104 @@
...
+
+// NewHealthCheckerOptions returns an empty health check options struct.
+func NewHealthCheckerOptions() *HealthCheckerOptions {
+	return &HealthCheckerOptions{}
+}
+
+// HealthCheckerOptions are the options used to configure the health checker.
+type HealthCheckerOptions struct {
+	Component          string
+	SystemService      string
```
Random-Liu, line 37: s/SystemService/SystemdService

cmd/healthchecker/options/options.go
```diff
@@ -0,0 +1,104 @@
...
+	CoolDownTime       time.Duration
+	HealthCheckTimeout time.Duration
+	CmdTimeout         time.Duration
+}
+
+// AddFlags adds health checker command line options to pflag.
+func (hco *HealthCheckerOptions) AddFlags(fs *pflag.FlagSet) {
+	fs.StringVar(&hco.Component, "component", types.KubeletComponent,
+		"The component to check health for. Supports kubelet, docker and cri")
+	fs.StringVar(&hco.SystemService, "system-service", "",
```
Random-Liu, line 50: s/system-service/systemd service.

cmd/healthchecker/options/options.go
```diff
@@ -0,0 +1,104 @@
...
+		"The path to the cri socket. Used with crictl to specify the socket path.")
+	fs.DurationVar(&hco.CoolDownTime, "cooldown-time", types.DefaultCoolDownTime,
+		"The duration to wait for the service to be up before attempting repair.")
+	fs.DurationVar(&hco.HealthCheckTimeout, "health-check-timeout", types.DefaultHealthCheckTimeout,
+		"The time to wait before marking the component as unhealthy.")
+	fs.DurationVar(&hco.CmdTimeout, "cmd-timeout", types.DefaultCmdTimeout,
+		"The time to wait for the exec commands to complete.")
+}
+
+// ValidOrDie validates health checker command line options.
```
Random-Liu, line 65: Update the comment.

pkg/healthchecker/health_checker.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+}
+
+// getHealthCheckFunc returns the health check function based on the component.
+func getHealthCheckFunc(hco *options.HealthCheckerOptions) func(time.Duration) bool {
+	switch hco.Component {
+	case types.KubeletComponent:
+		return kubeletHealthCheck
+	case types.DockerComponent:
+		return func(timeout time.Duration) bool {
```
Random-Liu, line 102: 1) Can we make the way we use timeout consistent among the 3 functions? Maybe outside function accept `HealthCheckerOptions`, internal function doesn't accept arguments?
2) What is the relationship between `HealthCheckTimeout` and `CmdTimeout`? It is a bit confusing. :P

pkg/healthchecker/types/types.go
abansal4032: Done.

cmd/healthchecker/options/options.go
abansal4032: Done.

cmd/healthchecker/options/options.go
abansal4032: Done.

cmd/healthchecker/options/options.go
abansal4032: Done.

pkg/healthchecker/health_checker.go
abansal4032: 1. Done. Made the kubelet function inline.
2. HealthCheckTimeout is used to timeout actions in health check function. CmdTimeout is the timeout used for all other CLI commands run. For example : getUptime. Kept these two different because we might want to have a longer timeout on health check than other commands.

Random-Liu: I know what they are, but it is hard for users to understand when see the flag description.

Do we need to make command timeout configurable? Maybe have a constant value for it? I don't think people will need to tweak it. If they do, we can think about how to name that flag then.

cmd/healthchecker/options/options_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"k8s.io/node-problem-detector/pkg/healthchecker/types"
+)
+
+func TestValidOrDie(t *testing.T) {
```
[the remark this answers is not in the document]
abansal4032: Done.

pkg/healthchecker/health_checker.go
abansal4032: Sounds good. Made it a constant and removed from the flags.

Issue #304, opened 2019-06-28
Node conditions status reset on problem detector restart.


eatwithforks:
When node problem detector is restarted, all the node's conditions' status reset to False until the plugin's complete and resubmit the correct status. 

This means there is a gap where node conditions are all False when problem detector restarts, leading to false reports.

Is there a way to make problem detector not reset a node's conditions on restart?

grosser:
`reset to False` -> `reset to their initial status`

xueweiz:
One idea is to use the [`problemclient.Client.GetConditions()`](https://github.com/kubernetes/node-problem-detector/blob/146dfd70b25090203162b1e9eee305e76b688dff/pkg/exporters/k8sexporter/problemclient/problem_client.go#L81) function to retrieve condition's initial state upon NPD startup.

Basically each time when NPD starts up, it could call the `GetConditions()` function to determine what are the node conditions before NPD starts up. And use the result as the initial status for node conditions.

Would that work?

eatwithforks:
sounds good, that would work.

grosser:
only downside could be something like a log-status that then gets reset because the offending line is no longer in the log, but that's just a minor issue 🤷‍♂

xueweiz:
/cc @andyxning 
Hi Andy, do you think the [idea above](https://github.com/kubernetes/node-problem-detector/issues/304#issuecomment-506883788) makes sense? If so, I'm happy to make a patch to implement it.

> only downside could be something like a log-status that then gets reset because the offending line is no longer in the log
True. But if that's the case (i.e. some problem daemon resets condition when it can no longer detect the permanent problem), then even if NPD is not restarted, it will still reset that condition eventually (when the log got deleted/rotated).
So, in my opinion, I think the reset behavior (feature-or-bug) is not in the scope of this issue. :P

Do you have any suggestions @grosser ?

grosser:
Good enough if it does not resets it on restart 👍 
(also avoid the whole confusion with lastTransition getting updated when nothing actually happened)

grosser:
@wangzhen127 got thoughts on this ? (since @andyxning seems to not be around)

wangzhen127:
Are you referring to the log monitor? NPD resets the condition because in the config, it has the default condition. For example:
https://github.com/kubernetes/node-problem-detector/blob/master/config/kernel-monitor.json#L10
```
	"conditions": [
		{
			"type": "KernelDeadlock",
[...]

Issue #428, opened 2020-05-18
Need a way to differentiate "not run yet" from "passed"


eatwithforks:
with the recent change (https://github.com/kubernetes/node-problem-detector/pull/320), we can no longer use the default condition reason to know if a rule already executed or not. Is there a way to do that? Otherwise would you accept a PR to add a way to determine if a check already ran or not?

PR #429, opened 2020-05-19
Allow not setting initial condition to avoid marking unhealthy nodes …

eatwithforks:
…as healthy

Fixes https://github.com/kubernetes/node-problem-detector/issues/304 and https://github.com/kubernetes/node-problem-detector/issues/428

Previously, when a node boots up. NPD sets all the custom plugin conditions to false. This would lead to marking the node as healthy when it is not known yet, because the custom checks have not ran yet. Even worse, when a node is unhealthy and NPD restarts, the node is marked as healthy.

Adding a flag option to not set initial conditions to avoid that behavior.

If this turns out fine, I would also suggest making this the new default in the next major version.

@wangzhen127 @andyxning

### DIFF
--- a/pkg/custompluginmonitor/custom_plugin_monitor.go
+++ b/pkg/custompluginmonitor/custom_plugin_monitor.go
@@ -286,11 +286,20 @@ func toConditionStatus(s cpmtypes.Status) types.ConditionStatus {
 func (c *customPluginMonitor) initializeStatus() {
 	// Initialize the default node conditions
 	c.conditions = initialConditions(c.config.DefaultConditions)
-	glog.Infof("Initialize condition generated: %+v", c.conditions)
+
+	// Initialize all conditions to their default state
+	initializedConditions := []types.Condition{}
+	for _, cond := range c.conditions {
+		if !cond.Uninitialized {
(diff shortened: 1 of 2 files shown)


eatwithforks:
I signed it

eatwithforks:
I signed it

eatwithforks:
@xueweiz @andyxning @Random-Liu could you take a look when you have the time? I tested this by building it locally to verify that it works. Thank you.

eatwithforks:
Hi @xueweiz @andyxning @Random-Liu, pinging again in case my last message didn't reach

grosser:
@wangzhen127 all green, good to go ?

wangzhen127:
Would setting the default condition to empty be working for this case? Like https://github.com/kubernetes/node-problem-detector/blob/f42281ee2658900bdb0571e1159a43f6ab712a19/config/network-problem-monitor.json#L11

grosser:
`Failed to validate custom plugin config`

grosser:
@wangzhen127 so good to go ?

grosser:
@Random-Liu can you take a look at this ?

PR #433, opened 2020-06-09
docker image: add health-checker binary

bengadbois:
The health-checker binary added in v0.8.2 is not present in `k8s.gcr.io/node-problem-detector:v0.8.2`. The example at https://github.com/kubernetes/node-problem-detector/blob/master/config/health-checker-kubelet.json#L23 suggests is should be available, so this adds it at that path within the image.


wangzhen127:
/ok-to-test
/lgtm

Thanks for catching that!

PR #441, opened 2020-07-16
Generate new status log only on condition change

abansal4032:
This is to limit the logging from custom plugin runs. Today every run is logged with the status, we can instead log the status only when there is a change in conditions.

### DIFF
--- a/pkg/custompluginmonitor/custom_plugin_monitor.go
+++ b/pkg/custompluginmonitor/custom_plugin_monitor.go
@@ -135,7 +135,7 @@ func (c *customPluginMonitor) monitorLoop() {
 			}
 			glog.V(3).Infof("Receive new plugin result for %s: %+v", c.configPath, result)
 			status := c.generateStatus(result)
-			glog.Infof("New status generated: %+v", status)
+			glog.V(3).Infof("New status generated: %+v", status)
 			c.statusChan <- status
 		case <-c.tomb.Stopping():
 			c.plugin.Stop()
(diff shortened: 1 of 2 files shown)


pkg/custompluginmonitor/custom_plugin_monitor.go
```diff
@@ -135,7 +135,6 @@ func (c *customPluginMonitor) monitorLoop() {
 			}
 			glog.V(3).Infof("Receive new plugin result for %s: %+v", c.configPath, result)
 			status := c.generateStatus(result)
-			glog.Infof("New status generated: %+v", status)
```
Random-Liu, line 138: Can we keep this line with lower log level?

pkg/custompluginmonitor/custom_plugin_monitor.go
```diff
@@ -263,12 +265,17 @@ func (c *customPluginMonitor) generateStatus(result cpmtypes.Result) *types.Stat
...
 		}
 	}
-	return &types.Status{
+	status := &types.Status{
 		Source: c.config.Source,
 		// TODO(random-liu): Aggregate events and conditions and then do periodically report.
 		Events:     append(activeProblemEvents, inactiveProblemEvents...),
 		Conditions: c.conditions,
 	}
+	// Log only if condition has changed for permanent error
```
Random-Liu, line 274: Can we check whether `activeProblemEvents` and `inactiveProblemEvents` are empty? I think that should cover all cases.

pkg/custompluginmonitor/custom_plugin_monitor.go
abansal4032: Done

pkg/custompluginmonitor/custom_plugin_monitor.go
abansal4032: Switched to checking for active and inactive problems. They so cover all the cases.

PR #442, opened 2020-07-16
Capture the logs from stderr of custom plugins

abansal4032:
The custom plugins : health-checker and log-counter currently log to stderr and the logs are not captured in NPD logs. This change allows to read the stderr from the custom plugin and log the output.

### DIFF
--- a/pkg/custompluginmonitor/plugin/plugin.go
+++ b/pkg/custompluginmonitor/plugin/plugin.go
@@ -19,6 +19,8 @@ package plugin
 import (
 	"context"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"os/exec"
 	"strings"
 	"sync"
@@ -29,6 +31,10 @@ import (
 	cpmtypes "k8s.io/node-problem-detector/pkg/custompluginmonitor/types"
 	"k8s.io/node-problem-detector/pkg/util/tomb"
 )
+
+// maxCustomPluginBufferBytes is the max bytes that a custom plugin is allowed to
+// send to stdout/stderr. Any bytes exceeding this value will be truncated.
+const maxCustomPluginBufferBytes = 1024 * 4
 
 type Plugin struct {
 	config     cpmtypes.CustomPluginConfig
@@ -115,6 +121,20 @@ func (p *Plugin) runRules() {
 	glog.Info("Finish running custom plugins")
 }
 
+// readFromReader reads the maxBytes from the reader and drains the rest.
+func readFromReader(reader io.ReadCloser, maxBytes int64) ([]byte, error) {
+	limitReader := io.LimitReader(reader, maxBytes)
+	data, err := ioutil.ReadAll(limitReader)
+	if err != nil {
+		return []byte{}, err
+	}
+	// Drain the reader
+	if _, err := io.Copy(ioutil.Discard, reader); err != nil {
+		return []byte{}, err
+	}
+	return data, nil
+}
+
 func (p *Plugin) run(rule cpmtypes.CustomRule) (exitStatus cpmtypes.Status, output string) {
 	var ctx context.Context
 	var cancel context.CancelFunc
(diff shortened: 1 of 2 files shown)


pkg/custompluginmonitor/plugin/plugin.go
```diff
@@ -127,16 +128,21 @@ func (p *Plugin) run(rule cpmtypes.CustomRule) (exitStatus cpmtypes.Status, outp
 	defer cancel()
 
 	cmd := exec.CommandContext(ctx, rule.Path, rule.Args...)
-	stdout, err := cmd.Output()
-	if err != nil {
+	var stdout, stderr bytes.Buffer
+	cmd.Stdout = &stdout
```
Random-Liu, line 132: I have some concern about this, especially stderr. If the stderr is too big, it could potentially cause high memory usage.

Can we implement a limit reader to limit the output size, and discard other outputs?

abansal4032: Added a limit reader for stdout and stderr limiting each to 4k bytes.

pkg/custompluginmonitor/plugin/plugin.go
```diff
@@ -127,14 +133,50 @@ func (p *Plugin) run(rule cpmtypes.CustomRule) (exitStatus cpmtypes.Status, outp
...
+		return cpmtypes.Unknown, "Error creating stderr pipe for plugin. Please check the error log"
+	}
+	stdoutReader := io.LimitReader(stdoutPipe, maxCustomPluginBufferBytes)
+	stderrReader := io.LimitReader(stderrPipe, maxCustomPluginBufferBytes)
+
+	if err := cmd.Start(); err != nil {
+		glog.Errorf("Error in starting plugin %q: error - %v", rule.Path, err)
+		return cpmtypes.Unknown, "Error in starting plugin. Please check the error log"
+	}
+
```
Random-Liu, line 154: Should we read stdout and stderr at the same time? Or else, the stderr side may potentially block the plugin process when the buffer is full?

And we should probably continue reading the output and discard them after reaching the limit, or else the plugin process may be blocked as well.

abansal4032: Makes sense. Reading from stdout and stderr at the same time.
Also added discard once we have read the maxBytes.

pkg/custompluginmonitor/plugin/plugin.go
```diff
@@ -127,14 +147,62 @@ func (p *Plugin) run(rule cpmtypes.CustomRule) (exitStatus cpmtypes.Status, outp
...
 	if err != nil {
+		glog.Errorf("Error creating stderr pipe for plugin %q: error - %v", rule.Path, err)
+		return cpmtypes.Unknown, "Error creating stderr pipe for plugin. Please check the error log"
+	}
+	if err := cmd.Start(); err != nil {
+		glog.Errorf("Error in starting plugin %q: error - %v", rule.Path, err)
+		return cpmtypes.Unknown, "Error in starting plugin. Please check the error log"
+	}
+
+	var wg sync.WaitGroup
```
Random-Liu, line 166: nit:
```
var (
  stdout []byte
  stderr []byte
  ...
)
```

pkg/custompluginmonitor/plugin/plugin.go
```diff
@@ -127,14 +147,62 @@ func (p *Plugin) run(rule cpmtypes.CustomRule) (exitStatus cpmtypes.Status, outp
...
-			return cpmtypes.Unknown, "Error in running plugin. Please check the error log"
+			glog.Errorf("Error in waiting for plugin %q: error - %v. output - %q", rule.Path, err, string(stdout))
+			return cpmtypes.Unknown, "Error in waiting for plugin. Please check the error log"
 		}
 	}
 
+	// log the stderr from the plugin
+	if len(stderr) != 0 {
+		glog.Infof("Start logs from plugin %q \n %s\n", rule.Path, string(stderr))
+		glog.Infof("End logs from plugin %q\n", rule.Path)
```
Random-Liu, line 203: Do we need new line for this log line?

pkg/custompluginmonitor/plugin/plugin.go
```diff
@@ -127,14 +147,62 @@ func (p *Plugin) run(rule cpmtypes.CustomRule) (exitStatus cpmtypes.Status, outp
...
+	wg.Add(2)
+	go func() {
+		stdout, stdoutErr = readFromReader(stdoutPipe, maxCustomPluginBufferBytes)
+		wg.Done()
+	}()
+	go func() {
+		stderr, stderrErr = readFromReader(stderrPipe, maxCustomPluginBufferBytes)
+		wg.Done()
+	}()
+	wg.Wait()
```
Random-Liu, line 181: It seems that this may block forever.

In theory, it should not happen, because the command should be stopped after timeout, and the IO should be closed.

However, I'm not completely sure about that. Maybe we can call `wg.Wait` after `Wait` the command to stop?

abansal4032: The documentation for stdoutpipe specifies: https://golang.org/src/os/exec/exec.go?s=17733:17782#L609
It is incorrect to call Wait before all reads from the pipe have completed.

Having the wg.Wait() after will call exec.Wait before we are done reading from pipes?

pkg/custompluginmonitor/plugin/plugin.go
abansal4032: Removed the redundant line. Did not realize Infof appends that automatically.

pkg/custompluginmonitor/plugin/plugin.go
abansal4032: Done.

PR #454, opened 2020-08-06
add automaxprocs

smileusd:
We find npd use lots of threads in our kuberentes cluster. This pr can limit the npd to use limit procs.
1. If limit of npd pod spec =< 1, the max proc = 1
2. If limit of npd pod spec > 1, the max proc = floor(limit)


Random-Liu:
> We find npd use lots of threads in our kuberentes cluster. This pr can limit the npd to use limit procs.

Is it because you've run a lot of custom plugins?

smileusd:
> > We find npd use lots of threads in our kuberentes cluster. This pr can limit the npd to use limit procs.
> 
> Is it because you've run a lot of custom plugins?

No, we use few plugin of file watcher. Increase the config number of plugin will increase the number of threads, but it is not very much. The most waste is our cluster nodes have 80 cpu but the NPD only use less 10 P. But it still create 80 P and about 60+ threads to serve the program. And in golang the idle threads can not be clean. https://github.com/golang/go/issues/14592

PR #456, opened 2020-08-25
fix: fix script timeout can't work

### LINKED ISSUE #372: custom plugin is blocked due to the timeout of my plugin script
I'm beginning to learn and test npd. When I try to use custom plugin and its timeout, I found an issue which I'm not sure it is a 'real' issue.

npd start cmd
`bin/node-problem-detector --v=2 --logtostderr --enable-k8s-exporter=false  --config.custom-plugin-monitor=./config/custom-plugin-monitor.json` 

#cat ./config/custom-plugin-monitor.json
{
  "plugin": "custom",
  "pluginConfig": {
    "invoke_interval": "30s",
    "timeout": "20s",
    "max_output_length": 80,
    "concurrency": 2,
    "enable_message_change_based_condition_update": false
  },
  "metricsReporting": true,
  "rules": [
    {
      "type": "temporary",
      "reason": "NTPIsDown",
      "path": "./config/plugin/check_ntp.sh",
      "timeout": "10s"
    },
    {
      "type": "temporary",
      "reason": "PidUsageExceed",
      "path": "./config/plugin/check_pid.sh",
      "timeout": "15s"
    },
    {
      "type": "temporary",
      "reason": "CollectdRunOnDocker",
      "path": "./config/plugin/check_collectd_for_docker.sh",
      "timeout": "8s"
    }
  ]
}

There are 3 scripts: check_ntp.sh, check_pid.sh, check_collectd_for_docker.sh.
I injected "sleep 200" into check_ntp.sh
```
#cat ./config/plugin/check_ntp.sh
#!/bin/bash

OK=0
NONOK=1
UNKNOWN=2

sleep 200

which systemctl >/dev/null
if [ $? -ne 0 ]; then
    echo "Systemd is not supported"
    exit $UNKNOWN
fi

systemctl status ntp.service | grep 'Active:' | grep -q running
if [ $? -ne 0 ]; then
    echo "NTP service is not running"
    exit $NONOK
fi

echo "NTP service is running"
exit $OK
```

When time is up to 10s,  npd sends kill signal to check_ntp.sh. Script check_ntp.sh can be killed successfully, but its child process "sleep 200" is still running which new parent is the process pid 1. 
The result is that custom plugin is blocked until "sleep 200" ends although check_ntp.sh is killed. After 200 seconds, I can see log
`Message:Timeout when running plugin "./config/plugin/check_ntp.sh": state - signal: kill`

So I think there are two problems in this scenario:
1. Script in custom plugin rules will be still waiting after timeout.
2. Other scripts are also blocked. They will not be invoked as expected in "invoke_interval 30s" and "concurrency 2".

IMO, Can we use the action as "kill -TERM -PID" to send kill signal to script and its children processes when timeout comes? Or one script's blocking won't block other scripts.

ZYecho:
Signed-off-by: zhangyue <huaihuan.zy@alibaba-inc.com>
fix #372, the io hang because of child process holds the pipe of stdout/stderr.
in this pr, use process group to create shell, and if timeout, send signal to whole process group.
I test on local machine with scripts which #372 provided, the problem don't happen.

### DIFF
--- a/pkg/custompluginmonitor/plugin/plugin.go
+++ b/pkg/custompluginmonitor/plugin/plugin.go
@@ -148,7 +148,12 @@ func (p *Plugin) run(rule cpmtypes.CustomRule) (exitStatus cpmtypes.Status, outp
 	}
 	defer cancel()
 
-	cmd := exec.CommandContext(ctx, rule.Path, rule.Args...)
+	// create a process group
+	sysProcAttr := &syscall.SysProcAttr{
+		Setpgid: true,
+	}
+	cmd := exec.Command(rule.Path, rule.Args...)
+	cmd.SysProcAttr = sysProcAttr
 
 	stdoutPipe, err := cmd.StdoutPipe()
 	if err != nil {
@@ -164,6 +169,26 @@ func (p *Plugin) run(rule cpmtypes.CustomRule) (exitStatus cpmtypes.Status, outp
 		glog.Errorf("Error in starting plugin %q: error - %v", rule.Path, err)
 		return cpmtypes.Unknown, "Error in starting plugin. Please check the error log"
 	}
+
+	waitChan := make(chan struct{})
+	defer close(waitChan)
+
+	go func() {
+		select {
+		case <-ctx.Done():
+			glog.Errorf("Error in running plugin timeout %q", rule.Path)
+			if cmd.Process == nil || cmd.Process.Pid == 0 {
+				glog.Errorf("Error in cmd.Process check %q", rule.Path)
+				break
+			}
+			err := syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL)
+			if err != nil {
+				glog.Errorf("Error in kill process %d, %v", cmd.Process.Pid, err)
+			}
+		case <-waitChan:
+			return
+		}
+	}()
 
 	var (
 		wg        sync.WaitGroup


ZYecho:
@xueweiz ptal, thanks

ForestCold:
I can reproduced #372 and also verified this PR fix the issue, @Random-Liu @vteratipally could you review/approve it?

pkg/custompluginmonitor/plugin/plugin.go
```diff
@@ -146,25 +146,52 @@ func (p *Plugin) run(rule cpmtypes.CustomRule) (exitStatus cpmtypes.Status, outp
...
+			if cmd.Process == nil || cmd.Process.Pid == 0 {
+				glog.Errorf("Error in cmd.Process check %q", rule.Path)
+				break
+			}
+			err := syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL)
+			if err != nil {
+				glog.Errorf("Error in kill process %d, %v", cmd.Process.Pid, err)
+			}
+		case <-waitChan:
+			cancel()
```
Random-Liu, line 191: How about we just return here? In that way, the cancel code can be unchanged?

pkg/custompluginmonitor/plugin/plugin.go
```diff
@@ -146,25 +146,52 @@ func (p *Plugin) run(rule cpmtypes.CustomRule) (exitStatus cpmtypes.Status, outp
...
+
+	go func() {
+		select {
+		case <-ctx.Done():
+			glog.Errorf("Error in running plugin timeout %q", rule.Path)
+			if cmd.Process == nil || cmd.Process.Pid == 0 {
+				glog.Errorf("Error in cmd.Process check %q", rule.Path)
+				break
+			}
+			err := syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL)
```
Random-Liu, line 186: What if the process exited just before this? We may want to check the error code and take that into consideration.

Random-Liu: And can we make the wait a helper function?

Random-Liu:
Hm, this seems to be a long standing golang issue https://github.com/golang/go/issues/23019.

Thanks for sending the fix!

pkg/custompluginmonitor/plugin/plugin.go
ZYecho: if process exited just before this, we can still get the result from cmd.wait and pipe in it's parent goroutine.
in this case, the kill operation is useless but no side effects.

pkg/custompluginmonitor/plugin/plugin.go
ZYecho: ok

PR #468, opened 2020-09-18
Adding disk_quota metric to get the disk size

vteratipally:
Collecting FSType and Mountoptions would be helpful to track the use of file system types. But adding fields to the existing metric is a breaking change because the fields would be suprising to the customers using with alerts. So removed the fstypes and mountoptions on the existing metric 'disk_bytes_used'.
Collecting the size of the boot disk would help in identifying the inconsistencies of the file system with the device. Adding metric 'disk_bytes_total' to track the size of the boot disk with the fstypes and mountoptions.

### DIFF
--- a/pkg/systemstatsmonitor/README.md
+++ b/pkg/systemstatsmonitor/README.md
@@ -40,7 +40,7 @@ Below metrics are collected from `disk` component:
 * `disk_operation_bytes_count`: # of Bytes used for reads/writes on this device
 * `disk_operation_time`: [# of milliseconds spent reading/writing][iostat doc]
 * `disk_bytes_used`: Disk usage in Bytes. The usage state is reported under the `state` metric label (e.g. `used`, `free`). Summing values of all states yields the disk size.
-FSType and MountOptions are also reported as additional information.
+* `disk_bytes_total`: Total disk space in Bytes. Additional information such as fstypes and mountoptions is also reported
 
 The name of the disk block device is reported in the `device_name` metric label (e.g. `sda`).
 
--- a/pkg/systemstatsmonitor/disk_collector.go
+++ b/pkg/systemstatsmonitor/disk_collector.go
@@ -38,6 +38,7 @@ type diskCollector struct {
 	mOpsBytes       *metrics.Int64Metric
 	mOpsTime        *metrics.Int64Metric
 	mBytesUsed      *metrics.Int64Metric
+	mBytesTotal     *metrics.Int64Metric
 
 	config *ssmtypes.DiskStatsConfig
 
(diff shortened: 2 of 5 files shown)


pkg/systemstatsmonitor/disk_collector.go
```diff
@@ -145,11 +146,22 @@ func NewDiskCollectorOrDie(diskConfig *ssmtypes.DiskStatsConfig) *diskCollector
...
-		[]string{deviceNameLabel, fsTypeLabel, mountOptionLabel, stateLabel})
+		[]string{deviceNameLabel, stateLabel})
 	if err != nil {
 		glog.Fatalf("Error initializing metric for %q: %v", metrics.DiskBytesUsedID, err)
 	}
 
+	dc.mBytesTotal, err = metrics.NewInt64Metric(
+		metrics.DiskBytesTotalID,
+		diskConfig.MetricsConfigs[string(metrics.DiskBytesTotalID)].DisplayName,
+		"Disk bytes used, in Bytes",
```
xueweiz, line 157: Maybe need a change on the description?

pkg/systemstatsmonitor/README.md
```diff
@@ -40,7 +40,7 @@ Below metrics are collected from `disk` component:
 * `disk_operation_bytes_count`: # of Bytes used for reads/writes on this device
 * `disk_operation_time`: [# of milliseconds spent reading/writing][iostat doc]
 * `disk_bytes_used`: Disk usage in Bytes. The usage state is reported under the `state` metric label (e.g. `used`, `free`). Summing values of all states yields the disk size.
-FSType and MountOptions are also reported as additional information.
+* `disk_bytes_total`: Total disk space in Bytes. Additional information such as fstypes and mountoptions is also reported
```
xueweiz, line 43: Why do we need this metric? I believe `disk_bytes_used[state='used']+disk_bytes_used[state='free']` should yield the result, right?
I believe the general principle is that the instrumentation should report the most fundamental & raw & duplication-free data, and we can use backend (Prometheus, Stackdriver, etc) to do more processing and piece them togerther.

pkg/systemstatsmonitor/disk_collector.go
```diff
@@ -278,8 +293,9 @@ func (dc *diskCollector) collect() {
 		deviceName := strings.TrimPrefix(partition.Device, "/dev/")
 		fstype := partition.Fstype
 		opttypes := partition.Opts
-		dc.mBytesUsed.Record(map[string]string{deviceNameLabel: deviceName, fsTypeLabel: fstype, mountOptionLabel: opttypes, stateLabel: "free"}, int64(usageStat.Free))
```
xueweiz, line 281: Could you let me know the reason for stop reporting `mountOptionLabel`?
And since we are essentially changing the definition of an existing metric, maybe we should also document this on the commit message? :P 
It'd be a bit surprising for folks to find out a commit that "adding xxx metric" changed the label of another existing metric lol. WDYT?

xueweiz (changes requested):
The code changes LGTM overall. But I think I'm missing some context here, as I couldn't figure out why are we doing 2 things:
1. Why are we adding a new metric while the information is already there in `disk_bytes_used`?
2. Why change the existing label on `disk_bytes_used`?

I think you probably have some good reasons, but sorry for my lack of context. I think it'd be helpful to document the context/reasoning on the commit message or PR description (or file an issue if that's not too much trouble for you), so that future people can also find the context easily.

vteratipally:
> The code changes LGTM overall. But I think I'm missing some context here, as I couldn't figure out why are we doing 2 things:
> 
> 1. Why are we adding a new metric while the information is already there in `disk_bytes_used`?
> 2. Why change the existing label on `disk_bytes_used`?
> 
> I think you probably have some good reasons, but sorry for my lack of context. I think it'd be helpful to document the context/reasoning on the commit message or PR description (or file an issue if that's not too much trouble for you), so that future people can also find the context easily.


Collecting Fstypes and mountoptions would be helpful for the developers to track the use of file system types either it be XFS or ext4. Adding labels to the existing metric would be a breaking change because there could be alerts and that might be suprising to see new labels in the exisiting metrics.  In a similar way, it would be helpful to collect the boot disk size as well, so adding a new metric like that helps us track the fstypes, mountoptions, size would be useful. This should be more like extension of the disk_bytes_used, may be the name is little misleading.

pkg/systemstatsmonitor/README.md
vteratipally: done

pkg/systemstatsmonitor/disk_collector.go
vteratipally: done

pkg/systemstatsmonitor/disk_collector.go
vteratipally: changed the commit message.

Issue #471, opened 2020-10-03
v0.8.4 release added labels to the existing metric is a breaking change


vteratipally:
Adding labels to the already existing metric disk_bytes_used is a breaking change and also which should report the size without other options. This is because the cardinality increases for the existing metric.

Issue #467, opened 2020-09-17
flakey npd tests in sig-node ci jobs


MHBauer:
I've documented at k/k repo before realizing this has it's own repo. Maybe you can help me debug?
https://github.com/kubernetes/kubernetes/issues/94493
Are these tests supposed to be running outside of the ci-npd jobs?
Do they need to be ignored?
Is there some special setup that is done?
Should they be marked [serial] or [disruptive]?

vteratipally:
these tests might be added as NPD runs as a default on kubernetes. I will try to debug the issue. I dont think there is any special setup done here.

SergeyKanzhelev:
@vteratipally can you comment on why serial is needed? What tests do that wouldn't run in parellel? Can the un-parallellable resources be named somehow to separate tests?

vteratipally:
Yes can we can try to separate the tests to parallelable resource because I believe that these tests are based on the event logs. I can give it a try

SergeyKanzhelev:
CC: @ForestCold and @vteratipally FYI

PR #472, opened 2020-10-03
Fix for issue https://github.com/kubernetes/node-problem-detector/issues/471

vteratipally:
Removing the labels fstype and options as the disk_bytes_used metric should report only size


vteratipally:
@bsdnet

vteratipally:
https://github.com/kubernetes/node-problem-detector/issues/467

PR #475, opened 2020-10-08
catching hung task with pattern like "tasks airflow scheduler: *"

vteratipally:
some of the events of the pattern "task airflow scheduler: *" is not been identified. changing the pattern to catch this type events

### DIFF
--- a/config/kernel-monitor.json
+++ b/config/kernel-monitor.json
@@ -26,7 +26,7 @@
 		{
 			"type": "temporary",
 			"reason": "TaskHung",
-			"pattern": "task [\\S\\s]+:\\w+ blocked for more than \\w+ seconds\\."
+			"pattern": "task [\\S ]+:\\w+ blocked for more than \\w+ seconds\\."
 		},
 		{
 			"type": "temporary",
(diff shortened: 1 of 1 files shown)


config/kernel-monitor.json
```diff
@@ -26,7 +26,7 @@
 		{
 			"type": "temporary",
 			"reason": "TaskHung",
-			"pattern": "task \\S+:\\w+ blocked for more than \\w+ seconds\\."
+			"pattern": "task [\\S\\s]+:\\w+ blocked for more than \\w+ seconds\\."
```
Random-Liu, line 29: This basically means matching everything.

I guess we only want to match \S + space.

vteratipally: but it was unable to identify the cases where there is a space like this task airflow scheduler:

vteratipally: i mean it is technically \S + space + \S + ":*"

Random-Liu: I mean, we can do something like `[\\S ]+`?

vteratipally: yes we can use this

Issue #480, opened 2020-10-22
NPD doesn't notice OOMs in newer Linux kernels


tosi3k:
Sample OOM message log part from `4.19.112`:
```
Memory cgroup out of memory: Kill process 4015 (python) score 1923 or sacrifice child
Killed process 4015 (python) total-vm:32812kB, anon-rss:23588kB, file-rss:1684kB, shmem-rss:0kB
```

Sample OOM message log part from `5.4.49`:
```
Memory cgroup out of memory: Killed process 4249 (python) total-vm:32812kB, anon-rss:23856kB, file-rss:1700kB, shmem-rss:0kB, UID:0 pgtables:112kB oom_score_adj:994
```

NPD parses kernel log file on the fly looking for its parts matching the regexes defined in [kernel-monitor-filelog.json](https://github.com/kubernetes/node-problem-detector/blob/master/config/kernel-monitor-filelog.json). The one of interest here is this one:
```
{
	"type": "temporary",
	"reason": "OOMKilling",
	"pattern": "Kill process \\d+ (.+) score \\d+ or sacrifice child\\nKilled process \\d+ (.+) total-vm:\\d+kB, anon-rss:\\d+kB, file-rss:\\d+kB.*"
}
```

This explains why the NPD manages to report an OOM in the former case and doesn't notice the OOM in the latter case (`Kill process \\d+ (.+) score \\d+ or sacrifice child\\n` part is absent).

I see three possible options to fix this:
* get rid of `Kill process \\d+ (.+) score \\d+ or sacrifice child\\n` part of the OOM regex,
* make this part of the regex optional,
* create an additional rule in [kernel-monitor-filelog.json](https://github.com/kubernetes/node-problem-detector/blob/master/config/kernel-monitor-filelog.json) to cover the new OOM logging format as well (probably would require adapting the first rule's regex pattern as well).

Third approach seems most sensible if feasible.

/cc @jkaniuk
/assign

tosi3k:
Looks like the aforementioned `Kill process %d (%s) score %u or sacrifice child\n` OOM message part was removed in https://github.com/torvalds/linux/commit/bbbe48029720d2c6b6733f78d02571a281511adb kernel commit (5.1 release candidate).

[...]

PR #481, opened 2020-10-22
Adapt OOMKilling pattern to old and new Linux kernels

tosi3k:
Context: https://github.com/kubernetes/node-problem-detector/issues/480.

Tl;dr: kernel commit https://github.com/torvalds/linux/commit/bbbe48029720d2c6b6733f78d02571a281511adb removed logging the `Kill process %d (%s) score %u or sacrifice child\n` OOM message part we look for when tracing the kernel log in the kernel monitor. Because of that, the change removes that part from the `kernel-monitor(-filelog).json` regex pattern to handle both old and new kernel versions.

Successfully tested the change using https://github.com/kubernetes/node-problem-detector#try-it-out instruction (appropriate K8s events are emitted for both old and new OOM message formats).

Fixes https://github.com/kubernetes/node-problem-detector/issues/480.

### DIFF
--- a/config/kernel-monitor-filelog.json
+++ b/config/kernel-monitor-filelog.json
@@ -20,7 +20,7 @@
 		{
 			"type": "temporary",
 			"reason": "OOMKilling",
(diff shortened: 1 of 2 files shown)


tosi3k:
Friendly ping for a review request @andyxning @xueweiz :).

When could one expect a new release of NPD (incorporating this fix)?

Random-Liu:
/lgtm
/approve

Thanks for the fix!

PR #488, opened 2020-11-06
Add Detectection logic for  I/O errors

vteratipally:
Add logic for detecting IO errors:

error pattern: Buffer I/O error on dev sdb, logical block 1039, lost async page write


Random-Liu:
/ok-to-test

Can you paste an example error kernel log?

vteratipally:
Buffer I/O error on dev sdb, logical block 1039, lost async page write

PR #489, opened 2020-11-06
Kubelet api server connection check in health checker

abansal4032:
This CL adds a check for kubelet-apiserver connection lost issue([#87615](https://github.com/kubernetes/kubernetes/issues/87615)) in kubelet health checker. If found unhealthy, the health checker will restart kubelet to restore the connection.

Detection of the lost connection is repeated occurrence of "use of closed network connection" in kubelet logs.

The kubelet health check is currently extended by adding code for the specific case but in the long run it makes sense to make it more generic/extensible by allowing to add more to health checks through a well defined interface. That can be improved when we have more use cases for such extensions.

### DIFF
--- a/pkg/healthchecker/health_checker.go
+++ b/pkg/healthchecker/health_checker.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"net/http"
 	"os/exec"
+	"strconv"
 	"strings"
 	"time"
 
@@ -33,7 +34,7 @@ import (
 type healthChecker struct {
 	component       string
 	enableRepair    bool
-	healthCheckFunc func() bool
+	healthCheckFunc func() (bool, error)
 	// The repair is "best-effort" and ignores the error from the underlying actions.
 	// The bash commands to kill the process will fail if the service is down and hence ignore.
 	repairFunc         func()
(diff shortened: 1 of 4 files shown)


pkg/healthchecker/health_checker.go
```diff
@@ -168,3 +162,61 @@ func execCommand(timeout time.Duration, command string, args ...string) (string,
...
+	if err != nil {
+		return true, err
+	}
+	logStartTime := time.Now().Add(-uptime).Format(types.LogParsingTimeLayout)
+	if err != nil {
+		return true, err
+	}
+	out, err := execCommand(types.CmdTimeout, "/bin/sh", "-c",
+		// Query kubelet logs since the logStartTime
+		"journalctl --unit kubelet --since \""+logStartTime+"\" "+
```
Random-Liu, line 192: nit: Use `` to avoid `\"`.

pkg/healthchecker/health_checker.go
```diff
@@ -168,3 +162,61 @@ func execCommand(timeout time.Duration, command string, args ...string) (string,
...
+			"| wc -l")
+	if err != nil {
+		return true, err
+	}
+	occurrences, err := strconv.Atoi(out)
+	if err != nil {
+		return true, err
+	}
+	glog.Infof("log occurrence pattern count : %v", occurrences)
+	return occurrences < 10, nil
```
Random-Liu, line 205: nit: Make 10 a constant.

pkg/healthchecker/health_checker.go
```diff
@@ -168,3 +162,61 @@ func execCommand(timeout time.Duration, command string, args ...string) (string,
...
+// getKubeletHealthCheckFunc returns a function that checks for kubelet health and
+// return false if identified as unhealthy, true otherwise.
+func getKubeletHealthCheckFunc(healthCheckTimeout time.Duration) func() bool {
+	return func() bool {
+		httpHealthy := kubeletHttpHealthCheck(healthCheckTimeout)
+		// If there is any error while checking for health, log and return healthy
+		// because we do not know if the problem is in kubelet or not.
+		connectionHealthy, err := kubeletConnectionHealthCheck()
+		if err != nil {
+			glog.Infof("Error in determining apiserver connection health: %v", err)
```
Random-Liu, line 217: The plugin should return `unknown` when the health detection failed unexpected.

abansal4032: Makes sense. Propagated the error up and eventually the plugin returns unknown when error occurs.

pkg/healthchecker/health_checker.go
abansal4032: Done

pkg/healthchecker/health_checker.go
abansal4032: Done

pkg/healthchecker/health_checker.go
```diff
@@ -168,3 +164,66 @@ func execCommand(timeout time.Duration, command string, args ...string) (string,
...
+	if err != nil {
+		return true, err
+	}
+	logStartTime := time.Now().Add(-uptime).Format(types.LogParsingTimeLayout)
+	if err != nil {
+		return true, err
+	}
+	out, err := execCommand(types.CmdTimeout, "/bin/sh", "-c",
+		// Query kubelet logs since the logStartTime
+		`journalctl --unit kubelet --since "`+logStartTime+
```
krmayankk, line 195: does this assume the presence of this binary ? Is it running this from the container or using hostPath ?

PR #492, opened 2020-11-13
add code to retrieve kernel modules in a linux system from /proc/modules

vteratipally:
Added code to retrieve kernel modules as a first step to add metrics that list all the modules and the number of instances. This metric would be useful in analyzing the usage of different modules.


pkg/util/metrics/system/module_stats.go
```diff
@@ -0,0 +1,85 @@
...
+
+func (d ModuleStat) String() string {
+	s, _ := json.Marshal(d)
+	return string(s)
+}
+
+// Module returns all the kernel modules and their
+// usage. It is read from cat /proc/modules.
+func Modules() ([]ModuleStat, error) {
+	filename := "/proc/modules"
```
karan, line 39: Suggest making this a const

pkg/util/metrics/system/module_stats.go
```diff
@@ -0,0 +1,85 @@
...
+package system
+
+import (
+	"encoding/json"
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+type ModuleStat struct {
```
karan, line 23: Just Module? (this struct isn't really a stat)

pkg/util/metrics/system/module_stats.go
```diff
@@ -0,0 +1,85 @@
...
+		(6)  memory offset for the loaded module.
+		(7)  return a string to represent the kernel taint state. (used here: "P" - Proprietary, "O" - out of tree kernel module, "E" - unsigned module
+	*/
+	for _, line := range lines {
+		fields := strings.Fields(line)
+		moduleName := fields[0] // name of the module
+		numberOfInstances, err :=
+			strconv.ParseUint((fields[1]), 10, 64) // instances of the module are currently loaded
+		if err != nil {
+			return nil, err
```
karan, line 63: Instead of freaking out here, maybe we should log the error but default to "0" so we at least get some metric out.

WDYT?

pkg/util/metrics/system/module_stats.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+
+		var isProprietary = false
```
karan, line 66: You can simplify this whole section by doing:

```go
stats = ModuleStat{
	ModuleName:  moduleName,
	Instances:   numberOfInstances,
}

if len(fields) > 6 {
	stats.Proprietary = strings.Contains(fields[6], "P")
	stats.OutOfTree = strings.Contains(fields[6], "O")
	stats.Unsigned = strings.Contains(fields[6], "E")
}
```

pkg/util/metrics/system/module_stats_test.go
```diff
@@ -0,0 +1,44 @@
...
+package system
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestModules(t *testing.T) {
```
karan, line 23: Please add some cases with a seeded proc file (can use the default from a COS instance, and one from Ubuntu)

pkg/util/metrics/system/module_stats.go
vteratipally: done

pkg/util/metrics/system/module_stats.go
vteratipally: done yes renamed to Module

pkg/util/metrics/system/module_stats.go
vteratipally: make sense, changed to numberOfInstances = 0

pkg/util/metrics/system/module_stats.go
vteratipally: done

pkg/util/metrics/system/module_stats_test.go
vteratipally: added test cases for cos and ubuntu

PR #493, opened 2020-11-15
add code to retrieve kernel command line parameters

vteratipally:
add code to retrieve kernel command line parameters from /proc/cmdline.

it would be retrieved as key and value pair 

example output: 
systemd.unified_cgroup_hierarchy=false
systemd.legacy_systemd_cgroup_controller=false


pkg/util/metrics/system/cmdline_stats.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (d CmdlineStat) String() string {
+	s, _ := json.Marshal(d)
+	return string(s)
+}
+
+// As per the documentation kernel command line parameters can also be specified
+// within double quotes and are separated with spaces.
+// https://www.kernel.org/doc/html/v4.14/admin-guide/kernel-parameters.html#the-kernel-s-command-line-parameters
+var isWithinQuotes = false
```
bsdnet, line 35: s/isWithinQuotes/withinQuotes/g

pkg/util/metrics/system/cmdline_stats.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// Module returns all the kernel modules and their
+// usage. It is read from cat /proc/modules.
+func CmdlineStats() ([]CmdlineStat, error) {
+	filename := "/proc/cmdline"
+	lines, err := ReadFile(filename)
+	if err != nil {
+		return nil, fmt.Errorf("error reading the file %v, %v", filename, err)
+	}
```
bsdnet, line 59: Should we also make sure len(lines) > 1 here

pkg/util/metrics/system/cmdline_stats_test.go
```diff
@@ -0,0 +1,29 @@
+package system
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCmdlineStats(t *testing.T) {
```
bsdnet, line 10: We need to the test the following:
1.) Use COS's default
2.) cmdline with key=value
3) cmdline with key="val val2="
3.) cmdline with single value like "rghb"

pkg/util/metrics/system/common.go
```diff
@@ -0,0 +1,38 @@
...
+
+package system
+
+import (
+	"bufio"
+	"os"
+)
+
+// ReadFile reads contents from a file and returns lines.
+func ReadFile(filename string) ([]string, error) {
```
bsdnet, line 22: maybe ReadFileIntoLines?

also /proc/cmdline only has one line.

pkg/util/metrics/system/common.go
```diff
@@ -0,0 +1,38 @@
...
+	file, err := os.Open(filename)
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+
+	var result []string
+	s := bufio.NewScanner(file)
+	for s.Scan() {
```
karan, line 31: Why not result the scanner and let the caller Scan a line at a time rather than load everything in memory at once?

pkg/util/metrics/system/cmdline_stats.go
```diff
@@ -0,0 +1,83 @@
...
+
+package system
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+)
+
+type CmdlineStat struct {
```
karan, line 22: nit: This isn't really "stat" but more of `CmdlineArgs`

pkg/util/metrics/system/cmdline_stats.go
```diff
@@ -0,0 +1,83 @@
...
+	if isWithinQuotes {
+		return false
+	}
+	return inputChar == ' '
+}
+
+// Module returns all the kernel modules and their
+// usage. It is read from cat /proc/modules.
+func CmdlineStats() ([]CmdlineStat, error) {
+	filename := "/proc/cmdline"
```
karan, line 55: Suggest making this a const above.

pkg/util/metrics/system/cmdline_stats_test.go
karan: +1

vteratipally: added the testcases

pkg/util/metrics/system/cmdline_stats.go
vteratipally: done, replaced

pkg/util/metrics/system/cmdline_stats.go
vteratipally: changed

pkg/util/metrics/system/cmdline_stats.go
vteratipally: changed

pkg/util/metrics/system/common.go
vteratipally: as /proc/* are smaller files, i thought loading everything at once would be fine and faster and reading from /proc/* files is faster. And these are frequent calls keeping the file open till all the lines are processed i am not sure how does npd would react.

pkg/util/metrics/system/common.go
vteratipally: done

pkg/util/metrics/system/cmdline_stats.go
vteratipally: yes

pkg/util/metrics/system/common.go
karan: Since it's a public func, I'm thinking of a case where we'd have dozens of stats reading various proc files. The memory usage can add up (again, not crazy but the performance gains are sufficiently easy to achieve here).

What are your concerns with open files? I can imagine some racy behavior with an open file while the kernel is writing to it (so we may have inconsistent data).

Not a blocker for this PR though.

vteratipally:
/test pull-npd-test
--



/test pull-npd-e2e-test
--

vteratipally:
gentle reminder @Random-Liu

pkg/util/metrics/system/cmdline_args.go
```diff
@@ -0,0 +1,88 @@
...
+func CmdlineArgs() ([]CmdlineArg, error) {
+	lines, err := ReadFileIntoLines(cmdlineFilePath)
+	if err != nil {
+		return nil, fmt.Errorf("error reading the file %v, %v", cmdlineFilePath, err)
+	}
+	if len(lines) < 1 {
+		return nil, fmt.Errorf("no lines are retured")
+	}
+	cmdlineArgs := strings.FieldsFunc(lines[0], splitAfterSpace)
+	var result = make([]CmdlineArg, 0, len(cmdlineArgs))
```
bsdnet, line 64: result := make()

pkg/util/metrics/system/cmdline_args.go
```diff
@@ -0,0 +1,88 @@
...
+	// for commandline only one line is returned.
+	for _, words := range cmdlineArgs {
+		// Ignore the keys that start with double quotes
+		if strings.Index(words, "\"") == 0 {
+			continue
+		}
+		splitsWords := strings.Split(words, "=")
```
bsdnet, line 71: s/splitWords/tokens/g

pkg/util/metrics/system/cmdline_args.go
```diff
@@ -0,0 +1,88 @@
...
+
+func (d CmdlineArg) String() string {
+	s, _ := json.Marshal(d)
+	return string(s)
+}
+
+// As per the documentation kernel command line parameters can also be specified
+// within double quotes and are separated with spaces.
+// https://www.kernel.org/doc/html/v4.14/admin-guide/kernel-parameters.html#the-kernel-s-command-line-parameters
+var withinQuotes = false
```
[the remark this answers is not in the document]
vteratipally: As we need to save the previous state of the run for each character, not sure if we can make it to local variable

pkg/util/metrics/system/cmdline_args.go
vteratipally: make function has the required argument to specify type and size

PR #494, opened 2020-11-16
add new metric to identify boot disk size

vteratipally:
added a new metric boot_disk_size in the disk collector to identify the boot disk size of all the attached boot disks.

the output of the metric would be:

boot_disk_size{device_name="loop0"} 5.36870912e+11
boot_disk_size{device_name="sda"} 2.147483648e+12

### DIFF
--- a/pkg/systemstatsmonitor/disk_collector.go
+++ b/pkg/systemstatsmonitor/disk_collector.go
@@ -19,6 +19,7 @@ package systemstatsmonitor
 import (
 	"context"
 	"os/exec"
+	"strconv"
 	"strings"
 	"time"
 
@@ -38,6 +39,7 @@ type diskCollector struct {
 	mOpsBytes       *metrics.Int64Metric
 	mOpsTime        *metrics.Int64Metric
 	mBytesUsed      *metrics.Int64Metric
+	mBootDiskSize   *metrics.Int64Metric
 
 	config *ssmtypes.DiskStatsConfig
 
@@ -148,6 +150,17 @@ func NewDiskCollectorOrDie(diskConfig *ssmtypes.DiskStatsConfig) *diskCollector 
 		[]string{deviceNameLabel, fsTypeLabel, mountOptionLabel, stateLabel})
 	if err != nil {
 		glog.Fatalf("Error initializing metric for %q: %v", metrics.DiskBytesUsedID, err)
+	}
+
+	dc.mBootDiskSize, err = metrics.NewInt64Metric(
+		metrics.BootDiskSizeID,
+		diskConfig.MetricsConfigs[string(metrics.BootDiskSizeID)].DisplayName,
+		"Boot disk size in Bytes",
+		"Byte",
+		metrics.LastValue,
+		[]string{deviceNameLabel})
+	if err != nil {
+		glog.Fatalf("Error initializing metric for %q: %v", metrics.BootDiskSizeID, err)
 	}
 
 	dc.lastIOTime = make(map[string]uint64)
@@ -290,7 +303,10 @@ func (dc *diskCollector) collect() {
 		dc.mBytesUsed.Record(map[string]string{deviceNameLabel: deviceName, fsTypeLabel: fstype, mountOptionLabel: opttypes, stateLabel: "free"}, int64(usageStat.Free))
 		dc.mBytesUsed.Record(map[string]string{deviceNameLabel: deviceName, fsTypeLabel: fstype, mountOptionLabel: opttypes, stateLabel: "used"}, int64(usageStat.Used))
 	}
-
+	if dc.mBootDiskSize == nil {
+		return
+	}
+	dc.recordBootDiskSize(dc.config.LsblkTimeout)
 }
 
 // listRootBlockDevices lists all block devices that's not a slave or holder.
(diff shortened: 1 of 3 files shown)


pkg/systemstatsmonitor/disk_collector.go
```diff
@@ -300,6 +316,34 @@ func listRootBlockDevices(timeout time.Duration) []string {
 	return strings.Split(strings.TrimSpace(string(stdout)), "\n")
 }
 
+// recordBootDiskSize records the disk block devices that are not a slave or holder.
```
karan, line 319: Insensitive language (unfortunately also in lsblk docs). Suggest leaving it out (here and line 324) or using an alternative (secondary, follower etc)

pkg/systemstatsmonitor/disk_collector.go
```diff
@@ -281,7 +294,10 @@ func (dc *diskCollector) collect() {
 		dc.mBytesUsed.Record(map[string]string{deviceNameLabel: deviceName, fsTypeLabel: fstype, mountOptionLabel: opttypes, stateLabel: "free"}, int64(usageStat.Free))
 		dc.mBytesUsed.Record(map[string]string{deviceNameLabel: deviceName, fsTypeLabel: fstype, mountOptionLabel: opttypes, stateLabel: "used"}, int64(usageStat.Used))
 	}
-
+	if dc.mBootDiskSize == nil {
```
karan, line 297: Why this check? When would it be nil?

pkg/systemstatsmonitor/disk_collector.go
```diff
@@ -300,6 +316,34 @@ func listRootBlockDevices(timeout time.Duration) []string {
...
+	// "-d" prevents printing slave or holder devices. i.e. /dev/sda1, /dev/sda2...
+	// "-b" display the size in bytes.
+	// "-n" prevents printing the headings.
+	// "-p NAME" specifies to only print the device name.
+	cmd := exec.CommandContext(ctx, "lsblk", "-d", "-nb", "-o", "NAME,SIZE")
+	stdout, err := cmd.Output()
+	if err != nil {
+		glog.Errorf("Error calling lsblk")
+	}
+	lines := strings.Split(strings.TrimSpace(string(stdout)), "\n")
```
karan, line 333: suggest using `bufio.NewScanner` which has better memory performance (since we want to keep the NPD resource usage low)

pkg/systemstatsmonitor/disk_collector.go
```diff
@@ -300,6 +316,34 @@ func listRootBlockDevices(timeout time.Duration) []string {
...
+// recordBootDiskSize records the disk block devices that are not a slave or holder.
+func (dc *diskCollector) recordBootDiskSize(timeout time.Duration) {
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	// "-d" prevents printing slave or holder devices. i.e. /dev/sda1, /dev/sda2...
+	// "-b" display the size in bytes.
+	// "-n" prevents printing the headings.
+	// "-p NAME" specifies to only print the device name.
+	cmd := exec.CommandContext(ctx, "lsblk", "-d", "-nb", "-o", "NAME,SIZE")
```
karan, line 328: I was looking into ways that don't fork to get this info:

```
$ time lsblk -d -nb -o NAME,SIZE 2>&1 >/dev/null

real    0m0.046s
user    0m0.025s
sys     0m0.027s

$ time cat /proc/partitions 2>&1 >/dev/null

real    0m0.045s
user    0m0.030s
sys     0m0.021s
```

Looks like they are comparable. Can you read from /proc instead of using lsblk? It may be a bit of work to get the boot disk only though.

vteratipally: optimized the call to just one. hope that should be fine

pkg/systemstatsmonitor/disk_collector.go
vteratipally: not required

pkg/systemstatsmonitor/disk_collector.go
vteratipally: updated the comment

karan:
/lgtm 

(the commend about using buffers is optional and non-blocker)

pkg/systemstatsmonitor/disk_collector.go
```diff
@@ -290,7 +303,10 @@ func (dc *diskCollector) collect() {
 		dc.mBytesUsed.Record(map[string]string{deviceNameLabel: deviceName, fsTypeLabel: fstype, mountOptionLabel: opttypes, stateLabel: "free"}, int64(usageStat.Free))
 		dc.mBytesUsed.Record(map[string]string{deviceNameLabel: deviceName, fsTypeLabel: fstype, mountOptionLabel: opttypes, stateLabel: "used"}, int64(usageStat.Used))
 	}
-
+	if dc.mBootDiskSize == nil {
```
ForestCold: (nit) suggest to write it like:

if dc.mBootDiskSize != nil {
 dc.recordBootDiskSize(dc.config.LsblkTimeout)
}

Then it will not prevent you from adding more metrics in future.

PR #503, opened 2020-12-08
changing the label names as per the standards

vteratipally:
renaming the label names as per the previous standard and correcting the comments based on the change


vteratipally:
Fixes: https://github.com/kubernetes/node-problem-detector/issues/471

Issue #461, opened 2020-09-08
Feature Request: Windows Support


jeremyje:
Since there's growing support for Windows in Kubernetes and Docker community it'd be nice to get a Windows build for the node-problem-detector. This has been discussed on [sig-windows](https://docs.google.com/document/d/1Tjxzjjuy4SQsFSUVXZbvqVb64hjNAG5CQX8bK7Yda9w/edit#heading=h.itpry7uee9tq) before. I'd like to start a discussion here on what's necessary.

I've been looking on what's necessary to bring multi-arch building to this project and what dependency detanglement is necessary to build Windows binaries. It's possible with a few patches and dependency upgrades to build Windows binaries.

Multi-arch building would also unlock ARM support requested from https://github.com/kubernetes/node-problem-detector/issues/135.

Design Document: https://docs.google.com/document/d/1eiK6KAp_TFR0PgBMu2WCf49fMZcg-HHnBHMc9fALquU/edit

jeremyje:
Design Doc: https://docs.google.com/document/d/1eiK6KAp_TFR0PgBMu2WCf49fMZcg-HHnBHMc9fALquU

SergeyKanzhelev:
/reopen

need to update the readme: https://github.com/kubernetes/node-problem-detector#windows

allyford:
@jeremyje - Is this work planned?

SergeyKanzhelev:
/reopen

need to update the readme: https://github.com/kubernetes/node-problem-detector#windows

PR #507, opened 2020-12-18
Scope ENABLE_JOURNALD make variable to Linux-only binaries, go build package instead of .go file.

jeremyje:
`ENABLE_JOURNALD` make argument is scoped to Linux binaries, Windows will ignore it since it's not supported.
`go build` targets the package directory instead of the specific go file so that multi-OS implementations can be built. Generally OS specific `func main()` must live in different .go files.

Issue: https://github.com/kubernetes/node-problem-detector/issues/461

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,9 +64,11 @@ CGO_ENABLED:=0
 
 # Construct the "-tags" parameter used by "go build".
 BUILD_TAGS?=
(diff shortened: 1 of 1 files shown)


jeremyje:
https://github.com/kubernetes/node-problem-detector/pull/509 obsoletes this PR.

PR #517, opened 2021-01-29
Windows Support: Fix Build Regressions, Tests Pass

jeremyje:
This change fixes a build regression on Windows where some Linux specific logic in syscall package was used.
This change also adds the new Linux and Windows binaries to the critical path of of the test-infra CI. The new locations are more friendly to multi-arch builds.


pkg/custompluginmonitor/plugin/plugin_windows.go
```diff
@@ -0,0 +1,29 @@
...
+limitations under the License.
+*/
+
+package plugin
+
+import (
+	"os/exec"
+)
+
+func execCommandInGroup(name string, arg ...string) *exec.Cmd {
```
Random-Liu, line 23: Does windows have the child process issue?

Makefile
```diff
@@ -197,7 +198,7 @@ e2e-test: vet fmt build-tar
 	-boskos-project-type=$(BOSKOS_PROJECT_TYPE) -job-name=$(JOB_NAME) \
 	-artifacts-dir=$(ARTIFACTS)
 
-build-binaries: ./bin/node-problem-detector ./bin/log-counter ./bin/health-checker
+build-binaries: ./bin/node-problem-detector ./bin/log-counter ./bin/health-checker $(WINDOWS_BINARIES) $(LINUX_BINARIES)
```
Random-Liu, line 201: `./bin/node-problem-detector ./bin/log-counter ./bin/health-checker` can be removed?

jeremyje: Eventually. There needs to be some updates through the upstream CI scripts to manage this.

pkg/custompluginmonitor/plugin/plugin_windows.go
Random-Liu: See https://stackoverflow.com/questions/22470193/why-wont-go-kill-a-child-process-correctly.

People mentioned a windows solution.

pkg/custompluginmonitor/plugin/plugin.go
```diff
@@ -234,7 +238,11 @@ func (p *Plugin) run(rule cpmtypes.CustomRule) (exitStatus cpmtypes.Status, outp
 	output = string(stdout)
 	output = strings.TrimSpace(output)
 
-	if cmd.ProcessState.Sys().(syscall.WaitStatus).Signaled() {
+	m.Lock()
+	wasKilled := killed
+	m.Unlock()
+
+	if cmd.ProcessState.Sys().(syscall.WaitStatus).Signaled() || wasKilled {
```
Random-Liu, line 245: Is `cmd.ProcessState.Sys().(syscall.WaitStatus).ExitStatus()` below supported on windows?

If the customer plugin is not supported on windows yet, should we avoid building the whole custom plugin part on windows for now?

Random-Liu: Just build with `disable_custom_plugin_monitor`.

jeremyje: I'd like to keep it for now since it can work. Also this logic works fine.
The ExitStatus does not work. It's pinned to false. Once you get into syscall territory you will have to confirm each OS. 

There's a potential use for this logic see the design document for the use cases.

jeremyje:
Ping, we cannot build Windows NPD yet until this goes in.

jeremyje:
Weekly ping, Windows NPD is blocked on this change.

pkg/custompluginmonitor/plugin/plugin.go
```diff
@@ -183,7 +182,12 @@ func (p *Plugin) run(rule cpmtypes.CustomRule) (exitStatus cpmtypes.Status, outp
 				glog.Errorf("Error in cmd.Process check %q", rule.Path)
 				break
 			}
-			err := syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL)
+
+			m.Lock()
+			killed = true
+			m.Unlock()
```
Random-Liu, line 188: Change `killed` to `timeout`?

It is a bit weird to mark `killed` before the actual kill.

pkg/custompluginmonitor/plugin/plugin.go
Random-Liu: If this is the case, maybe just always use the `timeout` variable? That makes the logic more OS independent.

pkg/util/exec_windows.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+// Exec creates a new process with the specified arguments.
+func Exec(name string, arg ...string) *exec.Cmd {
+	// Windows does not handle relative path names in exec very well.
+	name = filepath.Clean(name)
+	cmdArgs := arg
+
+	// Unlike Linux Windows not create a shell context
+	if strings.ToLower(filepath.Ext(name)) == ".cmd" {
```
Random-Liu, line 36: Can you elaborate a bit more here?

Random-Liu (commented):
LGTM with nits and a question.

pkg/util/exec_windows.go
jeremyje: I've added more information. Also added powershell support because that's generally preferred as it provides more flexibility than cmd.exe.

jeremyje:
Thanks for the review. Answered questions in the code changes itself.

pkg/custompluginmonitor/plugin/plugin.go
```diff
@@ -234,7 +238,11 @@ func (p *Plugin) run(rule cpmtypes.CustomRule) (exitStatus cpmtypes.Status, outp
 	output = string(stdout)
 	output = strings.TrimSpace(output)
 
-	if cmd.ProcessState.Sys().(syscall.WaitStatus).Signaled() {
+	m.Lock()
+	cmdKilled := timeout
```
Random-Liu, line 242: Actually it is possible that the command exit itself after we mark time, and the kill didn't actually work. That is why I prefer `timeout` than `killed`.

Let's just keep using `timeout`.

Random-Liu (commented):
LGTM

pkg/custompluginmonitor/plugin/plugin.go
Random-Liu: This is not as important, we can always change it later if needed.

Issue #403, opened 2019-12-12
flake in e2e test: Got value for metric problem_counter with label map[reason:Ext4Error]: 3, expect at most 2.


xueweiz:
See this [test result](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/node-problem-detector/385/ci-npd-e2e-test/1204905639302664194):

```
• Failure [52.528 seconds]
NPD should export Prometheus metrics.
/home/prow/go/src/k8s.io/node-problem-detector/test/e2e/metriconly/metrics_test.go:36
  When ext4 filesystem error happens
  /home/prow/go/src/k8s.io/node-problem-detector/test/e2e/metriconly/metrics_test.go:104
    NPD should update problem_counter{reason:Ext4Error} and problem_gauge{type:ReadonlyFilesystem} [It]
    /home/prow/go/src/k8s.io/node-problem-detector/test/e2e/metriconly/metrics_test.go:113
    Got value for metric problem_counter with label map[reason:Ext4Error]: 3, expect at most 2.
      
    Expected
        <float64>: 3
    to be <=
        <float64>: 2
    /home/prow/go/src/k8s.io/node-problem-detector/test/e2e/metriconly/metrics_test.go:190
```

Go to the [kernel log test artifact](https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/node-problem-detector/385/ci-npd-e2e-test/1204905639302664194/artifacts/NPD_should_export_Prometheus_metrics._When_ext4_filesystem_error_happens_NPD_should_update_problem_counter%7breason:Ext4Error%7d_and_problem_gauge%7btype:ReadonlyFilesystem%7d/), we see:
```
Dec 11 23:29:38 npd-metrics-cos-73-11647-348-0-0d055070 kernel: Bridge firewalling registered
Dec 11 23:29:38 npd-metrics-cos-73-11647-348-0-0d055070 kernel: IPv6: ADDRCONF(NETDEV_UP): docker0: link is not ready
Dec 11 23:30:02 npd-metrics-cos-73-11647-348-0-0d055070 kernel: EXT4-fs (sda1): re-mounted. Opts: commit=30,data=ordered
Dec 11 23:30:03 npd-metrics-cos-73-11647-348-0-0d055070 kernel: EXT4-fs error (device sda1): trigger_test_error:124: comm problem-maker: fake filesystem error from problem-maker
[...]

PR #529, opened 2021-02-02
Allow building on macOS

stmcginnis:
Journald is not available on mac. To allow building the rest of the
project while working on a mac, use the same flag as the Windows build
to skip inclusion of journald.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,7 +53,9 @@ IMAGE:=$(REGISTRY)/node-problem-detector:$(TAG)
 # support needs libsystemd-dev or libsystemd-journal-dev.
 ENABLE_JOURNALD?=1
 
-ifeq ($(shell uname -s 2>/dev/null), Darwin)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -53,6 +53,10 @@ IMAGE:=$(REGISTRY)/node-problem-detector:$(TAG)
 # support needs libsystemd-dev or libsystemd-journal-dev.
 ENABLE_JOURNALD?=1
 
+ifeq ($(shell uname -s 2>/dev/null), Darwin)
```
Random-Liu, line 56: Can we use `go env` instead and also add windows here?

stmcginnis: Great idea, I'll update shortly.

stmcginnis:
/retest

Looks like this may be the issue: https://github.com/kubernetes/node-problem-detector/issues/403

gengwg:
Is there instructions how to build the image on macos? the README is for debian.

PR #536, opened 2021-02-17
Make log pattern check configurable in health checker

abansal4032:
This CL makes the log pattern and the threshold configurable through health checker plugin config through flags --log-pattern and --log-pattern-count-threshold respectively. An empty --log-pattern(default) will skip the log pattern check and not mark the service as unhealthy.

### DIFF
--- a/cmd/healthchecker/options/options.go
+++ b/cmd/healthchecker/options/options.go
@@ -40,6 +40,7 @@ type HealthCheckerOptions struct {
 	CriSocketPath      string
 	CoolDownTime       time.Duration
 	HealthCheckTimeout time.Duration
+	LogPatterns        types.LogPatternFlag
 }
 
 // AddFlags adds health checker command line options to pflag.
@@ -57,6 +58,8 @@ func (hco *HealthCheckerOptions) AddFlags(fs *pflag.FlagSet) {
 		"The duration to wait for the service to be up before attempting repair.")
 	fs.DurationVar(&hco.HealthCheckTimeout, "health-check-timeout", types.DefaultHealthCheckTimeout,
 		"The time to wait before marking the component as unhealthy.")
+	fs.Var(&hco.LogPatterns, "log-pattern",
+		"The log pattern to look for in service journald logs. The format for flag value <failureThresholdCount>:<logPattern>")
 }
 
 // IsValid validates health checker command line options.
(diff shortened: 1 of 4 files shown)


cmd/healthchecker/options/options.go
```diff
@@ -33,13 +33,15 @@ func NewHealthCheckerOptions() *HealthCheckerOptions {
...
-	CoolDownTime       time.Duration
-	HealthCheckTimeout time.Duration
+	Component                string
+	SystemdService           string
+	EnableRepair             bool
+	CriCtlPath               string
+	CriSocketPath            string
+	CoolDownTime             time.Duration
+	HealthCheckTimeout       time.Duration
+	LogPattern               string
```
Random-Liu, line 43: We should support multiple log patterns. Can we have a customized flag supporting this, e.g. https://play.golang.org/p/Ig5sm7jA14?

It could be something like this:
```
health-checker --log-pattern 10:aaaaaa --log-pattern 20:bbbbbb --log-pattern 15:ccccc
```

abansal4032: I see. Yeah that can be helpful.

Is the idea in the above example <threshold count>:<pattern>?

Random-Liu: Yeah. If we do pattern:threshold_count, it is possible that the pattern may contain `:`, and make the parsing harder.

abansal4032: Make sense. Changed.

pkg/healthchecker/types/types.go
```diff
@@ -33,11 +38,60 @@ const (
...
+	LogPatternFlagSeparator = ":"
 )
 
 type HealthChecker interface {
 	CheckHealth() (bool, error)
 }
+
+// LogPatternFlag defines the flag for log pattern health check.
+// It contains a map of <log pattern> to <failure threshold for the pattern>
+type LogPatternFlag struct {
```
Random-Liu, line 52: Add a unit test for this.

Random-Liu:
LGTM, but please add a unit test.

pkg/healthchecker/types/types.go
abansal4032: Added unit tests.

Issue #482, opened 2020-10-26
docker healthcheck - systemctl unable to get information


azman0101:
When enabling health-checker for docker. The Unhealthy condition is detected when, in fact docker is healthy

```bash
I1026 18:25:35.879338       1 plugin.go:239] Start logs from plugin {Type:permanent Condition:ContainerRuntimeUnhealthy Reason:DockerUnhealthy Path:/home/kubernetes/bin/health-checker Args:[--component=docker --enable-repair=true --cooldown-time=2m --health-check-timeout=60s] TimeoutString:0xc000284ed0 Timeout:3m0s}
 I1026 18:25:35.876378      46 health_checker.go:166] command &{docker [docker ps] []  <nil>  0xc0000a02d0 [] <nil> <nil> <nil> 0xc00009c300 0xc0000a83c0 false [] [] [] [] <nil> <nil>} failed: exec: "docker": executable file not found in $PATH, []
I1026 18:25:35.878837      46 health_checker.go:166] command &{/bin/systemctl [systemctl show docker --property=InactiveExitTimestamp] []  <nil>  0xc0000a03c0 [] <nil> 0xc000094c90 exit status 1 0xc00009c420 <nil> true [0xc00009a058 0xc00009a070 0xc00009a088] [0xc00009a058 0xc00009a070 0xc00009a088] [0xc00009a068 0xc00009a080] [0x61a150 0x61a150] 0xc00009c6c0 0xc000082180} failed: exit status 1, []
I1026 18:25:35.878866      46 health_checker.go:147] error in getting uptime for docker: exit status 1
I1026 18:25:35.878875      46 health_checker.go:149] docker is unhealthy, component uptime: 0s
I1026 18:25:35.879398       1 plugin.go:240] End logs from plugin {Type:permanent Condition:ContainerRuntimeUnhealthy Reason:DockerUnhealthy Path:/home/kubernetes/bin/health-checker Args:[--component=docker --enable-repair=true --cooldown-time=2m --health-check-timeout=60s] TimeoutString:0xc000284ed0 Timeout:3m0s}
```

I tested the command `systemctl show docker --property=InactiveExitTimestamp` inside a npd container:

```bash
kubectl exec -it -n kube-system node-problem-detector-8z9lj -- /bin/sh

# systemctl show docker --property=InactiveExitTimestamp
System has not been booted with systemd as init system (PID 1). Can't operate.
```

To dig into this issue, I've added hostPID to pod spec.

```bash
[...]

Issue #439, opened 2020-07-03
health-checker posting wrong status for kubelet - `KubeletUnhealthy`


OkBeacon:
I am trying to use the `health-check-monitor` to monitor kubelet and docker. I built the binaries from the source and created a docker image.

Although the pod is running fine, it consistently posts -
```
Conditions:
  Type                 Status  LastHeartbeatTime                 LastTransitionTime                Reason                       Message
  ----                 ------  -----------------                 ------------------                ------                       -------
  KubeletUnhealthy     True    Fri, 03 Jul 2020 17:04:55 +0100   Fri, 03 Jul 2020 16:54:53 +0100   KubeletUnhealthy             kubelet:kubelet was found unhealthy; repair flag : false
```

Pod logs:
```
I0703 16:16:01.211066       1 plugin.go:86] Start to run custom plugins
I0703 16:16:01.222034       1 plugin.go:110] Add check result {Rule:0xc00003b1f0 ExitStatus:1 Message:kubelet:kubelet was found unhealthy; repair flag : false} for rule &{Type:permanent Condition:KubeletUnhealthy Reason:KubeletUnhealthy Path:/home/kubernetes/bin/health-checker Args:[--component=kubelet --enable-repair=false --cooldown-time=1m --health-check-timeout=10s] TimeoutString:0xc0000a8a50 Timeout:3m0s}
I0703 16:16:01.222104       1 plugin.go:115] Finish running custom plugins
I0703 16:16:01.222139       1 custom_plugin_monitor.go:138] New status generated: &{Source:health-checker Events:[] Conditions:[{Type:KubeletUnhealthy Status:True Transition:2020-07-03 15:54:31.225162591 +0000 UTC m=+0.065349002 Reason:KubeletUnhealthy Message:kubelet:kubelet was found unhealthy; repair flag : false}]}
```

I `exec` to the pod:
```
$ kubectl exec -it node-problem-detector-kl9kj -- /bin/sh
# /home/kubernetes/bin/health-checker
I0703 16:17:36.720049    1019 health_checker.go:136] health-checker: component is unhealthy, proceeding to repair
I0703 16:17:36.720251    1019 health_checker.go:156] health-checker: executing command : &{systemctl [systemctl show kubelet --property=ActiveEnterTimestamp] []  <nil> <nil> <nil> [] <nil> <nil> <nil> 0xc00008c840 0xc000096460 false [] [] [] [] <nil> <nil>}
[...]

PR #539, opened 2021-03-10
improvement health-checker

smileusd:
/kind bug
1. add loopbacktime to reduce time of journalctl call
If use start time of kubelet as the journalctl, it could be take too much time to call. 
I add loopkbacktime to set loop back time for journalctl, if not set or set = 0, it uses start time directly, else it use loopkbacktime.

2. fix timelayout for all timezones
Different nodes usally use different time zone:
`[root@kubernetes-master-505-pkshx-4589367 software]# systemctl show kubelet --property=InactiveExitTimestamp
InactiveExitTimestamp=Wed 2021-03-10 06:05:51 PST`
I think npd doesn't need to care about the timezone if can ensure timezone is agreement with runtime/kubelet and journald.
----
this point has been fixed by https://github.com/kubernetes/node-problem-detector/pull/558

3. add npd health-checker deploy file
Many users reflect that can not use health-checker because use the wrong specs: [!439](https://github.com/kubernetes/node-problem-detector/issues/439), [!540](https://github.com/kubernetes/node-problem-detector/issues/540), [!482](https://github.com/kubernetes/node-problem-detector/issues/482). I thinks this commit can fix these issues.

### DIFF
--- a/config/health-checker-kubelet.json
+++ b/config/health-checker-kubelet.json
@@ -25,6 +25,7 @@
         "--component=kubelet",
         "--enable-repair=true",
         "--cooldown-time=1m",
+        "--loopback-time=0",
         "--health-check-timeout=10s"
       ],
       "timeout": "3m"
--- a/deployment/node-problem-detector-healthchecker.yaml
+++ b/deployment/node-problem-detector-healthchecker.yaml
@@ -1,0 +1,102 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: node-problem-detector
+  namespace: kube-system
+  labels:
+    app: node-problem-detector
+spec:
+  selector:
+    matchLabels:
+      app: node-problem-detector
+  template:
+    metadata:
+      labels:
+        app: node-problem-detector
+    spec:
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+              - matchExpressions:
+                  - key: kubernetes.io/os
+                    operator: In
+                    values:
+                      - linux
+      containers:
+      - name: node-problem-detector
+        command:
+        - /node-problem-detector
+        - --logtostderr
+        - --config.system-log-monitor=/config/kernel-monitor.json,/config/docker-monitor.json
+        - --config.custom-plugin-monitor=/config/health-checker-kubelet.json
+        image: k8s.gcr.io/node-problem-detector/node-problem-detector:v0.8.6
+        resources:
+          limits:
+            cpu: 10m
+            memory: 80Mi
+          requests:
+            cpu: 10m
+            memory: 80Mi
+        imagePullPolicy: Always
+        securityContext:
+          privileged: true
+        env:
+        - name: NODE_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        volumeMounts:
+        - name: log
+          mountPath: /var/log
+          readOnly: true
+        - name: kmsg
+          mountPath: /dev/kmsg
+          readOnly: true
+        # Make sure node problem detector is in the same timezone
+        # with the host.
+        - name: localtime
+          mountPath: /etc/localtime
+          readOnly: true
+        - name: config
+          mountPath: /config
+          readOnly: true
+        - mountPath: /etc/machine-id
+          name: machine-id
+          readOnly: true
+        - mountPath: /run/systemd/system
+          name: systemd
+        - mountPath: /var/run/dbus/
+          name: dbus
+          mountPropagation: Bidirectional
+      volumes:
+      - name: log
+        # Config `log` to your system log directory
+        hostPath:
+          path: /var/log/
+      - name: kmsg
+        hostPath:
+          path: /dev/kmsg
+      - name: localtime
+        hostPath:
+          path: /etc/localtime
+      - name: config
+        configMap:
+          name: node-problem-detector-config
+          items:
+          - key: kernel-monitor.json
+            path: kernel-monitor.json
+          - key: docker-monitor.json
+            path: docker-monitor.json
+      - name: machine-id
+        hostPath:
+          path: /etc/machine-id
+          type: "File"
+      - name: systemd
+        hostPath:
+          path: /run/systemd/system/
+          type: ""
+      - name: dbus
+        hostPath:
+          path: /var/run/dbus/
+          type: ""
--- a/pkg/healthchecker/health_checker.go
+++ b/pkg/healthchecker/health_checker.go
@@ -36,6 +36,7 @@ type healthChecker struct {
 	crictlPath         string
 	healthCheckTimeout time.Duration
 	coolDownTime       time.Duration
+	loopBackTime       time.Duration
 	logPatternsToCheck map[string]int
 }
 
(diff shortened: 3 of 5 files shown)


smileusd:
@Random-Liu PTAL

Random-Liu:
@abansal4032 Hey archit, can you take a look at this change?

config/health-checker-kubelet.json
```diff
@@ -25,6 +25,7 @@
         "--component=kubelet",
         "--enable-repair=true",
         "--cooldown-time=1m",
+        "--loopback-time=5m",
```
abansal4032, line 28: Maybe let the default config use the default loopback time to keep the same behavior as it was before this change?

pkg/healthchecker/health_checker.go
```diff
@@ -139,11 +142,21 @@ func getHealthCheckFunc(hco *options.HealthCheckerOptions) func() (bool, error)
...
 func (hc *healthChecker) CheckHealth() (bool, error) {
+	var logStartTime string
 	healthy, err := hc.healthCheckFunc()
 	if err != nil {
 		return healthy, err
 	}
-	logPatternHealthy, err := logPatternHealthCheck(hc.systemdService, hc.logPatternsToCheck)
+	uptime, err := hc.uptimeFunc()
+	if err != nil {
+		return false, err
```
abansal4032, line 152: This will return the service as unhealthy in case of parsing errors in uptime. Might lead to unnecessary restarts of the service. We can log the error here and return true as we are not sure if the service was unhealthy at this point.

deployment/node-problem-detector-healthchecker.yaml
```diff
@@ -0,0 +1,102 @@
...
+          mountPath: /etc/localtime
+          readOnly: true
+        - name: config
+          mountPath: /config
+          readOnly: true
+        - mountPath: /etc/machine-id
+          name: machine-id
+          readOnly: true
+        - mountPath: /run/systemd/system
+          name: systemd
```
abansal4032, line 68: Were you able to run the deployment and get the health signals without errors using this configuration?

abansal4032 (changes requested):

deployment/node-problem-detector-healthchecker.yaml
smileusd: Yes, this deployment has been ran in our productions.

config/health-checker-kubelet.json
smileusd: Default loopback time is zero, so if not this set. It is the same behavior as it was before. Here is a example to ues.

pkg/healthchecker/health_checker.go
smileusd: done

config/health-checker-kubelet.json
smileusd: I set to 0.

abansal4032: Thanks!

pkg/healthchecker/health_checker.go
```diff
@@ -82,7 +82,8 @@ func getUptimeFunc(service string) func() (time.Duration, error) {
 		if len(val) < 2 {
 			return time.Duration(0), errors.New("could not parse the service uptime time correctly")
 		}
-		t, err := time.Parse(types.UptimeTimeLayout, val[1])
+		timeEndIndex := strings.LastIndex(val[1], " ")
+		t, err := time.Parse(types.UptimeTimeLayout, val[1][:timeEndIndex])
```
Random-Liu, line 86: IIUC, if you run NPD in a container, it may not be in the same time zone with the host, in that case would this cause any problem?

What is the problem if we don't make this change? I don't quite get it after reading the PR description.

smileusd: The resoult for the cmd out is like this:
```$ systemctl show kubelet --property=InactiveExitTimestamp
InactiveExitTimestamp=Wed 2021-04-14 01:11:39 -07
```
But types.UptimeTimeLayout is `UptimeTimeLayout          = "Mon 2006-01-02 15:04:05"`
It is failed to parse time

Random-Liu: The `UptimeTimeLayout` was `"Mon 2006-01-02 15:04:05 UTC"` before your change.

And in your PR description, the systemctl output was:
```
[root@kubernetes-master-505-pkshx-4589367 software]# systemctl show kubelet --property=InactiveExitTimestamp InactiveExitTimestamp=Wed 2021-03-10 06:05:51 PST
```

And on my node it is also:
```
$ systemctl show docker --property=InactiveExitTimestamp
InactiveExitTimestamp=Thu 2021-04-15 17:20:37 PDT
```

Why would the parsing fail?

smileusd: I rollback to old version and retry on my node, it is still wrong and reture error: 
```
I0425 22:54:26.604844   18356 plugin.go:87] Add check result {Rule:0xc4200bd9d0 ExitStatus:2 Message:systemctl show: Sun 2021-04-25 11:22:50 PDT
error checking kubelet health: pars}
```
Then I code a test case like this:
```package main

import (
    "fmt"
    "time"
)

const UptimeTimeLayout = "Mon 2006-01-02 15:04:05 UTC"

func main() {
    strs := []string{
        "Wed 2021-03-10 03:39:13 PST",
        "Sun 2021-04-25 11:22:50 PDT",
    }

    t, err := time.Parse(UptimeTimeLayout, strs[0])
    if err != nil {
        fmt.Println("error: ", err)
    }
    t, err = time.Parse(UptimeTimeLayout, strs[0])
    if err != nil {
        fmt.Println("error: ", err)
    }
    fmt.Println(t.String())
}
```
It is also wrong:
```error:  parsing time "Wed 2021-03-10 03:39:13 PST" as "Mon 2006-01-02 15:04:05 UTC": cannot parse "PST" as " UTC"
error:  parsing time "Wed 2021-03-10 03:39:13 PST" as "Mon 2006-01-02 15:04:05 UTC": cannot parse "PST" as " UTC"
0001-01-01 00:00:00 +0000 UTC
Exiting.
```

Random-Liu: I see. In that case, we probably need a better fix.

Random-Liu:
```
package main
import (
    "fmt"
    "time"
)

const UptimeTimeLayout = "Mon 2006-01-02 15:04:05 MST"

func main() {
    strs := []string{
        "Wed 2021-03-10 03:39:13 PST",
        "Sun 2021-04-25 11:22:50 PDT",
    }

    t, err := time.Parse(UptimeTimeLayout, strs[0])
    if err != nil {
        fmt.Println("error: ", err)
    }
    fmt.Println(t.String())
    t, err = time.Parse(UptimeTimeLayout, strs[1])
    if err != nil {
        fmt.Println("error: ", err)
    }
    fmt.Println(t.String())
}
```

Can you change it to MST?
> When parsing a time with a zone abbreviation like MST, if the zone abbreviation has a defined offset in the current location, then that offset is used.

Random-Liu:
Hi @smileusd Can you rebase your change and fix the timestamp format?

We plan to cut a release this week. :)

smileusd:
> Hi @smileusd Can you rebase your change and fix the timestamp format?
> 
> We plan to cut a release this week. :)

Sorry to late see this. 
Done

smileusd:
@Random-Liu @abansal4032 PTAL

smileusd:
@Random-Liu Any updates?

Random-Liu (commented):
/lgtm
/approve

PR #541, opened 2021-03-16
Enable NPD to run as a Windows Service.

jeremyje:
The change allows NPD to run as a Windows Service. This is similar to a systemd service on Linux but Windows services allow for control states (pause, continue, etc).
This implementation only allows for running and stopped states since there's no benefit to add support for paused.
To run NPD as a Windows Service you must use the service control, `sc.exe`, to install it, instructions added to the readme. NPD will not support the notion of `--register-service` given how it processes flags and the complexity of making this work does is not worth it.

Lastly, since this is a significant change to the main loop for NPD, unit tests have been added to catch future regressions.

See https://github.com/kubernetes/node-problem-detector/issues/461 for Windows Support and design.


dchen1107:
/assign @chardch 

Richard, can you help review this PR? Please let me and Lantao know if you need more help here. Thanks!

cmd/nodeproblemdetector/node_problem_detector_windows.go
```diff
@@ -0,0 +1,123 @@
...
+	for {
+		select {
+		case c := <-r:
+			switch c.Cmd {
+			case svc.Interrogate:
+				changes <- c.CurrentStatus
+				// Testing deadlock from https://code.google.com/p/winsvc/issues/detail?id=4
+				time.Sleep(100 * time.Millisecond)
+				changes <- c.CurrentStatus
+			case svc.Stop, svc.Shutdown:
```
liyanhui1228, line 110: Do we need to set the svc status in this case? Something like `changes <- svc.Status{State: svc.StopPending, Accepts: 0}`

cmd/nodeproblemdetector/node_problem_detector_windows.go
```diff
@@ -0,0 +1,123 @@
...
+	handler := &npdService{
+		options: npdo,
+	}
+
+	run := debug.Run
+	if runningAsService {
+		run = svc.Run
+	}
+
+	if err := run(svcName, handler); err != nil {
```
liyanhui1228, line 64: Maybe setup windows event logging before starting the service?

cmd/nodeproblemdetector/node_problem_detector_windows.go
```diff
@@ -0,0 +1,123 @@
...
+	changes <- svc.Status{State: svc.Stopped, Accepts: svcCommandsAccepted}
+
+	return
+}
+
+func serviceLoop(r <-chan svc.ChangeRequest, changes chan<- svc.Status, termCh chan error) {
+	for {
+		select {
+		case c := <-r:
+			switch c.Cmd {
```
liyanhui1228, line 104: Curious how is service restart handled here?

pkg/problemdetector/problem_detector.go
```diff
@@ -53,21 +53,32 @@ func (p *problemDetector) Run() error {
...
 		}
 	}
-	if len(p.monitors) == failureCount {
+	allMonitors := p.monitors
+
+	if len(allMonitors) == failureCount {
 		return fmt.Errorf("no problem daemon is successfully setup")
 	}
+
+	defer func() {
```
chardch, line 69: Just wondering, why is added. Was it not needed before?

cmd/nodeproblemdetector/node_problem_detector_windows.go
```diff
@@ -0,0 +1,123 @@
...
+
+type npdService struct {
+	sync.Mutex
+	options *options.NodeProblemDetectorOptions
+}
+
+func (s *npdService) Execute(args []string, r <-chan svc.ChangeRequest, changes chan<- svc.Status) (ssec bool, errno uint32) {
+	termCh := make(chan error, 1)
+	changes <- svc.Status{State: svc.StartPending}
+	changes <- svc.Status{State: svc.Running, Accepts: svcCommandsAccepted}
```
chardch, line 77: Perhaps obvious, but I'm not very familiar with the details of how NPD works yet, but why are these two states `StartPending` and `Running` not considered in `serviceLoop` below?

cmd/nodeproblemdetector/node_problem_detector_windows.go
```diff
@@ -0,0 +1,123 @@
...
+				changes <- c.CurrentStatus
+				// Testing deadlock from https://code.google.com/p/winsvc/issues/detail?id=4
+				time.Sleep(100 * time.Millisecond)
+				changes <- c.CurrentStatus
+			case svc.Stop, svc.Shutdown:
+				glog.Infof("Stopping %s service, %v", svcName, c.Context)
+				termCh <- errors.New("stopping service")
+				return
+			case svc.Pause:
+				changes <- svc.Status{State: svc.Paused, Accepts: svcCommandsAccepted}
```
chardch, line 115: Do we also want to log out pausing and continuing the service, like in the stop/shutdown case?

chardch (commented):
Mostly looks good to me. I left a couple clarifying questions

pkg/problemdetector/problem_detector.go
jeremyje: It didn't appear that NPD ever stopped the individual problem detectors. I added it because while trying to implement pause/continue it made sense to add. Although pause/continue was abandoned because it introduced more complexity and was not necessary. Functionally this logic shouldn't disrupt the critical path for NPD because it'll only run on shutdown of the service.

cmd/nodeproblemdetector/node_problem_detector_windows.go
jeremyje: Windows stops the process and then a new process is created. There's no explicit concept of "restart" in Windows.
See: https://docs.microsoft.com/en-us/dotnet/framework/windows-services/introduction-to-windows-service-applications#service-lifetime

> A running service can exist in this state indefinitely until it is either stopped or paused or until the computer shuts down. A service can exist in one of three basic states: Running, Paused, or Stopped. The service can also report the state of a pending command: ContinuePending, PausePending, StartPending, or StopPending. These statuses indicate that a command has been issued, such as a command to pause a running service, but has not been carried out yet. You can query the Status to determine what state a service is in, or use the WaitForStatus to carry out an action when any of these states occurs.

cmd/nodeproblemdetector/node_problem_detector_windows.go
jeremyje: That shouldn't be necessary. The shutdown generally happens quickly if Windows kills the application that's fine to.

See: https://docs.microsoft.com/en-us/dotnet/framework/windows-services/introduction-to-windows-service-applications#service-lifetime

> A running service can exist in this state indefinitely until it is either stopped or paused or until the computer shuts down. A service can exist in one of three basic states: Running, Paused, or Stopped. The service can also report the state of a pending command: ContinuePending, PausePending, StartPending, or StopPending. These statuses indicate that a command has been issued, such as a command to pause a running service, but has not been carried out yet. You can query the Status to determine what state a service is in, or use the WaitForStatus to carry out an action when any of these states occurs.

cmd/nodeproblemdetector/node_problem_detector_windows.go
chardch: Got it, thanks for clarifying

pkg/problemdetector/problem_detector.go
chardch: Thanks for the info

chardch (approved):

jeremyje:
Thanks Richard for the review!

Ping others for review.

cmd/nodeproblemdetector/node_problem_detector_windows.go
liyanhui1228: Thanks for the info!

cmd/nodeproblemdetector/node_problem_detector_windows.go
liyanhui1228: Got it, thanks!

liyanhui1228 (approved):
LGTM!

liyanhui1228:
Looks like we still need @wangzhen127 for owner approval.

/assign @wangzhen127

cmd/nodeproblemdetector/node_problem_detector.go
```diff
@@ -72,13 +83,39 @@ func main() {
...
 	}
 
 	// Initialize NPD core.
 	p := problemdetector.NewProblemDetector(problemDaemons, npdExporters)
-	if err := p.Run(); err != nil {
-		glog.Fatalf("Problem detector failed with error: %v", err)
+	return p.Run(termCh)
+}
+
+func getProgramPath() (string, error) {
```
Random-Liu, line 95: Hm, I can't tell who is using this function.

cmd/options/options.go
```diff
@@ -87,6 +87,9 @@ type NodeProblemDetectorOptions struct {
 
 	// NodeName is the node name used to communicate with Kubernetes ApiServer.
 	NodeName string
+
+	// UnderTest indicates that the options will bypass certain checks.
+	UnderTest bool
```
Random-Liu, line 92: What does this mean? I don't quite get it.

cmd/nodeproblemdetector/node_problem_detector.go
```diff
@@ -34,15 +36,24 @@ import (
 	"k8s.io/node-problem-detector/pkg/version"
 )
 
-func main() {
-	npdo := options.NewNodeProblemDetectorOptions()
-	npdo.AddFlags(pflag.CommandLine)
+const (
+	serviceControlName = "service-action"
```
Random-Liu, line 40: who's using this?

cmd/nodeproblemdetector/node_problem_detector.go
```diff
@@ -34,15 +36,24 @@ import (
...
+	hourInSeconds      = uint32(60 * 60)
```
Random-Liu, line 41: ditto.

Random-Liu (commented):
Hm, not sure I missed any files.

I saw some unused code. Could you clean them up?

cmd/options/options.go
jeremyje: I created a nullexporter in the test instead of adding this.

jeremyje:
@Random-Liu Cleaned up the code should be good now.

PR #542, opened 2021-03-30
Update windows containerd config file to run without errors

mcshooter:


README.md
```diff
@@ -221,13 +221,13 @@ To develop NPD on Windows you'll need to setup your Windows machine for Go devel
...
-# Build
+# Build in MINGW64 Window
 make clean windows-binaries
 
-# Test
+# Test in MINGW64 Window
 make test
 
-# Run with containerd log monitoring enabled. (Assumes containerd is installed.)
+# Run with containerd log monitoring enabled in Powershell/Command Prompt. (Assumes containerd is installed.)
```
jeremyje, line 230: `%CD%` won't work in Powershell. Perhaps we should make the command exclusive to MinGW command prompt. In that case you need `$PWD`.

jeremyje (changes requested):

README.md
jeremyje: Also the `sc.exe` can only run in an Administrator mode prompt.

PR #545, opened 2021-04-23
Build Tarballs for each OS, Generate SHA-512 Checksums, and make Makefile easier to expand OS-Arch options.

jeremyje:
This change primarily extends the `Makefile` so that OS-Architecture specific tarballs are produced during `make build-tar`. Each tarballs will be suffixed with `$(GOOS)_$(GOARCH).tar.gz`. For example `node-problem-detector-v0.8.7-linux_amd64.tar.gz`.

This change is backwards compatible so the original file locations will be preserved, the artifacts in old locations are `cp`'ed from their new locations. The new locations are produced in an `output/$(GOOS)_$(GOARCH)/` directory which acts as a staging area for the tarball contents.

`make build-tar` will now produce tarballs for Linux and Windows.
`make push-tar` will also push those tarballs to GCS.

`*.tar.gz.sha512` checksum files are also produced for the new OS-Arch specific tarballs.
The `Makefile` has been tuned to allow it to be extended for different OS-Architectures in the future.


Makefile
```diff
@@ -243,13 +214,12 @@ push-container: build-container
 
 push-tar: build-tar
 	gsutil cp $(TARBALL) $(UPLOAD_PATH)/node-problem-detector/
+	gsutil cp node-problem-detector-$(VERSION)-*.tar.gz* $(UPLOAD_PATH)/node-problem-detector/
```
Random-Liu, line 217: `node-problem-detector-$(VERSION)-*` means that the old `node-problem-detector-$(VERSION).tar.gz` tarball won't be pushed?

jeremyje: Good catch! fixed.

jeremyje: Actually this was fine. The $(TARBALL) copy above will cover the legacy tarball.

jeremyje:
I'm adding back the `./bin/*` makefile rules because it can break the build if a user specifies `make GOOS=arm64 ENABLE_JOURNALD=0 build-tar`. The output will be a amd64 tarball rather than an arm64 tarball.

Makefile
Random-Liu: ACK

PR #551, opened 2021-04-30
add tolerations in daemonset config file

cmssczy:
No NPD pod will be deployed on the node with taint. Add this configuration to solve this problem.


Random-Liu:
/ok-to-test
/lgtm
/approve

Having NPD tolerate taints make sense to me. And this is just a reference yaml, people can always define their own.

PR #552, opened 2021-05-04
Add healthChecker functionality for kube-proxy service

mcshooter:
There have been some customer complaints that on windows node start ups, that when kube-proxy fails to start, some pods pods were entering into a crashloop state as services were inaccessible. Adding functionality to detect when kube-proxy is down, can attempt to repair the service and prevent pods from entering into a crashloop state.

### DIFF
--- a/config/windows-health-checker-kubeproxy.json
+++ b/config/windows-health-checker-kubeproxy.json
@@ -1,0 +1,34 @@
+{
+    "plugin": "custom",
+    "pluginConfig": {
+      "invoke_interval": "10s",
+      "timeout": "3m",
+      "max_output_length": 80,
+      "concurrency": 1
+    },
+    "source": "health-checker",
+    "metricsReporting": true,
+    "conditions": [
+      {
+        "type": "KubeProxyUnhealthy",
+        "reason": "KubeProxyIsHealthy",
+        "message": "kube-proxy on the node is functioning properly"
+      }
+    ],
+    "rules": [
+      {
+        "type": "permanent",
+        "condition": "KubeProxyUnhealthy",
+        "reason": "KubeProxyUnhealthy",
+        "path": "C:\\etc\\kubernetes\\node\\bin\\health-checker.exe",
+        "args": [
+          "--component=kube-proxy",
+          "--enable-repair=true",
+          "--cooldown-time=1m",
+          "--health-check-timeout=10s"
+        ],
+        "timeout": "3m"
+      }
+    ]
+  }
+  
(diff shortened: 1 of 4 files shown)


config/windows-health-checker-kubeproxy.json
```diff
@@ -0,0 +1,34 @@
...
+          "--component=kube-proxy",
+          "--enable-repair=true",
+          "--cooldown-time=1m",
+          "--health-check-timeout=10s"
+        ],
+        "timeout": "3m"
+      }
+    ]
+  }
+
```
jeremyje, line 34: newline if others dont have this then its ok

pkg/healthchecker/health_checker_windows.go
```diff
@@ -94,6 +89,18 @@ func getHealthCheckFunc(hco *options.HealthCheckerOptions) func() (bool, error)
 	return nil
 }
 
+// healthCheckEndpointOKFunc returns a function to check the status of an http endpoint
+func healthCheckEndpointOKFunc(endpoint string, timeout time.Duration) func() (bool, error) {
```
jeremyje, line 93: Seems like a missed opportunity to not make this generic health check through config. IE remove the hardcoded addresses and pass that in as a json argument.

/cc @Random-Liu what do you think?

jeremyje (changes requested):

config/windows-health-checker-kubeproxy.json
mcshooter: Other files do not have it.

jingxu97:
@mcshooter this seems can be extended to support other windows services like csi-proxy which is installed in the same place? 

cc @ddebroy

mcshooter:
> @mcshooter this seems can be extended to support other windows services like csi-proxy which is installed in the same place?
> 
> cc @ddebroy

@jingxu97 I am not too familiar with what csi-proxy is, but if it's similar to kubelet and kube-proxy and having csi-proxy being down would cause the node to become unhealthy, then yes, it should be something that NPD should be able to support.

pkg/healthchecker/health_checker_windows.go
Random-Liu: > @Random-Liu what do you think?

Yeah, we thought about introducing the Kubernetes like "healthiness probe": Exec probe + http probe.

The only problem is that the health checker is already a plugin of NPD, and now we need another layer of plugin. :P 

Some design is needed here.

Please feel free to suggest ideas or propose a design for this, but that doesn't necessarily need to block this change.

Random-Liu (commented):
/lgtm
/approve

PR #556, opened 2021-05-12
Add csi-proxy monitoring to health checker

mcshooter:
CSI-Proxy is being integrated on windows nodes, running as a windows service. It is used to help CSI Node Plugin interact with the node given that privileged containers are not yet supported on Windows. This change monitors to ensure that csi-proxy is running and available through NPD's health checker.

[#461](https://github.com/kubernetes/node-problem-detector/issues/461)

### DIFF
--- a/pkg/healthchecker/health_checker_windows.go
+++ b/pkg/healthchecker/health_checker_windows.go
@@ -85,6 +85,13 @@ func getHealthCheckFunc(hco *options.HealthCheckerOptions) func() (bool, error) 
 			}
 			return true, nil
 		}
+	case types.CsiProxyComponent:
+		return func() (bool, error) {
+			if _, err := powershell("Get-Process", types.CsiProxyComponent); err != nil {
+				return false, nil
+			}
+			return true, nil
+		}
 	}
 	return nil
 }
--- a/Makefile
+++ b/Makefile
@@ -214,6 +214,9 @@ $(NPD_NAME_VERSION)-%.tar.gz: $(ALL_BINARIES) test/e2e-install.sh
 	(cd output/$*/ && tar -zcvf ../../$@ *)
 	sha512sum $@ > $@.sha512
 
+windows-binaries: ENABLE_JOURNALD=0
+windows-binaries: $(foreach binary, $(BINARIES), output/windows_amd64/$(binary).exe)
+
 build-binaries: $(ALL_BINARIES)
 
 build-container: build-binaries Dockerfile
--- a/cmd/healthchecker/options/options.go
+++ b/cmd/healthchecker/options/options.go
@@ -46,16 +46,13 @@ type HealthCheckerOptions struct {
 
 // AddFlags adds health checker command line options to pflag.
 func (hco *HealthCheckerOptions) AddFlags(fs *pflag.FlagSet) {
-	fs.StringVar(&hco.Component, "component", types.KubeletComponent,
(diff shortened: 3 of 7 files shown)


pkg/healthchecker/health_checker_windows.go
```diff
@@ -85,6 +85,13 @@ func getHealthCheckFunc(hco *options.HealthCheckerOptions) func() (bool, error)
 			}
 			return true, nil
 		}
+	case types.CsiProxyComponent:
+		return func() (bool, error) {
+			if _, err := powershell("Get-Process", types.CsiProxyComponent); err != nil {
```
jeremyje, line 90: Does csi-proxy host an HTTP endpoint? I'd figure it would.

mcshooter: I have been looking for it as well, kind of like what kube-proxy has. But I haven't found anything like it. For now, I am using this to get some support in for csi-proxy. If I find it, I'll update it

jeremyje: It's using gRPC via named pipes in Windows. https://github.com/kubernetes-csi/csi-proxy/blob/9e1e33da998e4c1e3c7c4c00ceae64cebca0c258/internal/server/server.go#L73-L98 You'll need to import those client libraries and connect to the csi proxy endpoint.

Random-Liu, line 90: If we just check whether the process exists, wouldn't windows service already handled it?

Why do we need this health checker now then?

mcshooter: I am currently looking into a way to connect to csi-proxy endpoint for NPD Health Checker to use to test if it is up and running. In the meantime, this was used to test it.

mauriciopoppe: The list of named pipes can be obtained with: `[System.IO.Directory]::GetFiles("\\.\\pipe\\")`, in my dev env it shows as:

```
\\.\\pipe\\csi-proxy-filesystem-v1alpha1
\\.\\pipe\\csi-proxy-filesystem-v1beta1
\\.\\pipe\\csi-proxy-filesystem-v1beta2
\\.\\pipe\\csi-proxy-disk-v1alpha1
\\.\\pipe\\csi-proxy-disk-v1beta1
\\.\\pipe\\csi-proxy-disk-v1beta2
\\.\\pipe\\csi-proxy-disk-v1beta3
\\.\\pipe\\csi-proxy-volume-v1alpha1
\\.\\pipe\\csi-proxy-volume-v1beta1
\\.\\pipe\\csi-proxy-volume-v1beta2
\\.\\pipe\\csi-proxy-volume-v1beta3
\\.\\pipe\\csi-proxy-smb-v1alpha1
\\.\\pipe\\csi-proxy-smb-v1beta1
\\.\\pipe\\csi-proxy-smb-v1beta2
\\.\\pipe\\csi-proxy-system-v1alpha1
\\.\\pipe\\csi-proxy-iscsi-v1alpha1
\\.\\pipe\\csi-proxy-iscsi-v1alpha2
```

Unfortunately there's no `csi-proxy-healthz` named pipe or similar, there are only versioned APIs for different API groups.

Checking that the process exists should be good enough or if a new endpoint is needed we can add something like `\\.\\pipe\\csi-proxy-healthz`, if we decide to go with this route you'd need to import the csi-proxy client library as @jeremyje says.

cc @jingxu97

PR #557, opened 2021-05-14
Make sure the path to known-modules.json is relative

vteratipally:
As we are adding the knownModuleConfig path through system-stats-monitor.json, it would make sense to handle the knownModuleCOnfig path to be relative to the system-stats-monitor.json rather than absolute which could result in errors where npd is installed in a different dir and config in other.


pkg/systemstatsmonitor/system_stats_monitor.go
```diff
@@ -89,6 +90,9 @@ func NewSystemStatsMonitorOrDie(configPath string) types.Monitor {
 		ssm.memoryCollector = NewMemoryCollectorOrDie(&ssm.config.MemoryConfig)
 	}
 	if len(ssm.config.OsFeatureConfig.MetricsConfigs) > 0 {
+		// update the known_modules_path to relative the system-stats-monitors path
+		ssm.config.OsFeatureConfig.KnownModulesConfigPath = filepath.Join(filepath.Dir(configPath),
```
SergeyKanzhelev, line 94: can the test be added here: https://github.com/kubernetes/node-problem-detector/blob/master/pkg/systemstatsmonitor/system_stats_monitor_test.go?

Random-Liu, line 94: We should only do this when `configPath` is relative, and we should update the README.md about this https://github.com/kubernetes/node-problem-detector/blob/master/pkg/systemstatsmonitor/README.md#os-features

vteratipally: updated the readme. 
>We should only do this when configPath is relative 
Could you please clarify.

SergeyKanzhelev: the suggestion here is to do `filepath.Join` only when `KnownModulesConfigPath` is a relative path. So it will work fine with this fix: https://github.com/kubernetes/kubernetes/pull/101629

pkg/systemstatsmonitor/README.md
```diff
@@ -92,7 +92,8 @@ UnknownModules are derived from the /proc/modules compared with the known-module
 
 And an option:
 `knownModulesConfigPath`: The path to the file that contains the known modules(default
-modules) can be set. By default, the path is set to `known-modules.json`
+modules) can be set. By default, the path is set to `known-modules.json`
```
Random-Liu, line 95: `guestosconfig/known-modules.json`

SergeyKanzhelev (approved):

PR #560, opened 2021-05-19
Rename problem_counter and problem_gauge metrics

adriananeci:
Rename the following metrics:
- `problem_counter` to `npd_problem_counter`
- `problem_gauge` to `npd_problem_gauge`

As a Prometheus [naming convention](https://prometheus.io/docs/practices/naming/) best practices, a metric name should have a (single-word) application prefix relevant to the domain the metric belongs to. The prefix is sometimes referred to as namespace by client libraries. 

For metrics specific to an application, the prefix is usually the application name itself.

This is mostly useful in order to avoid interfering with metrics from other systems.


vteratipally:
this seems to be a breaking change as we are renaming the metrics.

vteratipally:
this would effect the existing pipelines that are dependent on the problem_counter metrics. Unfortunately we cannot accept this change as it is renaming the metrics.

Issue #562, opened 2021-05-21
NPD cannot work because of missing rbac config


cmssczy:
NPD's log as follows:
`E0521 10:04:21.057999       1 manager.go:162] failed to update node conditions: nodes "node1" is forbidden: User "system:serviceaccount:kube-system:default" cannot patch resource "nodes/status" in API group "" at the cluster scope`

pezzak:
Also need to add 'get' permission to 'nodes' resources in api group "", because https://github.com/kubernetes/node-problem-detector/blob/master/pkg/exporters/k8sexporter/k8s_exporter.go#L110 getNode() requires it.

pezzak:
And also NPD needs permissions to events:
```
  - apiGroups:
    - ""
    resources:
    - events
    verbs:
    - create
    - patch
```

PR #563, opened 2021-05-21
add rbac config

cmssczy:
fix https://github.com/kubernetes/node-problem-detector/issues/562

### DIFF
--- a/deployment/rbac.yaml
+++ b/deployment/rbac.yaml
@@ -1,0 +1,18 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: node-problem-detector
+subjects:
+- kind: ServiceAccount
+  name: node-problem-detector
+  namespace: kube-system
+roleRef:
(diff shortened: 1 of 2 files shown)


deployment/rbac.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: node-problem-detector
```
Random-Liu, line 4: NPD has a built-in role in Kubernetes `system:node-problem-detector`. We can probably use that.

Random-Liu (commented):
LGTM with one nit.

deployment/rbac.yaml
cmssczy: done

cmssczy: @Random-Liu PTAL

cmssczy:
friendly ping @Random-Liu for approval

yankay:
I had the same problem，It's nice to see that.

Issue #559, opened 2021-05-15
Windows Defender Plugin not found when running as a Service


jeremyje:
Windows Defender plugin does not start in a Windows Service because the $PWD is `%SystemRoot%\Windows\System32`. It's very difficult to change the `$PWD` in Windows (https://serverfault.com/questions/114238/windows-service-can-i-configure-the-current-working-directory). A better approach is to just use fully qualified paths.

/sig windows
/assign @mcshooter

PR #567, opened 2021-05-24
config/windows-defender-monitor.json

mcshooter:
Currently, the windows defender custom plugin path is a local reference. This needs to be updated to a full path location. 

https://github.com/kubernetes/node-problem-detector/issues/559


Random-Liu (commented):
LGTM

PR #572, opened 2021-06-02
fix build in docker

Iceber:
1. fix `make build-in-docker`
2. update base image of the builder image

ref: https://github.com/kubernetes/node-problem-detector/blob/master/.travis.yml

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -256,8 +256,8 @@ docker-builder:
 
 build-in-docker: clean docker-builder
 	docker run \
-		-v `pwd`:/gopath/src/k8s.io/node-problem-detector/ npd-builder:latest bash \
-		-c 'cd /gopath/src/k8s.io/node-problem-detector/ && make build-binaries'
+		-v `pwd`:/node-problem-detector/ npd-builder:latest bash \
+		-c 'cd /node-problem-detector/ && make build-binaries'
 
 push-container: build-container
 	gcloud auth configure-docker
--- a/builder/Dockerfile
+++ b/builder/Dockerfile
@@ -12,14 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.11.0
+FROM golang:1.16
 LABEL maintainer="Andy Xie <andy.xning@gmail.com>"
 
-ENV GOPATH /gopath/
-ENV PATH $GOPATH/bin:$PATH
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -256,8 +256,8 @@ docker-builder:
 
 build-in-docker: clean docker-builder
 	docker run \
-		-v `pwd`:/gopath/src/k8s.io/node-problem-detector/ npd-builder:latest bash \
-		-c 'cd /gopath/src/k8s.io/node-problem-detector/ && make build-binaries'
+		-v `pwd`:/node-problem-detector/ npd-builder:latest bash \
+		-c 'cd /node-problem-detector/ && make build-binaries'
```
Iceber, line 260: It could be confusing to put the project that use `go module` under `/gopath/...`

Iceber:
@Random-Liu PTAL, Thanks

builder/Dockerfile
```diff
@@ -12,14 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.11.0
+FROM golang:1.16.4
```
Random-Liu, line 15: How about just 1.16?

Random-Liu (commented):
LGTM with one nit

builder/Dockerfile
Iceber: Updated, Thanks

Iceber:
@jeremyje  @Random-Liu PTAL, Thanks :)

Makefile
jeremyje: I like this change. Can you verify that the `chown` is similar to the gopath for a docker build? I think it might adjust it to root as compared to the `/gopath` directory.

builder/Dockerfile
```diff
@@ -12,14 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.11.0
+FROM golang:1.16
 LABEL maintainer="Andy Xie <andy.xning@gmail.com>"
```
jeremyje, line 16: @Random-Liu Should we have a separate PR to basically change this (and friends) to Kubernetes Authors?

jeremyje (changes requested):

jeremyje (approved):
/lgtm for current state. There are comments but I think follow up PR is appropriate.

jeremyje:
@Random-Liu can you approve this change?

PR #575, opened 2021-06-08
Check kube-proxy health on linux

uthark:
Right now healthchecker supports kube-proxy on windows only.

This PR adds support for linux.

Other changes:
- refactored code for better reuse
- log if component is not supported.

### DIFF
--- a/pkg/healthchecker/health_checker.go
+++ b/pkg/healthchecker/health_checker.go
@@ -17,6 +17,10 @@ limitations under the License.
 package healthchecker
 
 import (
+	"context"
+	"net/http"
+	"os/exec"
+	"strings"
 	"time"
 
 	"github.com/golang/glog"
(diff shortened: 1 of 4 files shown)


jeremyje:
Thanks for the PR. Seems fine I'd optionally suggest that there's a test/restructure added to ensure that Windows and Linux are supported moving forward but perhaps a bug is sufficient. @Random-Liu 

/lgtm
/ok-to-test

uthark:
@Random-Liu could you please take a look?

uthark:
@jeremyje @Random-Liu 

I moved `getHealthCheckFunc` to common code and added test to confirm it returns non-nil function for the supported components. 

Not sure what is the best way to abstract path to docker file. For CRI it is passed with CLI args, but docker/docker.exe were hardcoded. Do you want it to be configurable too?

pkg/healthchecker/health_checker.go
```diff
@@ -110,3 +112,64 @@ func logPatternHealthCheck(service string, logPatternsToCheck map[string]int) (b
...
+		}
+	default:
+		glog.Warningf("Unsupported component: %v", hco.Component)
+	}
+
+	return nil
+}
+
+func getDockerPath() string {
+	if runtime.GOOS == "windows" {
```
jeremyje, line 158: Given that there's already a _windows.go and _linux.go file just define the methods there.

This has 2 benefits:
1. It's clear that when a new OS is supported that this method needs to be considered for porting.
2. The logic between OSes have lower risk of leaking into each other's binaries.

pkg/healthchecker/health_checker_test.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package healthchecker
 
 import (
+	"k8s.io/node-problem-detector/cmd/healthchecker/options"
```
Random-Liu, line 20: nit: Put this into the import group below

pkg/healthchecker/health_checker.go
Random-Liu: +1

Random-Liu (commented):
LGTM, please address the comments.

Random-Liu:
And thanks for working on this! This is very useful feature. :)

jeremyje:
/lgtm

It's possible the failures are coming about due to the kube-proxy checks are failing. The easiest way to verify is to create a Kubernetes cluster and swap out the NPD binary and see how it responds.
`kubectl get node [NODE_NAME] -o yaml` and look at the conditions.

PR #580, opened 2021-06-15
update CriCtl path for windows

mcshooter:
Containerd is constantly being restarted due to NPD's healthchecker. The CriCtl path that it uses to detect the health of containerd is incorrect. This change updates the path for windows nodes to the correct path and will prevent containerD from restarting all the time.

### DIFF
--- a/pkg/healthchecker/types/types_windows.go
+++ b/pkg/healthchecker/types/types_windows.go
@@ -17,7 +17,7 @@ limitations under the License.
 package types
 
 const (
-	DefaultCriCtl        = "C:/node/crictl.exe"
+	DefaultCriCtl        = "C:/etc/kubernetes/node/bin/crictl.exe"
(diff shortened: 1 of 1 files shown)


pkg/healthchecker/types/types_windows.go
```diff
@@ -17,7 +17,7 @@ limitations under the License.
 package types
 
 const (
-	DefaultCriCtl        = "C:/node/crictl.exe"
```
Random-Liu, line 20: You can configure this in the health checker config, instead of changing the default.

Random-Liu: Offline discussed. `C:/etc/kubernetes/node/bin/crictl.exe` is the default path in the OSS setup https://github.com/kubernetes/kubernetes/blob/1400be60ff7f01fd2799ee897a95859693c93e85/cluster/gce/windows/k8s-node-setup.psm1#L1310.

It should be fine to default to that path.

PR #583, opened 2021-06-17
Log error from kube-api

### LINKED ISSUE #582: Log error from kube-apiserver
NPD has check if kube-apiserver ready. This check uses wait.PullImmediate function and it doesn't return error from kube-apiserver, instead of this it returns 'timed out waiting for the condition' in any case:

```
I0617 12:29:26.326219       1 k8s_exporter.go:54] Waiting for kube-apiserver to be ready (timeout 30s)...
W0617 12:29:57.423619       1 k8s_exporter.go:56] kube-apiserver did not become ready: timed out on waiting for kube-apiserver to return the node object: timed out waiting for the condition
```

And it makes it harder to find the error

pezzak:
fix #582


pezzak:
I signed it

Random-Liu:
The log only generates at NPD startup, so should be fine.

/lgtm
/approve

PR #587, opened 2021-06-30
Add a check if the metric is nil so that collector doesn't collect metrics.

vteratipally:
Add a check if the metric is nil so that collector doesn't collect metrics.

NPD collects the metrics based on the input system-stats-monitor.json. If the metric config is present in the input, NPD shouldn't collect the metrics, but currently NPD still tries to collect the metrics even the metric is not present and which would result in nil pointer. 

This should fix the issue.

### DIFF
--- a/pkg/systemstatsmonitor/cpu_collector_linux.go
+++ b/pkg/systemstatsmonitor/cpu_collector_linux.go
@@ -25,7 +25,9 @@ import (
 )
 
 func (cc *cpuCollector) recordLoad() {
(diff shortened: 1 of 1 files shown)


vteratipally:
/test pull-npd-e2e-kubernetes-gce-gci
--


<br class="Apple-interchange-newline">

vteratipally:
closes issue https://github.com/kubernetes/node-problem-detector/issues/614

PR #589, opened 2021-07-02
Update the instructions for build image in the readme

xinydev:
`libsystemd-journal-dev` only exists in `ubuntu trusty`,https://www.ubuntuupdates.org/package_metas?q=libsystemd-journal-dev. 

we should use `libsystemd-dev` instead,https://www.ubuntuupdates.org/package_metas?q=libsystemd-dev


xinydev:
It's weird that no `approved label` added

PR #596, opened 2021-07-23
remove aufs hung check

lizhuqi:
Remove aufs related checks.

Dcoker now perfers overlay2 to aufs for its current versions and it no longer depends on aufs.


Random-Liu:
/ok-to-test

It should be OK to remove AUFS from the default config.

/lgtm
/apprve

lizhuqi:
last run was still testing against old test code. My change in https://github.com/kubernetes/kubernetes/pull/104048 was not applied.

Rerun the test.

/retest

PR #597, opened 2021-07-27
add mount_point for metric disk/bytes_used.

yozora-hitagi:
Most of the time, we are more concerned with disk usage at the mount point.

Signed-off-by: yozora <506119423@qq.com>


yozora-hitagi:
I signed it

PR #598, opened 2021-07-27
add mount_point for metric disk/bytes_used.

yozora-hitagi:
Most of the time, we are more concerned with disk usage at the mount point.

This add mountpoint label for disk_bytes_used

Signed-off-by: yozora <506119423@qq.com>


vteratipally:
Thank you for the change. Adding a new field to the metric could be a breaking change especially for metrics exporting from stackdriver. Would it makes sense to add a optional flag  like includeMountPts in the metricconfig to control adding the mount points.

Random-Liu:
@vteratipally  will rely on you to give a judgement call about this change.

PR #599, opened 2021-07-29
Fix NPD hash for test.

Random-Liu:
The Kubernetes startup script downloaded the amd64 image, but we are taking hash for the no-arch one.


Random-Liu:
@mcshooter

PR #607, opened 2021-08-12
Create cloudbuild file

mmiranda96:
## What does this PR do?
Create file `cloudbuild.yaml`, which is required for [image pushing job](https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/k8s-staging-node-problem-detector.yaml).

I took [kubernetes/dns](https://github.com/kubernetes/dns/blob/master/cloudbuild.yaml) as an example. Please let me know if there's anything that must be changed.

Closes https://github.com/kubernetes/test-infra/issues/23202

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,0 +1,21 @@
+# See https://cloud.google.com/cloud-build/docs/build-config
+
+# this must be specified in seconds. If omitted, defaults to 600s (10 mins)
+timeout: 7200s
+steps:
+  - name: 'gcr.io/k8s-testimages/gcb-docker-gcloud:latest'
+    entrypoint: bash
+    env:
+    - PROW_GIT_TAG=$_GIT_TAG
+    - PULL_BASE_REF=$_PULL_BASE_REF
+    - VERSION=$_PULL_BASE_REF
+    args:
+    - -c
+    - |
+      echo "Building/Pushing NPD containers"
+      make push-container
+substitutions:
+  # _GIT_TAG will be filled with a git-based tag for the image, of the form vYYYYMMDD-hash, and
+  # can be used as a substitution
+  _GIT_TAG: 'PLACE_HOLDER'
+  _PULL_BASE_REF: 'master'


pacoxu:
LGTM
👍

cloudbuild.yaml
```diff
@@ -0,0 +1,21 @@
+# See https://cloud.google.com/cloud-build/docs/build-config
+
+# this must be specified in seconds. If omitted, defaults to 600s (10 mins)
+timeout: 7200s
+steps:
+  - name: 'gcr.io/k8s-testimages/gcb-docker-gcloud:v20210722-085d930'
```
SergeyKanzhelev, line 6: do you need a specific version here? Would latest suffice?

cloudbuild.yaml
```diff
@@ -0,0 +1,21 @@
...
+    - VERSION=$_PULL_BASE_REF
+    args:
+    - -c
+    - |
+      echo "Building/Pushing NPD containers"
+      make push-container
+substitutions:
+  # _GIT_TAG will be filled with a git-based tag for the image, of the form vYYYYMMDD-hash, and
+  # can be used as a substitution
+  _GIT_TAG: '12345'
```
elfinhe, line 20: minor: is this '12345' a placeholder? better use 'PLACE_HOLDER' I guess.

cloudbuild.yaml
mmiranda96: This is the latest version available (at the time of the commit). But I guess it's really not important, I can switch to `latest`.

cloudbuild.yaml
Random-Liu: Agree

Random-Liu (commented):
Based on my understanding, with this, we don't need to run `make push-container` ourselves any more?

Once we send a change to https://github.com/kubernetes/test-infra/tree/master/config/jobs/image-pushing, the image will be automatically pushed?

cloudbuild.yaml
mmiranda96: Updated to latest.

mmiranda96:
Yes, it looks like this job will build new images for `node-problem-detector` when submitted to master.

elfinhe (approved):

Random-Liu:
> Yes, it looks like this job will build new images for node-problem-detector when submitted to master.

Cool! Let's try it out!

Thanks

/lgtm
/approve

PR #609, opened 2021-08-13
Prevent uptimeFunc from being called everytime CheckHealth is called

mcshooter:
There is currently an issue for Windows nodes where NPD is causing the CPU to hit and persist at 100%. After investigating, it looks like there was a change that moved uptimeFunc to be called every time we checked the health through health checker. We shouldn't be calling uptimeFunc unless we know it has been determined that one of the services is unhealthy and there are no patterns to be checking. For Windows, because the query to get and calculate the uptime is pretty heavy, having it be called every 10s or so for each of the separate services would unnecessarily increase usage. This change will revert part of how we CheckHealth and move the loopBackTime logic inside of the logPatternHealthCheck function to only be done if there are log patterns that need to be checked.

### DIFF
--- a/pkg/healthchecker/health_checker.go
+++ b/pkg/healthchecker/health_checker.go
@@ -65,32 +65,27 @@ func NewHealthChecker(hco *options.HealthCheckerOptions) (types.HealthChecker, e
 // CheckHealth checks for the health of the component and tries to repair if enabled.
 // Returns true if healthy, false otherwise.
 func (hc *healthChecker) CheckHealth() (bool, error) {
-	var logStartTime string
 	healthy, err := hc.healthCheckFunc()
 	if err != nil {
 		return healthy, err
 	}
-	uptime, err := hc.uptimeFunc()
-	if err != nil {
-		glog.Warningf("Failed to get the uptime: %+v", err)
-		return true, err
-	}
-	if hc.loopBackTime > 0 && uptime > hc.loopBackTime {
-		logStartTime = time.Now().Add(-hc.loopBackTime).Format(types.LogParsingTimeLayout)
-	} else {
-		logStartTime = time.Now().Add(-uptime).Format(types.LogParsingTimeLayout)
-	}
-	logPatternHealthy, err := logPatternHealthCheck(hc.service, logStartTime, hc.logPatternsToCheck)
+	logPatternHealthy, err := logPatternHealthCheck(hc.service, hc.loopBackTime, hc.logPatternsToCheck)
 	if err != nil {
 		return logPatternHealthy, err
 	}
 	if healthy && logPatternHealthy {
 		return true, nil
 	}
+
 	// The service is unhealthy.
 	// Attempt repair based on flag.
 	if hc.enableRepair {
 		// repair if the service has been up for the cool down period.
+		uptime, err := hc.uptimeFunc()
+		if err != nil {
+			glog.Infof("error in getting uptime for %v: %v\n", hc.component, err)
+			return false, nil
+		}
 		glog.Infof("%v is unhealthy, component uptime: %v\n", hc.component, uptime)
 		if uptime > hc.coolDownTime {
 			glog.Infof("%v cooldown period of %v exceeded, repairing", hc.component, hc.coolDownTime)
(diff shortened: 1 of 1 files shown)


mcshooter:
cc/ @Random-Liu

pkg/healthchecker/health_checker.go
```diff
@@ -65,32 +65,28 @@ func NewHealthChecker(hco *options.HealthCheckerOptions) (types.HealthChecker, e
...
-	}
-	logPatternHealthy, err := logPatternHealthCheck(hc.service, logStartTime, hc.logPatternsToCheck)
+	logPatternHealthy, err := logPatternHealthCheck(hc.service, hc.loopBackTime, hc.logPatternsToCheck)
 	if err != nil {
 		return logPatternHealthy, err
 	}
 	if healthy && logPatternHealthy {
 		return true, nil
 	}
+	glog.Infof("healthy boolean: %v\n", healthy)
```
jeremyje, line 79: Remove debug statements

pkg/healthchecker/health_checker.go
```diff
@@ -102,10 +98,21 @@ func (hc *healthChecker) CheckHealth() (bool, error) {
...
 		return true, nil
 	}
+	uptimeFunc := getUptimeFunc(service)
+	uptime, err := uptimeFunc()
+	if err != nil {
+		glog.Warningf("Failed to get the uptime: %+v", err)
+		return true, err
+	}
+
+	logStartTime := time.Now().Add(-uptime).Format(types.LogParsingTimeLayout)
```
jeremyje, line 112: For more predictability capture `time.Now()` at 1 point above, probably before the `uptimeFunc()` call.

jeremyje (changes requested):

pkg/healthchecker/health_checker.go
mcshooter: the logStartTime needs the uptime so it cannot go before uptimeFunc() call

pkg/healthchecker/health_checker.go
```diff
@@ -65,32 +65,28 @@ func NewHealthChecker(hco *options.HealthCheckerOptions) (types.HealthChecker, e
...
 	}
+	glog.Infof("healthy boolean: %v\n", healthy)
+	glog.Infof("logPatternHealthy boolean: %v\n", logPatternHealthy)
+
 	// The service is unhealthy.
 	// Attempt repair based on flag.
 	if hc.enableRepair {
 		// repair if the service has been up for the cool down period.
+		uptime, err := hc.uptimeFunc()
+		if err != nil {
```
ibabou, line 87: We shouldn't continue executing if this failed ?

pkg/healthchecker/health_checker.go
```diff
@@ -102,10 +96,22 @@ func (hc *healthChecker) CheckHealth() (bool, error) {
 
 // logPatternHealthCheck checks for the provided logPattern occurrences in the service logs.
 // Returns true if the pattern is empty or does not exist logThresholdCount times since start of service, false otherwise.
-func logPatternHealthCheck(service, logStartTime string, logPatternsToCheck map[string]int) (bool, error) {
+func logPatternHealthCheck(service string, loopBackTime time.Duration, logPatternsToCheck map[string]int) (bool, error) {
```
ibabou, line 99: Will this function always return in the windows case ? len(logPatternsToCheck) == 0 -> true ?

pjh (commented):
I reviewed but don't have much to offer. At a quick glance it's challenging to read the code here (existing and modified) since I'm not sure what loopBackTime means.

But +1 to the goal of not calling the uptimeFunc as often to reduce CPU utilization. We'll still need to work on further changes to avoid starting numerous Powershell sessions on Windows.

pkg/healthchecker/health_checker.go
mcshooter: Not sure. As of now, currently with our windows health checkers we don't have any patterns to check, but that's not to say that it may be different for future problem detectors

pkg/healthchecker/health_checker.go
mcshooter: it will eventually still return false. It will output another log that may not be super necessary, but yes, it is something we can do to return early.

mcshooter:
Was able to test locally on a kubernetes node 
Screen Shot 2021-08-16 at 6 25 53 PM

pkg/healthchecker/health_checker.go
ibabou: I believe we should return then early. having err here means that uptime might not have a proper value.

ibabou (changes requested):

ibabou (approved):

Random-Liu:
@liyanhui1228

Random-Liu:
In this case, do we still intend to use log pattern or repair on windows?

If not, can we just declare it as not properly supported, and return error if users try to use log pattern or repair on windows?

In that way, we can no-op those functions for windows instead.

mcshooter:
> In this case, do we still intend to use log pattern or repair on windows?
> 
> If not, can we just declare it as not properly supported, and return error if users try to use log pattern or repair on windows?
> 
> In that way, we can no-op those functions for windows instead.
> 
> The current code makes me feel like we will never use it, because it has serious performance issue. In that case, I prefer we disable the feature and cleanup the corresponding code, instead of moving the code around to make the code never run just for us. Because others may still hit the issue if they try to use it.

We still do use the repair functionality. So it doesn't make sense to remove this as a windows functionality. The code will still run to repair and call uptime but only in the case when it expects it to.

Random-Liu:
Offline discussed, there is no good way to get the last start time when the service is down.

So the current fix makes sense to me

PR #618, opened 2021-10-07
Fix pattern for kubelet starting up

danlenar:
Fix pattern for kubelet.

Match the description found in
https://github.com/kubernetes/release/blob/master/packages/rpm/kubelet.service


vteratipally:
Thank you for the PR, kubelet.service is configured differently based on platform and the package manager. As these files can be configured based on the requirement, we cannot accept the PR change.

PR #619, opened 2021-10-21
feat: pod restart monitor for detecting broken daemonsets

iamnoah:
Hi, I wrote a really simple daemon to solve an issue we are having with the aws-node daemon restarting frequently and making nodes unstable. I'd love feedback on how to make it better and contribute.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -257,10 +257,11 @@ build: build-container build-tar
 docker-builder:
 	docker build -t npd-builder . --target=builder
 
+DOCKER_BUILD_VARS ?=
 build-in-docker: clean docker-builder
 	docker run \
 		-v `pwd`:/gopath/src/k8s.io/node-problem-detector/ npd-builder:latest bash \
-		-c 'cd /gopath/src/k8s.io/node-problem-detector/ && make build-binaries'
+		-c 'cd /gopath/src/k8s.io/node-problem-detector/ && make build-binaries $(DOCKER_BUILD_VARS)'
 
 push-container: build-container
 	# So we can push to docker hub by setting REGISTRY
--- a/pkg/podrestartmonitor/README.md
+++ b/pkg/podrestartmonitor/README.md
@@ -1,0 +1,62 @@
+# Pod Restart Monitor
+
+*Pod Restart Monitor* is a problem daemon in node problem detector. It monitors the number of times a pod has been
+restarted.
+
+The original use case was for monitoring the `aws-node` DaemonSet in EKS clusters. Multiple restarts of that pod can
+indicate a hard to define, but definite underlying condition on a node that makes it unsuitable for scheduling.
+
+### Plugin Configuration
+
+The plugin checks a very simple condition: how many times any container in a pod has restarted in its lifetime. 
+This condition will not go away unless the pod is deleted.
+
+The defaults are tuned for the aws-node daemon in EKS clusters, but any pod can be monitored:
+
+/config/aws-restart-monitor.json:
+
+```json
+{
+  "plugin": "podrestart",
+  "source": "pod-restart-monitor",
+  "namespace": "kube-system",
+  "podSelector": "k8s-app=aws-node",
+  "restartThreshold": 5,
+  "checkInterval": "5m"
+}
+```
+
+### RBAC Requirements
+
+The plugin needs to be able to get pods and nodes, which can be accomplished with an additional ClusterRole:
+
+```yaml
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: get-pods-and-nodes
+rules:
+  - verbs:
+      - get
+      - list
+      - watch
+    apiGroups:
+      - ""
+    resources:
+      - pods
+      - nodes
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: node-problem-detector-get-pods-and-nodes
+subjects:
+  - kind: ServiceAccount
+    name: node-problem-detector
+    namespace: kube-system
+roleRef:
+  kind: ClusterRole
+  name: npd-list-pods
+  apiGroup: rbac.authorization.k8s.io
+```
--- a/pkg/podrestartmonitor/pod_restart_monitor.go
+++ b/pkg/podrestartmonitor/pod_restart_monitor.go
@@ -1,0 +1,199 @@
+/*
+Copyright 2016 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package podrestartmonitor
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"time"
+
+	"github.com/golang/glog"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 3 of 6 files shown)


iamnoah:
/remove-lifecycle stale

Is this project dead?

pkg/podrestartmonitor/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+The original use case was for monitoring the `aws-node` DaemonSet in EKS clusters. Multiple restarts of that pod can
+indicate a hard to define, but definite underlying condition on a node that makes it unsuitable for scheduling.
+
+### Plugin Configuration
+
+The plugin checks a very simple condition: how many times any container in a pod has restarted in its lifetime. 
+This condition will not go away unless the pod is deleted.
+
+The defaults are tuned for the aws-node daemon in EKS clusters, but any pod can be monitored:
```
mmiranda96, line 14: Already mentioned, but I think the pod selector should not have a default value.

pkg/podrestartmonitor/pod_restart_monitor.go
```diff
@@ -0,0 +1,183 @@
...
+			}
+			for _, pod := range list.Items {
+				for _, c := range pod.Status.ContainerStatuses {
+					count := c.RestartCount
+					if count > l.config.RestartThreshold {
+						l.statuses <- &types.Status{
+							Source: PodRestartMonitorName,
+							Events: nil,
+							Conditions: []types.Condition{
+								{Type: ConditionTooManyRestarts, Status: types.True, Transition: time.Now(),
```
mmiranda96, line 146: What happens when we want to monitor more than one pod (either by having a broad `PodSelector` or configuring multiple monitors)? It looks like the condition will collide. Is there a workaround to prevent this?

mmiranda96 (changes requested):
Thank you for your contribution @iamnoah! I've reviewed your change. Some comments are just questions and/or suggestions, other are hard requests. Let me know if you have any questions.

builder/Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM golang:1.11.0
+FROM golang:1.15.0
```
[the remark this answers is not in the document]
iamnoah: The go.mod file specifies `go 1.15`. Trying to build docker images without this change fails with multiple "module requires Go 1.15" errors.

Makefile
```diff
@@ -254,10 +254,11 @@ build: build-container build-tar
 docker-builder:
 	docker build -t npd-builder ./builder
 
+DOCKER_BUILD_VARS ?=
 build-in-docker: clean docker-builder
 	docker run \
 		-v `pwd`:/gopath/src/k8s.io/node-problem-detector/ npd-builder:latest bash \
-		-c 'cd /gopath/src/k8s.io/node-problem-detector/ && make build-binaries'
+		-c 'cd /gopath/src/k8s.io/node-problem-detector/ && make build-binaries $(DOCKER_BUILD_VARS)'
```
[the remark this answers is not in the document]
iamnoah: I don't remember exactly, but I think I added this so I could override the value of ALL_BINARIES and build a docker image in a reasonable time. I can't get it to build at all now, build-in-docker just hangs when I try make.

pkg/podrestartmonitor/pod_restart_monitor.go
iamnoah: I don't have a good answer because I don't have that use case. I suppose the condition type could be configurable.

mmiranda96: Looking at [official docs](https://kubernetes.io/docs/concepts/architecture/nodes/#condition), it appears that we can use reason for this. Let's keep `condition.Type` equal across all pods, but `condition.Reason` can contain the pod name to distinguish.

pkg/podrestartmonitor/pod_restart_monitor.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+	if config.RestartThreshold == 0 {
+		config.RestartThreshold = 5
+	}
+	if config.Namespace == "" {
+		config.Namespace = "kube-system"
+	}
+	if config.PodSelector == "" {
+		return config, fmt.Errorf("did not specify a PodSelector")
+	}
```
2rs2ts, line 92: Although it would make the process of filtering pods more complex, it sure would be nice to be able to filter by annotations or by priority classes, in addition to labels. Feel free to ignore my comment here if you think it'd be a pain to implement, though.

pkg/podrestartmonitor/pod_restart_monitor.go
```diff
@@ -0,0 +1,199 @@
...
+func (l *podRestartMonitor) Stop() {
+	glog.Infof("Stopping monitor %s", l.configPath)
+	close(l.stop)
+}
+
+// ConditionTooManyRestarts indicates that a Pod has restarted too many times, indicating a problem with the node.
+const ConditionTooManyRestarts = "TooManyRestarts"
+
+// monitorLoop is the main loop of log monitor.
+func (l *podRestartMonitor) monitorLoop() {
```
2rs2ts, line 136: I think this comment, and the name `l`, are relics of this being copied from the log monitor package. It might be good to update these, or at least, the comment 😅

pkg/podrestartmonitor/pod_restart_monitor.go
```diff
@@ -0,0 +1,183 @@
...
+	nodeName := getNodeName()
+MainLoop:
+	for {
+		select {
+		case <-l.stop:
+			glog.Warningf("monitor stopped")
+			return
+		case <-time.After(time.Duration(l.config.CheckInterval)):
+			list, err := l.clientset.CoreV1().Pods(l.config.Namespace).List(metav1.ListOptions{
+				LabelSelector: l.config.PodSelector,
```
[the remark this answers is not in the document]
2rs2ts: I believe it will be an empty list. Try doing `kubectl get pod -l fakelabel=fakevalue -o json` in your cluster, you will get this:

```json
{
    "apiVersion": "v1",
    "items": [],
    "kind": "List",
    "metadata": {
        "resourceVersion": "",
        "selfLink": ""
    }
}
```

So the loop will just do nothing, I think. It might trip up some people who get their namespace settings wrong or typo the selector when configuring their checks, but it would be consistent with how `kubectl` works.

2rs2ts (commented):
Not a maintainer, just a user with a keen interest in this leaving some thoughts. I've written a custom plugin monitor that does something similar, but am always keen to delete custom code in favor of official, bundled code, so I'd love to see this PR merged.

pkg/podrestartmonitor/pod_restart_monitor.go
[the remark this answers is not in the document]
mmiranda96: That sounds good. I was worried that this might return an error, but an empty list is a handleable output.

For misconfigured selectors, we could have a warning log. But I don't see much value, I think it's good as is.

sb3lbx:
Until this is merged, are there suitable alternatives?  Either available custom plugins or solutions outside of node-problem-detector?

I ask because our `aws-node` pods seem to randomly timeout on health checks.  In an effort to steady them, we're locking the resource requests and limits to tell Kubernetes to ensure the resources are always available.  However, it'd be real nice to have a solution that would taint our nodes where `aws-node` restarts.  Every time it restarts it breaks all workloads on that node.  The issue persists so we have to manually clean it up which is contrary to the idea of a "self-healing platform".

2rs2ts:
@sb3lbx although I am not at liberty to share the code, I can at least say that at my company, we worked around this by writing a custom plugin monitor that runs a bash script that checks the readiness state, and the number of container restarts in a rolling window, of any pod marked with a certain annotation. (In our case, we use the same annotation as the one we chose for [nodetaint](https://github.com/wish/nodetaint).) We made the thresholds for failure configurable because we found that we ended up needing to tweak how many restarts were a problem before the node should actually be considered unhealthy.

Maybe someday, if this PR dies of inactivity, I will look into whether we can make an open-source version of our script and contribute it back to this repo. I'd have to clear it with my company's open sourcing department, though, and it'd be better if it wasn't a custom plugin but rather its own kind of check, so I really do hope the owner of this PR comes back soon.

sb3lbx:
Brilliant.  Thank you, @2rs2ts

Issue #586, opened 2021-06-26
multi-arch image


duckman:
I was excited to see arm64 added but when I pull k8s.gcr.io/node-problem-detector/node-problem-detector:v0.8.9 I still get `standard_init_linux.go:228: exec user process caused: exec format error`

nvtkaszpir:
I can confirm this, on rpi4 I get amd64 linux image:

`docker inspect k8s.gcr.io/node-problem-detector/node-problem-detector:v0.8.9`

```json
[
    {
        "Id": "sha256:ac45e7eae5e0ad5cf81ef5724f9a217d893c0b5946658fa7e2ce56df2cb623e1",
        "RepoTags": [
            "k8s.gcr.io/node-problem-detector/node-problem-detector:v0.8.9"
        ],
        "RepoDigests": [
            "k8s.gcr.io/node-problem-detector/node-problem-detector@sha256:c4a75e50c3ee30daa78b7149de781f66236885850b1ea7b0c1a062af5019e019"
        ],
        "Parent": "",
        "Comment": "",
        "Created": "2021-06-25T20:52:07.859708727Z",
        "Container": "a4855bd4a5acdbbcbab8348b964092bf16a651a5902fb926d9d40214cad68333",
        "ContainerConfig": {
            "Hostname": "a4855bd4a5ac",
            "Domainname": "",
            "User": "",
            "AttachStdin": false,
            "AttachStdout": false,
            "AttachStderr": false,
            "Tty": false,
            "OpenStdin": false,
            "StdinOnce": false,
            "Env": [
                "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
            ],
            "Cmd": [
                "/bin/sh",
                "-c",
                "#(nop) ",
                "ENTRYPOINT [\"/node-problem-detector\" \"--config.system-log-monitor=/config/kernel-monitor.json\"]"
            ],
            "Image": "sha256:8fd5e1e7afce44135ba7320c7c51f1793c6617e6d2f7284e7401e21c2ea79db2",
            "Volumes": null,
            "WorkingDir": "",
            "Entrypoint": [
                "/node-problem-detector",
                "--config.system-log-monitor=/config/kernel-monitor.json"
            ],
            "OnBuild": null,
            "Labels": {}
        },
        "DockerVersion": "20.10.2",
        "Author": "Random Liu <lantaol@google.com>",
        "Config": {
            "Hostname": "",
[...]

PR #644, opened 2022-02-13
Create multi-arch image

kelvie:
For linux arm64 and amd64, as per #586.

I moved the builder image into the same dockerfile, and bumped the Go
version on it. It didn't seem like the builder dockerfile worked with
the latest code anyway (the go modules require go 1.15 and higher).

This requires a recent enough docker install with buildx, as well as
an arm64 builder.

BASEIMAGE is changed to not specify an arch, so that the image will
build on its native arch in buildx.

Example image is on docker hub as:

    kelvie/node-problem-detector:v0.8.10-5-gb0fa610


better0332:
I need this too

yankay:
Greate PR.

rotarur:
if works but the health-checker reports wrong kubelet status KubeletUnhealthy

rotarur:
found the issue with kubelet. It uses http to get the status but without hostNetwork enabled it doesn't work. Now everything seems to work fine and in a cluster with mixed architectures, both amd64 and arm64 architectures, using multi-arch docker image

acumino:
/assign @vteratipally
Can you please review this?

otterley:
Hi maintainers! We have an increasing number of customers who would like to use NPD on arm64 CPUs. Is there anything I can do to help unblock this effort? Please feel free to contact me privately as well.

Makefile
```diff
@@ -239,8 +240,8 @@ $(NPD_NAME_VERSION)-%.tar.gz: $(ALL_BINARIES) test/e2e-install.sh
 
 build-binaries: $(ALL_BINARIES)
 
-build-container: build-binaries Dockerfile
-	docker build -t $(IMAGE) --build-arg BASEIMAGE=$(BASEIMAGE) --build-arg LOGCOUNTER=$(LOGCOUNTER) .
+build-container: clean Dockerfile
```
otterley, line 243: Needs

```
docker buildx create --use
```

to be run first. I recommend a prerequisite `.PHONY` target for this.

Makefile
```diff
@@ -252,16 +253,20 @@ build-tar: $(TARBALL) $(ALL_TARBALLS)
...
 		-c 'cd /gopath/src/k8s.io/node-problem-detector/ && make build-binaries'
 
 push-container: build-container
+	# So we can push to docker hub by setting REGISTRY
+ifneq (,$(findstring gcr.io,$(REGISTRY)))
 	gcloud auth configure-docker
-	docker push $(IMAGE)
+endif
+	# Build should be cached from build-container
+	docker buildx build --push --platform $(DOCKER_PLATFORMS) -t $(IMAGE) --build-arg BASEIMAGE=$(BASEIMAGE) --build-arg LOGCOUNTER=$(LOGCOUNTER) .
```
otterley, line 269: Needs

```
docker buildx create --use
```

to be run first. I recommend a prerequisite `.PHONY` target (e.g. `create-buildx-builder`) for this.

otterley:
I don't think this PR is going to work as written, because it requires an arm64 build instance. The build nodes are all x86.

Issue #645, opened 2022-02-13
Custom Plugin: Allow users to define condition startup behavior


notchairmk:
## Problem
There is currently no way for `custompluginmonitor` users to define startup behavior for ComponentCondition. Both `custompluginmonitor` and `systemlogmonitor` will emit a ComponentCondition with default `Status: "False"` [during initialization](https://github.com/kubernetes/node-problem-detector/blob/master/pkg/custompluginmonitor/custom_plugin_monitor.go#L308) for all monitored conditions.

Assuming that no problems exist at NPD startup doesn't always make sense. For example being, when the NPD container is restarting on a node. In this case, depending on restart frequency, the ConditionStatus on the node ends up flip-flopping between True and False/Unknown

### Option A: Allow user to specify startup behavior on the plugin
- Example:
  ```json
  "plugin": "custom",
  "source": "custom-health",
  "pluginConfig": {
      "skip_initial_status": true
  },
  "conditions": [{
      "type": "KubeletUnhealthy",
      "reason": "KubeletIsHealthy",
      "message": "kubelet on the node is functioning properly"
  }]
  ```
- Pros:
  - Allowing user to skip during initialization still falls within [status API conventions](https://github.com/kubernetes/community/blob/ab46b8ce07ef17aa83b862c5053270f8180ef5f9/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties)
  - Reduced complexity in how condition status/message change over time (compare to Option B)
  - Stops status flip-flopping startup behavior
- Cons:
  - Does not allow for condition-specific behavior within one plugin use
  - Ignoring initial status emitting logic would deviate from behavior in `systemlogmonitor` although could be updated there as well if perceived benefit
### Option B: Allow user to specify ConditionStatus in the condition configuration
- Example:
  ```json
  "plugin": "custom",
  "source": "custom-health",
  "conditions": [{
      "type": "KubeletUnhealthy",
      "reason": "KubeletIsHealthy",
      "message": "kubelet on the node is functioning properly",
      "status": "Unknown"
  }]
  ```
-  Pros:
   - Better behavior of flip-flopping statuses during unhealthy node events False <-> Unknown
   - More granular customization of condition-specific startup behavior
-  Cons:
   - Complexity in incorporating status/message changes over time (compare to Option A)
[...]

PR #646, opened 2022-02-13
Allow skipping condition during customplugin initialization

notchairmk:
Implements Option A from #645 

Allows user to specify `skip_initial_status` in plugin configuration.


Random-Liu:
This fix and reasoning make sense to me.

PR #647, opened 2022-02-19
fix ginkgo version for pull-npd-e2e-test

notchairmk:
- Currently e2e tests are failing for two apparent reasons:
  - ginkgo version mismatch (example seen [here](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/node-problem-detector/646/pull-npd-e2e-test/1492950378952527872))
  - actual e2e failures [e.g. CRI runtime mismatch ](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/node-problem-detector/638/pull-npd-e2e-node/1482077304808017920#1:build-log.txt%3A1060)
- This PR fixes the ginkgo version mismatch by explicitly including ginkgo version used to run e2e tests in go.mod/vendor
  - The sooner we can address ginkgo version mismatch issues the better because this is now bleeding into further e2e test failures since default k/k and test-infra tests are using more recent versions of ginkgo
- Further PRs will be required to get to update [prow jobs](https://github.com/kubernetes/test-infra/blob/7f5c364f8e38b4a4f28bc68f52e5c1257fe43f86/config/jobs/kubernetes/node-problem-detector/node-problem-detector-presubmits.yaml#L272)
- Fixes part of https://github.com/kubernetes/kubernetes/issues/108166
  - There are additional steps that will need to be taken to fix further e2e tests including correct ginkgo version and CRI runtime changes, but this is a starting point
- Updates builder dockerfile to go v1.15

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -221,7 +221,7 @@ test: vet fmt
 	GO111MODULE=on go test -mod vendor -timeout=1m -v -race -short -tags "$(HOST_PLATFORM_BUILD_TAGS)" ./...
 
 e2e-test: vet fmt build-tar
-	GO111MODULE=on ginkgo -nodes=$(PARALLEL) -mod vendor -timeout=10m -v -tags "$(HOST_PLATFORM_BUILD_TAGS)" -stream \
+	GO111MODULE=on go run github.com/onsi/ginkgo/ginkgo -nodes=$(PARALLEL) -mod vendor -timeout=10m -v -tags "$(HOST_PLATFORM_BUILD_TAGS)" -stream \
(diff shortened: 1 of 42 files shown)


notchairmk:
image

notchairmk:
this test was removed in [this test-infra PR](https://github.com/kubernetes/test-infra/pull/25542). 

it would maybe be nice to get e2e tests working without having to upgrade ginkgo, but above PR is moving away from that

vteratipally:
/hold cancel

Thank you for the PR, the test is now fixed and we are waiting on a PR to add back the tests to presubmit again

PR #657, opened 2022-03-25
remove installing ginkgo using "go get" 

vteratipally:
Remove install ginkgo using go get as the current supported version is 2.0 and 2.0 still has some conflicting changes. We can rely on the ginkgo binary compiled in kubernetes repo from the current version.

Fixes: https://github.com/kubernetes/kubernetes/issues/108856


vteratipally:
https://prow.k8s.io/pr-history/?org=kubernetes&repo=node-problem-detector&pr=657 tests are passing now

Issue #661, opened 2022-03-29
Is the forcible heartbeat mechanism necessary?


lianghao208:
We found `heartbeatPeriod` will periodically update node conditions even there is no update event of nodes.
```go
// needHeartbeat checks whether a forcible heartbeat is needed.
func (c *conditionManager) needHeartbeat() bool {
	return c.clock.Since(c.latestTry) >= c.heartbeatPeriod
}
```
In a large cluster(over 3000 nodes), this might cause high cpu and network traffic usage on node in every `heartbeatPeriod`
<img width="1354" alt="image" src="https://user-images.githubusercontent.com/14137033/160616250-3b501fda-90d3-48de-bc55-aff97ef7a713.png">

Is this forcible heartbeat mechanism necessary when we already have update mechanism once nodes' status change?

```go
// needUpdates checks whether there are recent updates.
func (c *conditionManager) needUpdates() bool {
	c.Lock()
	defer c.Unlock()
	needUpdate := false
	for t, update := range c.updates {
		if !reflect.DeepEqual(c.conditions[t], update) {
			needUpdate = true
			c.conditions[t] = update
		}
		delete(c.updates, t)
	}
	return needUpdate
}
```

Random-Liu:
>Is this forcible heartbeat mechanism necessary when we already have update mechanism once nodes' status change?

The force heartbeat is to prevent something changes the conditions NPD owns by mistake.

The heartbeat period is configurable.

PR #662, opened 2022-03-30
feat: add random initial time duration for heartbeat ticker

lianghao208:
Add random initial time duration for heartbeat ticker to avoid massive node conditions update at the same time in large clusters.

In a large cluster, NPD pods are likely to create at almost the same time, which means all the NPD pods will trigger the heartbeat sync at the same time in a certain period. This might cause cpu/network traffic peak glitch.

This problem can be solved by starting heartbeat ticker after a random time duration sleep.

FIX: #661

### DIFF
--- a/pkg/exporters/k8sexporter/condition/manager.go
+++ b/pkg/exporters/k8sexporter/condition/manager.go
@@ -17,6 +17,7 @@ limitations under the License.
 package condition
 
 import (
+	"math/rand"
 	"reflect"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


Random-Liu:
NPD is just one of the node daemons talk with apiserver.

Other daemons could cause similar spike IIUC, most daemons talk/sync with apiserver periodically.

You probably want to avoid deploying NPD onto hundreds of nodes at the same time, and roll it out gradually instead.

Issue #667, opened 2022-04-26
log failed rules at higher log level


grosser:
https://github.com/kubernetes/node-problem-detector/blob/1e8008bdedbeae39074c93cfe3fcdad7735f4db1/pkg/custompluginmonitor/plugin/plugin.go#L118

should be 2 (or 1) when exit status is not 0

PR #668, opened 2022-04-26
show failed statuses as warning

grosser:
fixes https://github.com/kubernetes/node-problem-detector/issues/667


grosser:
this seems to not work, I did not see the failures when running with -v=2 🤷

grosser (commented):
double checked and this seems fine ... the glog.Level converts the int level

grosser:
@vteratipally good to go ?

Issue #659, opened 2022-03-26
when the network interfaces come and go metric cardinality grows unbounded


SergeyKanzhelev:
Calico creates network interfaces for each pod with the unique name. Collecting net metrics (like `kubernetes.io/internal/node/guest/net/tx_packets`) uses the interface name as a dimension `interface_name`. OpenCensus go library has no bounds for the metric cardinality and doesn't allow to clean up old values. 

Need to find a best way to reset label values for net interfaces that don't exist any longer.

mmiranda96:
For cases such as Calico, do we want to keep metrics for each individual interface? If not, a viable alternative would be to group these interfaces within the same dimension. We can use a config model like this:
```json
"net": { 
  "interfaceGroupings": {
    [
      "name": "calico",
      "groupRegexp": "^cali[0-9a-f]+$"
    ]
  }
}
```

In the implementation, we would aggregate all interfaces that match the regex under the same name. Interfaces that don't match any grouping will be reported independently.

SergeyKanzhelev:
I'm not sure how useful this metric is. If we want to keep it - the best long term solution would be to switch to OpenTelemetry delta metrics and use this exporter: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/tree/main/exporter/metric.

@dashpole do you think OpenTelemetry is in a good state to adopt it here for metrics?

dashpole:
The Metrics SDK is currently being (entirely) rewritten in preparation for the first stable release, so I wouldn't recommend it right now.  That effort is tracked in https://github.com/orgs/open-telemetry/projects/22.  After that is published, we will adapt the GCM exporter and publish a stable release of that as well.

mmiranda96:
This issue has been fixed via https://github.com/kubernetes/node-problem-detector/pull/675. This fix has been released in https://github.com/kubernetes/node-problem-detector/releases/tag/v0.8.11.

/remove-lifecycle stale
/close

PR #675, opened 2022-06-03
Add ExcludeInterfaceRegexp to Net Dev monitor

mmiranda96:
## What does this PR do?
As a solution for #659, this PR adds field `ExcludeInterfaceRegexp` to network device monitor. This field is used to stop tracking metrics for certain network interfaces. 

This change also edits config `net-cgroup-system-stats-monitor.json` to exclude Calico interfaces (according to [this doc](https://projectcalico.docs.tigera.io/getting-started/kubernetes/quickstart?gclid=CjwKCAjw46CVBhB1EiwAgy6M4iLRp-n6jcNQycdqjpEOZapTChXAi3en-x9XdIBgQzT2wa2gnBqDpRoCUGwQAvD_BwE), `cali` and `tunl` interfaces) and `veth`, which are interfaces created dynamically and only add noise to the metrics.

Additionally, I refactored the code in order to make it more testable and added some unit tests.

## How to test this change?
1. Create a Kubernetes cluster that uses Calico (I personally used GKE and followed instructions [here](https://cloud.google.com/kubernetes-engine/docs/how-to/network-policy)). Ensure the cluster uses the new NPD version and config (in my case, I manually copied the binaries and config to the node).
2. Run some workload, trigger new network interfaces creation, and ensure metrics are not being reported for Calico interfaces.

### DIFF
--- a/pkg/systemstatsmonitor/net_collector_test.go
+++ b/pkg/systemstatsmonitor/net_collector_test.go
@@ -1,0 +1,190 @@
+package systemstatsmonitor
+
+import (
+	"io/ioutil"
+	"os"
+	"path"
+	"regexp"
+	"testing"
+
+	ssmtypes "k8s.io/node-problem-detector/pkg/systemstatsmonitor/types"
+	"k8s.io/node-problem-detector/pkg/util/metrics"
+)
+
+var defaultMetricsConfig = map[string]ssmtypes.MetricConfig{
+	string(metrics.NetDevRxBytes):      ssmtypes.MetricConfig{DisplayName: string(metrics.NetDevRxBytes)},
+	string(metrics.NetDevRxPackets):    ssmtypes.MetricConfig{DisplayName: string(metrics.NetDevRxPackets)},
+	string(metrics.NetDevRxErrors):     ssmtypes.MetricConfig{DisplayName: string(metrics.NetDevRxErrors)},
+	string(metrics.NetDevRxDropped):    ssmtypes.MetricConfig{DisplayName: string(metrics.NetDevRxDropped)},
+	string(metrics.NetDevRxFifo):       ssmtypes.MetricConfig{DisplayName: string(metrics.NetDevRxFifo)},
+	string(metrics.NetDevRxFrame):      ssmtypes.MetricConfig{DisplayName: string(metrics.NetDevRxFrame)},
+	string(metrics.NetDevRxCompressed): ssmtypes.MetricConfig{DisplayName: string(metrics.NetDevRxCompressed)},
+	string(metrics.NetDevRxMulticast):  ssmtypes.MetricConfig{DisplayName: string(metrics.NetDevRxMulticast)},
+	string(metrics.NetDevTxBytes):      ssmtypes.MetricConfig{DisplayName: string(metrics.NetDevTxBytes)},
+	string(metrics.NetDevTxPackets):    ssmtypes.MetricConfig{DisplayName: string(metrics.NetDevTxPackets)},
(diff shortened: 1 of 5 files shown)


pkg/systemstatsmonitor/net_collector_test.go
```diff
@@ -0,0 +1,189 @@
...
+						t.Fatalf("Failed to convert metric %s to fakeMetric", string(metricID))
+					}
+					for _, repr := range fakeInt64Metric.ListMetrics() {
+						interfaceName, ok := repr.Labels[interfaceNameLabel]
+						if !ok {
+							t.Fatalf("Failed to get label %q for ", interfaceNameLabel)
+						}
+						expectedValue, ok := interfaceValues[interfaceName]
+						if !ok {
+
```
bobbypage, line 172: t.Fatal?

pkg/systemstatsmonitor/net_collector_test.go
```diff
@@ -0,0 +1,189 @@
...
```
bobbypage, line 133: same as below, is it intended for this case to be empty?

pkg/systemstatsmonitor/net_collector_test.go
```diff
@@ -0,0 +1,189 @@
...
+	string(metrics.NetDevTxErrors):     ssmtypes.MetricConfig{DisplayName: string(metrics.NetDevTxErrors)},
+	string(metrics.NetDevTxDropped):    ssmtypes.MetricConfig{DisplayName: string(metrics.NetDevTxDropped)},
+	string(metrics.NetDevTxFifo):       ssmtypes.MetricConfig{DisplayName: string(metrics.NetDevTxFifo)},
+	string(metrics.NetDevTxCollisions): ssmtypes.MetricConfig{DisplayName: string(metrics.NetDevTxCollisions)},
+	string(metrics.NetDevTxCarrier):    ssmtypes.MetricConfig{DisplayName: string(metrics.NetDevTxCarrier)},
+	string(metrics.NetDevTxCompressed): ssmtypes.MetricConfig{DisplayName: string(metrics.NetDevTxCompressed)},
+}
+
+// docker:	1500	100		8		7		0		0		0		0		9000	450		565		200		20		30		0		0
+const fakeNetProcContent = `Inter-|   Receive                                                |  Transmit
```
bobbypage, line 34: can you leave a comment on how you got this data? i.e. what command I should run to get something like this on my machine

bobbypage:
Change LGTM, a few small nits about the test. Thanks for adding the the exclude feature and test coverage, it's a big improvement!

pkg/systemstatsmonitor/net_collector_test.go
mmiranda96: Done, added a comment.

pkg/systemstatsmonitor/net_collector_test.go
mmiranda96: Good catch! It looks like I forgot to update this. Fixed.

pkg/systemstatsmonitor/net_collector_test.go
mmiranda96: Fixed.

mmiranda96:
Addressed your comments @bobbypage. Thanks for the review!

bobbypage:
Thanks

/lgtm

/assign @Random-Liu

Random-Liu:
@vteratipally Could you take a look at this change? Thanks!

vteratipally:
@mmiranda96 Did you happen to check if adding the filter improving the time reporting the metrics

mmiranda96:
@vteratipally I didn't check that specifically, I just ensured that metrics are properly filtered.

PR #678, opened 2022-06-08
Add condition message to event message

2rs2ts:
If you're using some monitoring solution that aggregates events from your Kubernetes cluster, (e.g., Datadog,) having the underlying reason why a condition triggered could be very useful, especially if you are using custom plugin monitors.

Obviously, the fact that a `message` field contains the word "message" in it is a little awkward, but I didn't know what else to call it, besides something that seemed redundant like "condition message." I'm totally open to suggestions here, as long as the message contents get included somehow.

I could not find a documented upper bound on the size of the `message` field in `core.v1.Event`, so I didn't attempt to truncate the message here. Please let me know if I should do so anyway.

### DIFF
--- a/pkg/util/helpers.go
+++ b/pkg/util/helpers.go
@@ -23,12 +23,12 @@ import (
 )
 
 // GenerateConditionChangeEvent generates an event for condition change.
-func GenerateConditionChangeEvent(t string, status types.ConditionStatus, reason string, timestamp time.Time) types.Event {
+func GenerateConditionChangeEvent(t string, status types.ConditionStatus, reason, message string, timestamp time.Time) types.Event {
 	return types.Event{
 		Severity:  types.Info,
 		Timestamp: timestamp,
 		Reason:    reason,
-		Message:   fmt.Sprintf("Node condition %s is now: %s, reason: %s", t, status, reason),
+		Message:   fmt.Sprintf("Node condition %s is now: %s, reason: %s, message: %s", t, status, reason, message),
 	}
 }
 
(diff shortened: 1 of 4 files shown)


pkg/util/helpers.go
```diff
@@ -23,12 +23,12 @@ import (
...
 
 // GenerateConditionChangeEvent generates an event for condition change.
-func GenerateConditionChangeEvent(t string, status types.ConditionStatus, reason string, timestamp time.Time) types.Event {
+func GenerateConditionChangeEvent(t string, status types.ConditionStatus, reason, message string, timestamp time.Time) types.Event {
 	return types.Event{
 		Severity:  types.Info,
 		Timestamp: timestamp,
 		Reason:    reason,
-		Message:   fmt.Sprintf("Node condition %s is now: %s, reason: %s", t, status, reason),
+		Message:   fmt.Sprintf("Node condition %s is now: %s, reason: %s, message: %s", t, status, reason, message),
```
2rs2ts, line 31: As mentioned in the PR summary, I'm totally open to changing the wording and formatting here, just let me know. I would also like to know if this message should be truncated.

mmiranda96: I suggest using `%q` instead of `%s`, since the message might include special render characters (such as \n) and displaying the value makes sense.

mmiranda96 (commented):
Left one small comment. Thanks for this contribution!

2rs2ts:
Thank you for the helpful suggestion, @mmiranda96. Looks like the tests are all good now.

PR #679, opened 2022-06-17
Use Warn severity on K8s Event when Node condition is True

2rs2ts:
If temporary errors generate an Event with a Warn severity, then surely
permanent errors should generate an Event with at least that high of a
severity level.


mmiranda96:
The change sounds reasonable to me, but I'm not sure if this can cause issues. I'd suggest to get one more review.
/ok-to-test
/lgtm

Issue #691, opened 2022-08-03
Update go version for node-problem-detector


jason1028kr:
Node Problem Detector currently uses go v1.15. 
We are seeing cases where it's being flagged by certain security tools (e.g. Palo Alto’s Prisma Compute (Twistlock)).

I propose we upgrade npd's go version (at least 1.18?). We will need to sort out possible dependency issues if we go forward.

mmiranda96:
As of v0.8.13 we're using Go 1.20. Closing this issue now.
/remove lifecycle/rotten
/close

PR #693, opened 2022-08-15
update go directive to 1.18

jason1028kr:
Issue: https://github.com/kubernetes/node-problem-detector/issues/691

Test: `make test` on `golang:1.17.7` container.

Let me know, if there are other dependencies I need to consider.


jason1028kr:
@vteratipally looks like `go mod vendor` removes the vendored ginkgo/ginkgo folder completely, leading to `make e2e-test` failure. Probably it's not picking up the make file dependency on it.

This also happens in current master. Is there a good way to resolve this?

PR #702, opened 2022-09-15
output stdout and stderr from custom commands, allow setting timeout for critctl

karlhungus:
### Problems

 - We are seeing intermittent failures on our custom command for containerd, errors from `crictl` appear to be printed on stderr; since only stdout is output it's difficult to debug what the failure is.

 - `crictl` at [one point](https://github.com/kubernetes-sigs/cri-tools/commit/f9bb11cc0b3993a7078b98fa559bf93ac12155e5) updated the tools timeout from 10s to 2s; it's possible this timeout is the source of our intermittent errors

### Solution

 - output stderr as well as stdout
 - allow setting critctl timeout via npd configs

### DIFF
--- a/cmd/healthchecker/options/options.go
+++ b/cmd/healthchecker/options/options.go
@@ -39,6 +39,7 @@ type HealthCheckerOptions struct {
 	EnableRepair       bool
 	CriCtlPath         string
 	CriSocketPath      string
+	CriTimeout         time.Duration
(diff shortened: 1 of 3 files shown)


karlhungus:
/assign @xueweiz

doing ^ because the bot told me to

karlhungus:
Is there anybody out there?

Issue #313, opened 2019-07-15
allow custom run intervals to improve node boot time and reduce runtime load


grosser:
# Problem
when the node first boots the status conditions change rapidly, so we want to check frequently
when a node is running for a while, the status change rarely, so we want to save overhead

# Solution
`invoke_interval: ["30s", "60s", "60s", "60s", "2m", "2m", "5m"]`

Random-Liu:
Thanks for the proposal!

I think this is a valid request. @xueweiz @wangzhen127 

We can add some kind of backoff mechanism for the interval, something like:
```
initial_invoke_interval: 30s
max_invoke_interval: 5m
backoff_factor: 2.0
```

And the interval should be reset whenever there is a state change:
1) When NPD startup, interval should start from `30s`, and eventually extend to `5m` if no problem is found.
2) When a problem is detected, interval is set back to `30s` to frequently check whether problem is gone, and eventually extend to `5m` if the problem resides.
3) When a problem is gone, interval is set back to `30s` to frequently check whether problem comes back again, and eventually extend to `5m` if the problem never comes back.

grosser:
I like the initial+max, maybe just reuse the current interval to mean max and add "initial_interval" to make migration simple ... backoff_factor can just be a sane default at first, few will need it anyway

grosser:
@Random-Liu can I get a review on https://github.com/kubernetes/node-problem-detector/pull/367 which was my first PR to get going ... but so far nobody showed up :(

PR #712, opened 2022-10-17
add initial-invoke-internal to speed up checks after booting a host

grosser:
fixes https://github.com/kubernetes/node-problem-detector/issues/313


grosser:
we did not end up using this yet, so closing until then

PR #716, opened 2022-10-19
remove "Start watching journald" to avoid plugin log spam

grosser:
atm we see lots of this, would be nice to not have it 🤷 
```
Start logs from plugin  ...
I1019 21:25:24.665373     571 log_watcher.go:80] Start watching journald","pattern":"plugin-logs"}
End logs from plugin ...
```
(if it's not there then the output is empty and the "logs from plugin" goes away)

I know this can be avoided by setting log-level to <=2 but it's pretty verbose even for info level "hey this plugin did nothing" 🤷

### DIFF
--- a/pkg/systemlogmonitor/logwatchers/journald/log_watcher.go
+++ b/pkg/systemlogmonitor/logwatchers/journald/log_watcher.go
@@ -1,3 +1,4 @@
+//go:build journald
 // +build journald
 
 /*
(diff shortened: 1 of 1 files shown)


vteratipally:
> I know this can be avoided by setting log-level to <=2 but it's pretty verbose even for info level

Makes sense. Thank you for the PR

grosser:
rebased to get tests passing, needs another lgtm

PR #717, opened 2022-10-20
health-checker cri: fix invalid command

grosser:
atm produces this, note the missing space between the 2 options
```
--timeout=2s--runtime-endpoint=unix:///var/run/containerd/containerd.sock path
```

broke [here](https://github.com/kubernetes/node-problem-detector/pull/702)
@karlhungus 
@vteratipally


karlhungus:
😻 Thanks! @vteratipally I missed a space in https://github.com/kubernetes/node-problem-detector/pull/702, this should probably make it in (also it improves readability).

vteratipally:
Thank you for the fix.

grosser:
rebased to get tests passing, needs another lgtm

PR #718, opened 2022-10-20
update go.mod to 1.17

grosser:
Dockerfile is on 1.17 so local development should also happen in 1.17
ran `go mod tidy` + `go mod vendor` inside the docker image to get a clean go.sum
... idk why it's moving so many things around, if anyone has a cleaner way of doing this plz lmk 🤷


vteratipally:
lets update to 1.19

grosser:
dockerfile is on 1.17 so prefer to keep that in sync
also this PR is broken and I don't think I have the go knowledge or tools to fix it (working on mac where journald fails)

PR #733, opened 2023-02-01
optionally read node and port information from env variables for kube* services

raghu-nandan-bs:
# Enable users to optionally provide node's IP and kubelet, kube-proxy ports.

If users do not wish to run node-problem-detector pod in host network mode, or if the kubelet / kube-proxy's ports are modified from its default value for some reason, we should be able to support configuring endpoint accordingly for performing health checks.

### DIFF
--- a/pkg/healthchecker/types/types.go
+++ b/pkg/healthchecker/types/types.go
@@ -44,7 +44,7 @@ const (
 	kubeletPortKey          = "KUBELET_PORT"
 	kubeProxyPortKey        = "KUBEPROXY_PORT"
 
-	defaultHost          = "127.0.0.1"
+	defaultHostAddress   = "127.0.0.1"
 	defaultKubeletPort   = "10248"
 	defaultKubeproxyPort = "10256"
 )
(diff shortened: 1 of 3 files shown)


pkg/healthchecker/types/types.go
```diff
@@ -38,12 +39,51 @@ const (
...
+	nodeEnvKey       = "HOST_IP"
+	kubeletPortKey   = "KUBELET_PORT"
+	kubeProxyPortKey = "KUBEPROXY_PORT"
+
+	defaultHostIP        = "127.0.0.1"
+	defaultKubeletPort   = "10248"
+	defaultKubeproxyPort = "10256"
+)
+
+var (
```
mmiranda96, line 53: Since these values are now public variables, they can be overwritten by other means. The only way I can think of preventing this is to have private strings and public functions that return those private values. Eg.
```go
var kubeletHealthCheckEndpoint string
// Assign the value via setKubeEndpoints

func KubeletHealthCheckEndpoint() string { return kubeletHealthCheckEndpoint }
```

mmiranda96 (commented):
Left a couple of comments. One more ask: could you add some unit tests for this new logic in https://github.com/kubernetes/node-problem-detector/blob/master/pkg/healthchecker/types/types_test.go?

mmiranda96:
BTW, you might find https://pkg.go.dev/testing#T.Setenv useful for unit testing OS environment variables.

mmiranda96:
/lgtm

Thanks for your contribution!

PR #748, opened 2023-04-13
Split proc default and validation between Linux and Windows

### LINKED ISSUE #747: Failed to validate windows-system-stats-monitor.json configuration: ProcPath /proc check failed
I noticed this issue when I was running NPD on Windows: when trying to run `systemstatsmonitor`, I get the following failure:
```
Failed to validate C:\node-problem-detector\config\windows-system-stats-monitor.json configuration [BIG JSON]: ProcPath /proc check failed: CreateFile /proc: The system cannot find the file specified.
```

After digging for a while, I found that this failure comes from https://github.com/kubernetes/node-problem-detector/blob/6e57ca6e6c0135ff4a73788e28563476ed9c9334/pkg/systemstatsmonitor/types/config.go#L138
It seems that we try to check the proc file on all platforms, and this is a problem because it doesn't apply for Windows.

This change was introduced in https://github.com/kubernetes/node-problem-detector/pull/594.

There are a couple of fixes I can think of:
- Entirely removing the check: simplest fix, could cause an issue if the user provides a wrong directory.
- Disabling the default `/proc` value (https://github.com/kubernetes/node-problem-detector/blob/6e57ca6e6c0135ff4a73788e28563476ed9c9334/pkg/systemstatsmonitor/types/config.go#L111) and skipping the check if it's empty: will fail if user doesn't explicitly provide a directory.
- Defining a specific value to skip validation (something like "undefined"): depending on a magic string is not always a good idea, and this value is kind of arbitrary.
- Splitting validation logic between Windows and Linux: requires splitting logic but is the cleanest option.

mmiranda96:
#### What does this PR do?

Split procPath default and validation for Windows and Linux.

Linux should default to `/proc` and validate by checking the file exists, while Windows should not have a default nor validation, as this is not supported.

Fixes #747.

### DIFF
--- a/pkg/systemstatsmonitor/types/config.go
+++ b/pkg/systemstatsmonitor/types/config.go
@@ -18,7 +18,6 @@ package types
 
 import (
 	"fmt"
-	"os"
 	"regexp"
 	"time"
 )
@@ -27,7 +26,6 @@ var (
 	defaultInvokeIntervalString   = (60 * time.Second).String()
 	defaultlsblkTimeoutString     = (5 * time.Second).String()
 	defaultKnownModulesConfigPath = "guestosconfig/known-modules.json"
-	defaultProcPath               = "/proc"
 )
 
 type MetricConfig struct {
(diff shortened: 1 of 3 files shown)


samuelkarp (approved):
LGTM

mmiranda96:
Trying once more...

/retest

ibabou (approved):

PR #758, opened 2023-05-25
add threshold of pattern match on systemlogmonitor

smileusd:
```release-note
This pr to add the feature on systemlogmonitor to trigger the condition only repeat match the rule pattern in a period.
For example I want add a pattern that only occur 3 times in 1 hour can trigger condition. This feature give a template for all systemlogmonitor to configure it.
If not configure them, that would be no change.
```

### DIFF
--- a/pkg/systemlogmonitor/counts_buffer.go
+++ b/pkg/systemlogmonitor/counts_buffer.go
@@ -1,0 +1,107 @@
+/*
+Copyright 2023 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package systemlogmonitor
+
+import (
+	utilclock "code.cloudfoundry.org/clock"
+	"fmt"
+	"github.com/golang/glog"
+	systemlogtypes "k8s.io/node-problem-detector/pkg/systemlogmonitor/types"
+	"time"
+)
+
+const (
+	MaxCountRuleMatched = 128
+	MaxTimeExpired      = time.Hour * 168
+)
+
+// CountRingBuffer is a struct to help store and calculate whether match the count and time period
+// It is a ring buffer that when new log matched this rule , ring buffer add this timestamp in the end, and compare the time and count,
+// decide to report error or not. The details in IsThresholdMatched() function.
+type CountRingBuffer struct {
+	data         []time.Time
+	index        int
+	length       int
+	expirePeriod time.Duration
+	clock        utilclock.Clock
+}
+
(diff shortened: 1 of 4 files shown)


smileusd:
@Random-Liu PTAL

mmiranda96:
/ok-to-test

I will take a look at this PR next week, I'm a bit swamped with other work this one.

smileusd:
@mmiranda96 Thank you trigger e2e. I will fix the test first.

pkg/systemlogmonitor/counts_buffer.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+	return false
+}
+
+func NewCountBuffer(rules []systemlogtypes.Rule) map[string]*CountRingBuffer {
+	if len(rules) == 0 {
+		return nil
+	}
+	countBuffer := make(map[string]*CountRingBuffer)
+	for _, rule := range rules {
```
mmiranda96, line 66: Going with my suggested nested config, we can add a check here:
```go
if rule.CountThresholdConfig == nil {
  continue
}
```

pkg/systemlogmonitor/counts_buffer.go
```diff
@@ -0,0 +1,87 @@
...
+		length:       thd,
+		expirePeriod: ep,
+	}
+}
+
+func (crb *CountRingBuffer) IsMatchthreshold() bool {
+	tempIndex := crb.index
+	crb.index = (crb.index + 1) % crb.length
+    if crb.data[crb.index].IsZero() {
+		crb.data[tempIndex] = time.Now()
```
mmiranda96, line 51: Also, why are we using `Now()`. Shouldn't we use the time when the log entry appeared?

mmiranda96 (changes requested):
Thanks for your contribution. There are several points on improvement on your PR, I have left a couple of comments. Also, could you explain more in detail how your implementation works? Please update this PR description with the details.

smileusd:
@mmiranda96 PTAL

pkg/systemlogmonitor/counts_buffer.go
```diff
@@ -0,0 +1,87 @@
...
+func NewCountRingBuffer(thd int, ep time.Duration) *CountRingBuffer {
+	return &CountRingBuffer{
+		data:         make([]time.Time, thd),
+		index:        0,
+		length:       thd,
+		expirePeriod: ep,
+	}
+}
+
+func (crb *CountRingBuffer) IsMatchthreshold() bool {
```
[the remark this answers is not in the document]
mmiranda96: Also, please rename to rename the function to `IsThresholdMatched`.

pkg/systemlogmonitor/counts_buffer.go
mmiranda96: I'm not sure if the implementation actually depends on "now". But if it does, please consider using a clock instead, that way we can mock it in tests. See https://github.com/kubernetes/node-problem-detector/blob/fd51f17ec1f7f85086556c97b0d5efb40ead0fd6/pkg/systemlogmonitor/logwatchers/filelog/log_watcher.go#L65 for an example.

pkg/systemlogmonitor/counts_buffer.go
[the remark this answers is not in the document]
mmiranda96: Please add a comment explaining how this algorithm work.

mmiranda96 (changes requested):
Please address the new and previous comments. Also, please go a bit more in depth on how the feature works, why it's needed, and what are the valid use cases.

PR #765, opened 2023-06-10
remove useless log

smileusd:


smileusd:
@mmiranda96 PTAL

mmiranda96:
I still don't understand what's the need for this change. Is it related to #758? Please ensure your feature works as intended, this test should not change since the new behavior is opt-in.

PR #761, opened 2023-06-01
Bump k8s dependecies to v1.25

yaron-idan:
This PR bumps k8s dependencies to v1.25 to address CVE-2020-8558 and close issue #742

PR #771, opened 2023-06-26
bump k8s.io to 1.24.15 and replace boskos client

MartinForReal:
add context. Context 
replace boskos client in test-infra with sigs.k8s.io/boskos
bump k8s.io to 1.24.15


MartinForReal:
related: #761

vteratipally:
> add context. Context
Any specific reason for splitting the context

MartinForReal:
because context is required in new k8s client

mmiranda96:
Can we split this PR into 2 or more? Adding context can be done before upgrading dependencies. And I think we can do the same with the Boskos client (correct me if I'm wrong on this one).

vteratipally:
> Can we split this PR into 2 or more? Adding context can be done before upgrading dependencies. And I think we can do the same with the Boskos client (correct me if I'm wrong on this one).

Yes,  I agree with Mike. It will be clear if we split the PR in two parts.

MartinForReal:
I've opened two pr. And I will bump the dependencies after these two pr has been merged. 
Could you please take a look? Thanks! @mmiranda96 @vteratipally

PR #772, opened 2023-06-29
Upgrade dependencies to latest

mmiranda96:
Upgrade a couple of dependencies. I tried to update the most independent ones. See `go.mod` for details on upgrades.

/assign @vteratipally


MartinForReal:
/lgtm

I will bump k8s.io dependencies after #774 and #773 are merged. We can put these changes into one pr if you agree.

PR #777, opened 2023-07-04
Fix Makefile for macOS pt. 2

btiernay:
This PR moves `bin/health-checker` to the `BINARIES_LINUX_ONLY` variable in the Makefile so that builds can complete successfully on macOS.


btiernay:
Turns out this is only required for my other branch, apologies!

PR #782, opened 2023-07-11
Switch to project k8s-staging-npd in test script

mmiranda96:
### What does this PR do?
Update test script to use `k8s-staging-npd` project. This project:
- Is owned by kubernetes.io (in contrast to `node-problem-detector-staging`, which is owned by Google)
- Has an existing bucket and image registry
- Is already used in the Makefile

I also added some quotes in the script (lint fixes).

/assign @vteratipally 
/cc @BenTheElder

### DIFF
--- a/test/build.sh
+++ b/test/build.sh
@@ -22,8 +22,8 @@ set -o nounset
 set -o pipefail
 
 
-NPD_STAGING_PATH=${NPD_STAGING_PATH:-"gs://node-problem-detector-staging"}
-NPD_STAGING_REGISTRY=${NPD_STAGING_REGISTRY:-"gcr.io/node-problem-detector-staging"}
+NPD_STAGING_PATH=${NPD_STAGING_PATH:-"gs://k8s-staging-npd"}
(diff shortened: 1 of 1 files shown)


wangzhen127:
@mmiranda96 Do you still plan to fix this?

test/build.sh
```diff
@@ -22,8 +22,8 @@ set -o nounset
 set -o pipefail
 
 
-NPD_STAGING_PATH=${NPD_STAGING_PATH:-"gs://node-problem-detector-staging"}
-NPD_STAGING_REGISTRY=${NPD_STAGING_REGISTRY:-"gcr.io/node-problem-detector-staging"}
+NPD_STAGING_PATH=${NPD_STAGING_PATH:-"gs://k8s-staging-npd"}
```
BenTheElder, line 25: we should fetch from github, not GCS

nobody should be fetching from GCS, it's not cost effective. sometimes we will have GCS behind a CDN, or better yet use github's infra

BenTheElder: oh I see, this is only for testing the project itself, that's OK.

mmiranda96:
Most likely I won't. Feel free to clone this PR.

/close

Issue #797, opened 2023-08-30
Exec format error on arm64


mycrEEpy:
When using the image `registry.k8s.io/node-problem-detector/node-problem-detector:v0.8.14` on arm64 nodes the pod crash loops with the following error: `exec /node-problem-detector: exec format error`

Pod manifest:
```yaml
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: "2023-08-30T11:10:36Z"
  generateName: node-problem-detector-
  labels:
    app: node-problem-detector
    controller-revision-hash: 757d568cb6
    pod-template-generation: "2"
  name: node-problem-detector-7z8x7
  namespace: kube-system
  ownerReferences:
  - apiVersion: apps/v1
    blockOwnerDeletion: true
    controller: true
    kind: DaemonSet
    name: node-problem-detector
    uid: 513debd7-5cc9-435a-b56e-9834abbe0601
  resourceVersion: "21761903"
  uid: 64b29b28-ea07-450f-ad85-befbce85e5ec
spec:
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchFields:
          - key: metadata.name
            operator: In
            values:
            - REDACTED
  containers:
  - command:
    - /node-problem-detector
    - --logtostderr
    - --enable-k8s-exporter
    - --prometheus-address=0.0.0.0
    - --prometheus-port=20257
    - --config.system-log-monitor=/config/kernel-monitor.json
    env:
    - name: NODE_NAME
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: spec.nodeName
    image: registry.k8s.io/node-problem-detector/node-problem-detector:v0.8.14
    imagePullPolicy: IfNotPresent
    name: node-problem-detector
    resources:
      limits:
        cpu: 100m
        memory: 80Mi
      requests:
        cpu: 10m
        memory: 80Mi
    securityContext:
      privileged: true
    terminationMessagePath: /dev/termination-log
    terminationMessagePolicy: File
    volumeMounts:
    - mountPath: /var/log
      name: log
      readOnly: true
    - mountPath: /dev/kmsg
      name: kmsg
      readOnly: true
    - mountPath: /etc/localtime
      name: localtime
      readOnly: true
    - mountPath: /config
      name: config
      readOnly: true
    - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
[...]

Issue #800, opened 2023-08-31
CVE-2022-3821 for NPD due to debian-base:bullseye-v1.4.3


yoliangyoliang:
Vulnerability scanner report:
 [CVE-2022-3821](https://security-tracker.debian.org/tracker/CVE-2022-3821) 
 [CVE-2022-4415](https://security-tracker.debian.org/tracker/CVE-2022-4415)
 [DSA 5417-1](https://lists.debian.org/debian-security-announce/2023/msg00108.html)
on debian-base:bullseye-v1.4.3,

hakman:
Base image was changed to `registry.k8s.io/build-image/debian-base:bookworm-v1.0.0` in https://github.com/kubernetes/node-problem-detector/pull/810.
/close

PR #801, opened 2023-09-01
Update docker build and fix ARM64 image

hakman:
Closes: https://github.com/kubernetes/node-problem-detector/issues/797

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,14 +71,14 @@ else ifeq ($(shell go env GOHOSTOS), windows)
 ENABLE_JOURNALD=0
 endif
 
-# TODO(random-liu): Support different architectures.
-# The debian-base:v2.0.0 image built from kubernetes repository is based on
-# Debian Stretch. It includes systemd 241 with support for both +XZ and +LZ4
-# compression. +LZ4 is needed on some os distros such as COS.
+# Set default base image to Debian 12 (Bookworm)
 BASEIMAGE:=registry.k8s.io/build-image/debian-base:bookworm-v1.0.0
 
 # Disable cgo by default to make the binary statically linked.
 CGO_ENABLED:=0
+
+# Set default Go architecture to AMD64.
+GOARCH ?= amd64
 
 # Construct the "-tags" parameter used by "go build".
 BUILD_TAGS?=
@@ -185,7 +185,7 @@ output/linux_arm64/test/bin/%: $(PKG_SOURCES)
 # In the future these targets should be deprecated.
 ./bin/log-counter: $(PKG_SOURCES)
 ifeq ($(ENABLE_JOURNALD), 1)
-	CGO_ENABLED=$(CGO_ENABLED) GOOS=linux GO111MODULE=on go build \
+	CGO_ENABLED=$(CGO_ENABLED) GOOS=linux GOARCH=$(GOARCH) GO111MODULE=on go build \
 		-mod vendor \
 		-o bin/log-counter \
 		-ldflags '-X $(PKG)/pkg/version.version=$(VERSION)' \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -71,15 +71,15 @@ else ifeq ($(shell go env GOHOSTOS), windows)
 ENABLE_JOURNALD=0
 endif
 
-# TODO(random-liu): Support different architectures.
-# The debian-base:v2.0.0 image built from kubernetes repository is based on
-# Debian Stretch. It includes systemd 241 with support for both +XZ and +LZ4
-# compression. +LZ4 is needed on some os distros such as COS.
-BASEIMAGE:=registry.k8s.io/build-image/debian-base:bullseye-v1.4.3
+# Set default base image to Debian 12.1 (Bookworm)
+BASEIMAGE:=debian:bookworm-20230814-slim@sha256:741bae561f5c2261f4cdd535e4fd4c248dec0aafc1b9a1410b3d67ad24571340
```
pacoxu, line 75: kubernetes/etcd is using `registry.k8s.io/build-image/debian-base:bookworm-v1.0.0` in https://github.com/kubernetes/kubernetes/blob/04292dd94bd9999dfced146f4de8b81a0e466e06/cluster/images/etcd/Makefile#L95

hakman: Is there any advantage to using `registry.k8s.io/build-image/debian-base`?

pacoxu: As I know, we always avoid docker hub as folks will run into rate limits.

hakman: This is just for runtime image, to build things. This is pushed to registry.k8s.io later.

hakman: Also, the etcd image is more or less (to my knowledge) for  internal testing. Not sure if there's any released image based on `registry.k8s.io/build-image/debian-base`.

tzneal: Any reason not to use the same base image?  A few more advantages to using it are that future updates will likely occur at the same time (e.g. someone updating registry.k8s.io/build-image/debian-base:bookworm-v1.0.0 will find all of the occurrences) and less network/disk usage by downstream users due to the shared base image.

hakman: In my opinion, a build image should be used for that and not for final artifacts. In this case, the image contains the Go toolkit and all its dependencies, plus other scripts, that should not be shipped.
For other kube images, there are different base images, see https://github.com/kubernetes/kubernetes/tree/master/build/server-image.
NPD is not tied to a specific K8s release so, the base image will probably be different between various components. 

In the end, `debian:bookworm-slim` is 30MB in size. Performance is not an issue at this size.
Updating the base image should also be easier by not depending on upstream builds and releases.

hakman: @tzneal @pacoxu Any thoughts on the above?

tzneal: My opinion is that moving from a registry.k8s.io hosted base image to a non-registry.k8s.io hosted base image requires a really good reason.

vteratipally: > 

+1 to this.

hakman: Would `registry.k8s.io/build-image/go-runner` be ok instead?
https://github.com/search?q=org%3Akubernetes+registry.k8s.io%2Fbuild-image%2Fgo-runner&type=code

hakman: I tried using `debian-base:bookworm-v1.0.0`, but for some reason tests failed with it and had to revert to `debian-base:bullseye-v1.4.3`. Tests are passing consistently now, but the base image contains know CVEs, as mentioned in https://github.com/kubernetes/node-problem-detector/issues/800.

hakman:
@tzneal @vteratipally Could you take one more look? Thanks!

hakman:
Tests are now passing consistently, especially since https://github.com/kubernetes/test-infra/pull/30710 was merged.

vteratipally:
/lgtm
/approve

Thank you for the fix.

pacoxu:
@vteratipally k/k tried to upgrade to the last release of NPD and reverted due to this issue. Will NPD release a new version in recent days? Or what is the release plan of NPD?

vteratipally:
We can release a new version this week to fix the issue

hakman:
@vteratipally @pacoxu maybe we want https://github.com/kubernetes/node-problem-detector/pull/814 too in the release, to be on the same Go version as k/k.

PR #804, opened 2023-09-07
Move gposutil to v3

mrueg:
Moves another dependency to something that is compatible with go modules.


pkg/systemstatsmonitor/disk_collector.go
```diff
@@ -23,7 +23,7 @@ import (
 	"time"
 
 	"github.com/golang/glog"
-	"github.com/shirou/gopsutil/disk"
+	"github.com/shirou/gopsutil/v3/disk"
```
hakman, line 26: At line 291, this should fix the build:
```go
opttypes := strings.Join(partition.Opts, ",")
```

hakman (changes requested):

pkg/systemstatsmonitor/disk_collector.go
mrueg: Thanks, fixed!

hakman:
Nice, this should remove all warnings when testing on macOS. Thanks!
/lgtm

PR #805, opened 2023-09-08
Remove direct dependency on google/cadvisor@v0.36.0

hakman:
The `tail` functionality was removed from google/cadvisor about 3 years ago by https://github.com/google/cadvisor/pull/2445, as it was no longer used.

This functionality is still used here and makes upgrades difficult.

@dims As one of the people on the cAdvisor Core Team, how do you see this approach. Any suggestions would be appreciated!


dims:
@hakman `third_party` please, like we have in k/k ( https://github.com/kubernetes/kubernetes/tree/master/third_party/forked/golang )

hakman:
Thanks for the feedback @dims. Updated to use `third_party` dir, like in k/k.
/unhold

hakman:
Thanks @vteratipally & @dims!

PR #807, opened 2023-09-08
Update tests to run also on macOS

hakman:
`make test` should also run (partially) on macOS.
This mostly uses functions that already existed in the `*_linux.go` or `*_windows.go` files.


hakman:
Rebased, to fix the conflict with #804.

PR #817, opened 2023-09-19
Remove vendoring for tests

hakman:
I think the dir was added unintentionally in https://github.com/kubernetes/node-problem-detector/pull/806.


MartinForReal:
/lgtm
Thanks!

PR #823, opened 2023-09-24
Update github.com/avast/retry-go to v4.5.0

hakman:


pkg/exporters/stackdriver/stackdriver_exporter.go
```diff
@@ -30,7 +30,7 @@ import (
 	"google.golang.org/api/option"
 	"k8s.io/klog/v2"
 
-	"github.com/avast/retry-go"
+	"github.com/avast/retry-go/v4"
 	"k8s.io/node-problem-detector/pkg/exporters"
```
MartinForReal, line 34: I'm wondering if import statements can be sorted using goimports

hakman: Probably not so hard. There's also a need to verify during automated testing.

PR #824, opened 2023-09-25
Add disk/percent_used and memory/percent_used

kasbert:
Add percentage of disk and memory usage based on the values in disk/bytes_used and memory/bytes_used .

Includes stackdriver configuration with "custom.googleapis.com" as it is not possible to add metrics to "compute.googleapis.com".


hakman:
@kasbert Please sign the CLA and squash and rebase the PR.

PR #825, opened 2023-09-26
Add disk and memory percent_used

kasbert:
Add percentage of disk and memory usage based on the values in disk/bytes_used and memory/bytes_used .

Includes stackdriver configuration with "custom.googleapis.com" as it is not possible to add metrics to "compute.googleapis.com".

Previous PR 824 had a wrong committer.


pkg/systemstatsmonitor/memory_collector_unix.go
```diff
@@ -58,6 +58,29 @@ func (mc *memoryCollector) collect() {
 		}
 	}
 
+	if mc.mPercentUsed != nil && meminfo.MemTotal != nil && *meminfo.MemTotal > 0{
```
mmiranda96, line 61: Why do we need so many percentage memory metrics (and only for Unix)? Shouldn't it just be "used"?

kasbert: That is a copy from mBytesUsed section.
Personally I don't find much use for other than 'used' metrics and I am happy to remove them.

mmiranda96: Yes, please remove it. Thanks!

hakman:
@kasbert Seems unrelated:
```
Oct 31 04:03:58 systemd[1]: Started Node problem detector.
 node-problem-detector[723]: /home/kubernetes/bin/node-problem-detector: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by /home/kubernetes/bin/node-problem-detector)
Oct 31 04:03:58 systemd[1]: node-problem-detector.service: Main process exited, code=exited, status=1/FAILURE
Oct 31 04:03:58 systemd[1]: node-problem-detector.service: Failed with result 'exit-code'.
Oct 31 04:04:08 systemd[1]: node-problem-detector.service: Scheduled restart job, restart counter is at 1.
Oct 31 04:04:08 systemd[1]: Stopped Node problem detector.
```

pkg/systemstatsmonitor/memory_collector_unix.go
kasbert: @mmiranda96 I made the changes, please check.

pkg/systemstatsmonitor/disk_collector.go
```diff
@@ -291,6 +303,9 @@ func (dc *diskCollector) collect() {
 		opttypes := strings.Join(partition.Opts, ",")
 		dc.mBytesUsed.Record(map[string]string{deviceNameLabel: deviceName, fsTypeLabel: fstype, mountOptionLabel: opttypes, stateLabel: "free"}, int64(usageStat.Free))
 		dc.mBytesUsed.Record(map[string]string{deviceNameLabel: deviceName, fsTypeLabel: fstype, mountOptionLabel: opttypes, stateLabel: "used"}, int64(usageStat.Used))
+		if dc.mPercentUsed != nil {
+			dc.mPercentUsed.Record(map[string]string{deviceNameLabel: deviceName, fsTypeLabel: fstype, mountOptionLabel: opttypes, stateLabel: "used"}, float64(usageStat.UsedPercent))
```
mmiranda96, line 307: Wouldn't `dc.mPercentused` always be initialized? If we fail to do so, we fail in line 162, right? We can probably remove the nil check (to match the other metrics).

wangzhen127:
@kasbert This PR has caused us trouble in GKE as the custom metric unexpectedly charged GKE customers. Even though we are switching over to `compute.googleapis.com` in https://github.com/kubernetes/node-problem-detector/pull/1024, I wonder why we need this given we have `bytes_used` for disk and memory. Aren't those good enough already? We are consulting with COS team within Google and we are considering reverting this PR if this is not really needed. So please let me know if you have any use case here. Thanks!

PR #828, opened 2023-10-07
fix: fix pprof by using DefaultServeMux

j4ckstraw:
replace http.NewServeMux with http.DefaultServeMux

see https://pkg.go.dev/net/http/pprof
> By default, all the profiles listed in runtime/pprof.Profile are
available (via Handler), in addition to the Cmdline, Profile, Symbol, and Trace profiles defined in this package. If you are not using DefaultServeMux, you will have to register handlers with the mux you are using.

### DIFF
--- a/pkg/exporters/k8sexporter/k8s_exporter.go
+++ b/pkg/exporters/k8sexporter/k8s_exporter.go
@@ -20,7 +20,7 @@ import (
 	"context"
 	"net"
 	"net/http"
-	_ "net/http/pprof"
+	"net/http/pprof"
 	"strconv"
 
 	"k8s.io/klog/v2"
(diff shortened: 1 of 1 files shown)


MartinForReal:
Do you think if it is a good idea to register the pprof handler with the new mux? @j4ckstraw

j4ckstraw:
> Do you think if it is a good idea to register the pprof handler with the new mux? @j4ckstraw

I think default mux is enough.  

1. Default mux is simple.
2. Maybe default mux have [known performance problem](https://github.com/julienschmidt/go-http-routing-benchmark#conclusions), while this is only for debug, performance is not a primary consideration.

MartinForReal:
It makes sense. But maybe it is better to explicitly define the required handler. And it will be less likely to be broken by package update .

j4ckstraw:
> It makes sense. But maybe it is better to explicitly define the required handler. And it will be less likely to be broken by package update .

OK, I'll file a new patch.

mmiranda96 (approved):
/lgtm /approve

j4ckstraw:
ping~

MartinForReal:
@vteratipally Any advice is appreciated!

j4ckstraw:
ping again~

PR #836, opened 2023-11-07
changing linux package

codemypath:
While trying to build RPM package on centos7 I was getting error:

```
# runtime/cgo
cgo: C compiler "x86_64-linux-gnu-gcc" not found: exec: "x86_64-linux-gnu-gcc": executable file not found in $PATH
make: *** [output/linux_amd64/bin/health-checker] Error 2
error: Bad exit status from /var/tmp/rpm-tmp.qNiIAj (%build)
```

One of the solution for this is to change `x86_64-linux-gnu-gcc` with `x86_64-redhat-linux-gcc`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@
 all: build
 
 # PLATFORMS is the set of OS_ARCH that NPD can build against.
-LINUX_PLATFORMS=linux_amd64 linux_arm64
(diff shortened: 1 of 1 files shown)


codemypath:
Created PR by accident on this repo instead of forked one.

PR #838, opened 2023-11-17
bump k8s.io to 0.28.4

MartinForReal:
bump k8s.io to 0.28.4


go.mod
```diff
@@ -1,46 +1,48 @@
 module k8s.io/node-problem-detector
 
-go 1.20
+go 1.21
+
+toolchain go1.21.4
```
hakman, line 5: I think this can be removed.

test/go.mod
```diff
@@ -1,64 +1,65 @@
 module k8s.io/node-problem-detector/test
 
-go 1.20
+go 1.21
+
+toolchain go1.21.4
```
hakman, line 5: I think this can also be removed.

go.mod
MartinForReal: k8s.io/test-infra@v0.0.0-20231117021824-70d582860a11 requires go@1.21 Do you think if we should bump to go 1.21?

hakman: That would probably be best.

MartinForReal: I'll keep the change because it is required..

hakman: I am curious, why is it required?

MartinForReal: I'm not sure if it is added by mistake.
https://github.com/kubernetes/test-infra/blob/59f3faaf5f0494159e6c56c04882db53563e2707/go.mod#L213

hakman: we need to bump to 1.21 for sure, just not have the toolchain directive (I think I was not very clear).

MartinForReal: Oh. I see.

MartinForReal: Thanks for the review! It has been removed.

test/go.mod
MartinForReal: it has been removed.

hakman:
One more thing @MartinForReal, I don't think we want to add the vendor folder for the test dir. Could you remove?

MartinForReal:
hi @vteratipally Do you have any advice on this? Thanks!

vteratipally:
updated to v0.29.0
/close

PR #841, opened 2023-12-04
Introduce disk/percent_used metric, refs #822

### LINKED ISSUE #822: Disk and memory percent used
Would you consider adding /disk/percent_used and /memory/percent_used ?
I know you can calculate these from other values, but it would be convenient to have them directly.
I hope they would some day be present in Google COS and you could create metrics and alarms based on them.

AndreMiras:
The Stackdriver exports the `/guest/disk/percent_used` metric to the `custom.googleapis.com` namespace as the reserved one `compute.googleapis.com` cannot be used at this stage.

This was tested within GCP Container-Optimized OS with the following:
```
/mnt/disks/scratch/node-problem-detector \
--enable-k8s-exporter=false \
--config.system-stats-monitor=/etc/node_problem_detector/system-stats-monitor.json \
--config.system-log-monitor=/etc/node_problem_detector/kernel-monitor.json \
--config.custom-plugin-monitor=/etc/node_problem_detector/boot-disk-size-consistency-monitor.json \
--exporter.stackdriver=/etc/node_problem_detector/stackdriver-exporter.json
```
The `/mnt/disks/scratch/` directory was mounted specifically to get execution permissions:
```
sudo mount -t tmpfs tmpfs /mnt/disks/scratch/
```

This is what it looks like in GCP Monitoring:
image

### DIFF
--- a/config/system-stats-monitor.json
+++ b/config/system-stats-monitor.json
@@ -44,6 +44,9 @@
       "disk/bytes_used": {
         "displayName": "disk/bytes_used"
       },
+      "disk/percent_used": {
+        "displayName": "disk/percent_used"
+      },
       "disk/io_time": {
         "displayName": "disk/io_time"
       },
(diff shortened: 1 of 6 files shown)


AndreMiras:
superseded by #825

PR #844, opened 2023-12-20
custom iptables version monitor plugin

aojea:
hosts and containers using the host can have different iptables versions, these versions are incompatible and can cause problems if both are present in the kernel.

Add a custom plugin that checks that the iptables rules are only from one version.

The plugin runs every day to avoid causing problems on large systems.

 https://github.com/kubernetes-sigs/iptables-wrappers/blob/97b01f43a8e8db07840fc4b95e833a37c0d36b12/iptables-wrapper-installer.sh


danwinship:
> hosts and containers using the host can have different iptables versions, these versions are incompatible and can cause problems if both are present in the kernel.

"incompatible" and "can cause problems" are overstating things. As long as your rules are self-contained and aren't trying to interact with or override another component's rules, it generally doesn't matter whether you're using the same iptables mode as anyone else. (Unless the system only has kernel support for one of the two modes, but this PR doesn't help people running into that.)

I think we got a little bit confused about this back when iptables-1.8 first came out, because (a) we were also dealing with various bugs in iptables-1.8.[0-2] at the same time, without fully understand what was going on; (b) at the time, kube-proxy depended on iptables chains that were created by kubelet, so kube-proxy and kubelet specifically needed to be using the same iptables mode; (c) kube-proxy purports to poke holes in the system firewall, which can only work if kube-proxy and the firewall both use the same iptables mode.

The bugs eventually got fixed, KEP-3178 made kube-proxy no longer depend on kubelet's rules, and the "poking holes in the system firewall" use case [turns out to be slightly dubious and probably unnecessary](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/3866-nftables-proxy/README.md#forcing-traffic-on-healthchecknodeports-to-be-accepted). (Some distros have moved to `nft`-based firewalls now, which kube-proxy's firewall-hole-poking code won't work with even if you're using `iptables-nft`, and no one has filed any bugs about this.)

So I'm not sure it really makes sense to call this out as a "problem"...

aojea:
well, at least if there are two systems using the different versions they can override each other rules ... I do not know exactly the problems today, but I remember having segfaults because of mixed versions https://bugzilla.netfilter.org/show_bug.cgi?id=1476

We want this plugin to run just to prevent any possible problems , as is impossible to say that there is zero risk

danwinship:
> segfaults because of mixed versions https://bugzilla.netfilter.org/show_bug.cgi?id=1476

That's not "segfaults because of mixed versions", that's "segfaults because of a bug in the iptables binaries that you happened to run into on a node with mixed versions".

danwinship:
> We want this plugin to run just to prevent any possible problems , as is impossible to say that there is zero risk

It seems to me that this is more likely to generate false positives and warn people that there is a problem when everything is actually working fine.

aojea:
> It seems to me that this is more likely to generate false positives and warn people that there is a problem when everything is actually working fine.

interesting, should I add a threshold to the number of lines?
At least for GKE COS clusters I don't expect to have rules from both modes in a node

aojea:
Based on Dan's feedback and npd docs, I'm going to update this to only report events and metrics to get signal, based on the feedback we can assess the criticality of the problem with real data and update to Condition if necessary

> node-problem-detector uses Event and NodeCondition to report problems to apiserver.

> NodeCondition: Permanent problem that makes the node unavailable for pods should be reported as NodeCondition.
> Event: Temporary problem that has limited impact on pod but is informative should be reported as Event.

aojea:
>  #6: Error fetching NPD metrics: error fetching NPD metrics: {prow 35.188.211.187 curl http://localhost:20257/metrics  
    curl: (7) Failed to connect to localhost port 20257 after 0 ms: Couldn't connect to server

/test pull-npd-e2e-test

at first sight in does not look related to my PR

danwinship:
> Based on Dan's feedback and npd docs, I'm going to update this to only report events and metrics to get signal, based on the feedback we can assess the criticality of the problem with real data and update to Condition if necessary

OK. I don't really know much about node-problem-detector...

config/plugin/iptables_mode.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# based on: https://github.com/kubernetes-sigs/iptables-wrappers/blob/97b01f43a8e8db07840fc4b95e833a37c0d36b12/iptables-wrapper-installer.sh
+readonly num_legacy_lines=$( (iptables-legacy-save || true; ip6tables-legacy-save || true) 2>/dev/null | grep -c '^-' || true)
+readonly num_nft_lines=$( (timeout 5 sh -c "iptables-nft-save; ip6tables-nft-save" || true) 2>/dev/null | grep -c '^-' || true)
+
+
+if [ "$num_legacy_lines" -gt 0 ] && [ "$num_nft_lines" -gt 0 ]; then
+  echo "Found rules from both versions, iptables-legacy: ${num_legacy_lines} iptables-nft: ${num_nft_lines}"
+  echo $NONOK
+elif [ "$num_legacy_lines" -gt 0 ] && [ "$num_nft_lines" -eq 0 ]; then
```
MartinForReal, line 21: I'm wondering if the following conditions are necessary. From my understanding, there is no rule found doesn't mean the node is using one backend..

aojea: In a Kubernetes cluster, kubelet always install an iptables rule to solve this specific problem https://github.com/kubernetes/enhancements/tree/master/keps/sig-network/3178-iptables-cleanup#iptables-wrapper

aojea: These mechanism of detecting the backends by counting rules has demonstrated to be the most effective and widely used

PR #856, opened 2024-02-02
fix bug in skip_initial_status

levaspb:
I have found a bug in in the merged PR that introduced the ﻿`skip_initial_status` feature. This bug unjustly leaves conditions stagnant when this feature is enabled in a custom plugin configuration. The root cause is that the conditions for the plugin aren't properly initialized. The PR implies that when `﻿skip_initial_status` is set, the `﻿initialConditions` function is called. As a result, custom plugins with ﻿`skip_initial_status` work correctly.


levaspb:
ping @sjenning for approval

levaspb:
all tests successfully passed
@sjenning please review

a7i:
amazing! tested this and it looks good 👍🏼

PR #864, opened 2024-02-20
Get Node object from APIServer cache

linxiulei:
The risk is pretty low given the returned Node object isn't actually used anywhere - two places invoking GetNode()

1. [k8s_exporter](https://github.com/kubernetes/node-problem-detector/blob/v0.8.15/pkg/exporters/k8sexporter/k8s_exporter.go#L118) calls it to check if readiness but the returned node object is discarded right away
2. [GetConditions](https://github.com/kubernetes/node-problem-detector/blob/v0.8.15/pkg/exporters/k8sexporter/problemclient/problem_client.go#L87), which is not referenced anywhere


hakman:
Thanks @linxiulei!
/lgtm

Issue #874, opened 2024-03-05
Improve release process


wangzhen127:
Currently the release process is very manual. And only @Random-Liu can run `make push`. We should simplify the release process.

CC @Random-Liu @vteratipally @hakman

wangzhen127:
Discussed with @BenTheElder offline. The main pain point of pushing and pulling from gs://kubernetes-release can be removed by just using the github.com file hosting in release notes. Since we already upload tar files in release notes, this would be a simple change.

See https://github.com/kubernetes/kubernetes/pull/123741

hakman:
@wangzhen127 We can also promote binaries, like we do for kOps to the bucket (via k8s.io repo) and just link them in release notes.

wangzhen127:
That's also fine. But using the github hosting file is the simplest way for now.

Does your organization set up NPD as system daemon of daemonset? So far, `cluster/gce/gci/configure.sh` is the only place setting it up as system daemon as far as I know. Given kube-up.sh will switch to kops in the future, and kops decides to treat NPD as addons, it may not worth the effort to set up pipelines to promote binaries.

hakman:
I am fine with images at the moment.
kOps already has NPD as an addon 😄 . See https://github.com/kubernetes/kops/pull/11381.
We should update the NPD tests to run via kOps in the near future.

wangzhen127:
Just realized a few things
- NPD releases tar files to `gs://kubernetes-release/node-problem-detector/` historically. No one has permission to push files to it now. We only have tar files up to `v0.8.10` in the GCS bucket.
- Currently k/k's default NPD version is:
  - k/k 1.28 uses NPD [v0.8.9](https://github.com/kubernetes/kubernetes/blob/v1.28.7/cluster/gce/gci/configure.sh#L29)
  - k/k 1.29 uses NPD [v0.8.13](https://github.com/kubernetes/kubernetes/blob/v1.29.2/cluster/gce/gci/configure.sh#L29)
  - k/k 1.30 uses NPD [v0.8.13](https://github.com/kubernetes/kubernetes/blob/v1.30.0-alpha.3/cluster/gce/gci/configure.sh)
- NPD's standalone mode configuration in k/k depends on the tar files in the above GCS bucket. Given none of k/k's release blocking tests fail, we already lost the test coverage for about half a year (for both k/k 1.29 and 1.30). I filed https://github.com/kubernetes/node-problem-detector/issues/878 for tracking separately.

[...]

PR #883, opened 2024-03-11
Add make release

wangzhen127:
Adds a `make release` command for releasing new NPD version. It stops pushing the tar files to gs://kubernetes-release, because no one has write permission to the GCS bucket any more. We haven't pushed NPD tar files to that GCS bucket after v0.8.10. k/k has been using NPD v0.8.13+ since 1.29. NPD release should just include the tar files in the release note.

Ref issue: https://github.com/kubernetes/node-problem-detector/issues/874

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,8 @@
 .PHONY: all \
         vet fmt version test e2e-test \
         build-binaries build-container build-tar build \
-        docker-builder build-in-docker push-container push-tar push clean depup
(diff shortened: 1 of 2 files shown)


wangzhen127:
/unhold

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/directory/pull-npd-e2e-kubernetes-gce-gci/1767044854166065152 passed.

PR #890, opened 2024-03-28
Add support for CBL-Mariner and Azure Linux

hbeberman:
This PR adds support for `mariner` and `azurelinux` in the getOSVersion helper function.

[Azure Linux](https://github.com/microsoft/azurelinux) (aka CBL-Mariner) is a Linux distribution maintained by Microsoft which is currently [available](https://learn.microsoft.com/en-us/azure/aks/use-azure-linux) in Azure Kubernetes Service and is also used in other parts of the cloud infrastructure and edge products. 

For a few years we've been using a [patched](https://github.com/microsoft/azurelinux/blob/592875ba994cf7fc55449f95bfcb3d26f618a50d/SPECS/node-problem-detector/002-add_mariner_OSVersion.patch) version of node-problem-detector which adds Mariner into this switch case. I'm looking to get "mariner" and the upcoming os-release name of "azurelinux" added into node-problem-detector directly so that we can drop the patch and be compatible with instances of the binary built outside of our distro.

### DIFF
--- a/pkg/util/helpers_linux.go
+++ b/pkg/util/helpers_linux.go
@@ -64,6 +64,10 @@ func getOSVersion(osReleasePath string) (string, error) {
 		return getDebianVersion(osReleaseMap), nil
 	case "sles":
 		return getDebianVersion(osReleaseMap), nil
+	case "mariner":
+		return getDebianVersion(osReleaseMap), nil
+	case "azurelinux":
+		return getDebianVersion(osReleaseMap), nil
 	default:
 		return "", fmt.Errorf("Unsupported ID in /etc/os-release: %q", osReleaseMap["ID"])
 	}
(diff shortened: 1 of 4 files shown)


hakman:
@hbeberman Please sign the CLA, to be able to review & merge this. Thanks!

wangzhen127:
The change looks ok to me. Today we do not have any OSS E2E test coverage for those os types. Do you want to add them? This is not a requirement for this PR. But I think it may be beneficial to you.

hbeberman:
@hakman sorry for the delay! I was making sure everything was squared away to sign the CLA from my side, but I've now signed it.

@wangzhen127 great suggestion on the E2Es. I'm looking into the implementation details and will follow up with a separate contribution if it's something we can add for Mariner / Azure Linux.

PR #897, opened 2024-04-05
Bump NPD versions to v0.8.18

wangzhen127:
/assign @hakman


wangzhen127:
> Could you also update https://github.com/kubernetes/node-problem-detector/blob/master/version.txt?

Updated

PR #909, opened 2024-05-15
Add rocky linux support to GetOSVersion

sudheernv:
**Issue:** Node Problem Detector fails to come up on rocky linux OS. 

**Error Log:**  Failed to retrieve OS version: Unsupported ID in /etc/os-release: "rocky"


mmiranda96 (commented):
/ok-to-test

Wondering if we should just default to `getDebianVersion` if the OS is not COS.

mmiranda96:
Your new unit test is failing.

sudheernv:
> Your new unit test is failing.

Sorry @mmiranda96.  Yesterday, I inadvertently omitted pushing the commit that contained the correct expected string for the newly introduced test case. Pushed new commit with necessary fix now.

-- expectedOSVersion: "8.5 (Green Obsidian)",
++ expectedOSVersion: "rocky 8.5 (Green Obsidian)",

vteratipally (approved):
/lgtm

Issue #920, opened 2024-06-21
Migrate to Community CI By August 1st


BenTheElder:
Node problem detector CI still needs migrating, per https://groups.google.com/a/kubernetes.io/g/dev/c/p6PAML90ZOU

You can find node problem detector jobs listed in:

https://github.com/kubernetes/test-infra/blob/master/docs/job-migration-todo.md

(see e.g. jobs with name containing "npd")

wangzhen127:
The following tests still need migration:
- [CI](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/node-problem-detector/node-problem-detector-ci.yaml): ci-npd-build
- [presubmit](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/node-problem-detector/node-problem-detector-presubmits.yaml): pull-npd-e2e-node, pull-npd-e2e-kubernetes-gce-gci, pull-npd-e2e-kubernetes-gce-gci-custom-flags, pull-npd-e2e-kubernetes-gce-ubuntu, pull-npd-e2e-kubernetes-gce-ubuntu-custom-flags

https://github.com/kubernetes/kubernetes/issues/119211 tracked the issues we encountered previously.

wangzhen127:
I will work with @digitalveer to look into this.

wangzhen127:
There is also similar work tracked in https://github.com/kubernetes/test-infra/issues/31794

BenTheElder:
Yeah, https://github.com/kubernetes/test-infra/issues/31794 was tracking the community migrating it, raised here for visibility that the deadline is approaching.

I spoke with @wangzhen127 and another colleague about this, we'll get this pushing to a community bucket and migrate over.

/assign

https://github.com/kubernetes/node-problem-detector/pull/925 will make the problem a bit simpler.

BenTheElder:
OK so after #925, I spoke to my fellow SIG K8s Infra TLs.

There is a `gs://k8s-staging-npd` already created for NPD in k8s infra, since we're not using it for publishing binaries (thank you!) we can add IAM permission to prow to upload there from the community cluster and switch the scripts over.

That, plus no longer pushing to GCR from CI, should fix everything.


The IAM is managed in https://github.com/kubernetes/k8s.io, can take a look at this later and then we can see about switching buckets.

BenTheElder:
[...]

PR #925, opened 2024-07-08
CI: only push tar instead of also container image

BenTheElder:
related to #920 

AFAICT we don't need to push the image, and it will be tricker to enable pushing to a container registry with GC enabled, instead we can only push the tarball for e2e builds and leave container image pushing to release

### DIFF
--- a/test/build.sh
+++ b/test/build.sh
@@ -128,7 +128,7 @@ function build-pr() {
   export REGISTRY="${NPD_STAGING_REGISTRY}/pr/${PR}"
   export VERSION=$(get-version)
   export TAG="${VERSION}"
-  make push
+  make push-tar
   write-env-file ${PR_ENV_FILENAME}
 }
 
@@ -138,7 +138,10 @@ function build-ci() {
   export REGISTRY="${NPD_STAGING_REGISTRY}/ci"
   export VERSION="$(get-version)-$(date +%Y%m%d.%H%M)"
   export TAG="${VERSION}"
-  make push
+  # e2e tests consume the tarball, not the container
+  # this is simpler to manage in the infra, and we still ensure the container
+  # build works locally
+  make push-tar build-container
 
   # Create the env file with and without custom flags at the same time.
   build-npd-custom-flags


wangzhen127:
/lgtm
/hold

@DigitalVeer Can you revert https://github.com/kubernetes/test-infra/pull/32893 now since we have accumulated enough failure signals?

test/build.sh
```diff
@@ -138,7 +138,7 @@ function build-ci() {
   export REGISTRY="${NPD_STAGING_REGISTRY}/ci"
   export VERSION="$(get-version)-$(date +%Y%m%d.%H%M)"
   export TAG="${VERSION}"
-  make push
+  make push-tar
```
BenTheElder, line 141: /hold

instead of reverting the test-infra change, let's switch to only `push-tar` here but add building but not pushing

wangzhen127: SGTM

test/build.sh
```diff
@@ -128,7 +128,7 @@ function build-pr() {
   export REGISTRY="${NPD_STAGING_REGISTRY}/pr/${PR}"
   export VERSION=$(get-version)
   export TAG="${VERSION}"
-  make push
+  make push-tar
```
BenTheElder, line 131: NOTE: `pull-npd-build` already runs `make build` which includes `build-container`, so there's no reason to build it in e2e runs.

BenTheElder: (specifically it runs `./test/build.sh install-lib && make build`)

BenTheElder:
/hold

Actually, we don't need it here either do we?


What we really want is:
1) presubmit build: ensures we can build, doesn't push. we have that already
2) presubmit e2es: upload and test tarball, don't need to push container (this PR)
3) postsubmit/periodic e2es: either upload and test tarball, or use a pre-uploaded tarball
4) postsubmit/periodic build: ensure we can build container image
5) release process: publishes builds for releases to permanent storage. we have that already.

wangzhen127:
> presubmit build: ensures we can build, doesn't push. we have that already

Yes

> presubmit e2es: upload and test tarball, don't need to push container (this PR)

Yes. And no need to build container either.

> postsubmit/periodic e2es: either upload and test tarball, or use a pre-uploaded tarball

This uses pre-uploaded tarball from postsubmit/periodic build.

> postsubmit/periodic build: ensure we can build container image
release process: publishes builds for releases to permanent storage. we have that already.

We use github file upload for release. We need to build both container and tar file. We only need to push tar file.

BenTheElder:
So right now`ci-npd-build` runs `test/build.sh ci` while presubmit runs `test/build.sh install-lib && make build`

Periodic e2e seem to do:
`./test/build.sh get-ci-env` ...
(like this one https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ci-npd-e2e-kubernetes-gce-ubuntu/1810419258287460352)

Presubmit e2e do: `./test/build.sh -p $(BUILD_ID) pr`

So "CI" mode is only used for `ci-npd-build` and push-tar + build-container makes sense there I think, and that should work on the community infra unless we run into bucket issues.

test/build.sh
wangzhen127: Can you add a comment here to explain that this is only used by presubmit e2e? The build job is using `make build` which covers container build already?

test/build.sh
```diff
@@ -138,7 +138,7 @@ function build-ci() {
   export REGISTRY="${NPD_STAGING_REGISTRY}/ci"
   export VERSION="$(get-version)-$(date +%Y%m%d.%H%M)"
   export TAG="${VERSION}"
-  make push
+  make push-tar build-container
```
wangzhen127, line 141: Maybe add a comment to explain that we do not need to push container here for CI tests.

BenTheElder:
So let's try this as-is, right now the ci-npd-build failure logs just point to the GCR push which we're dropping
/hold cancel
🙃

test/build.sh
BenTheElder: done

PR #938, opened 2024-08-15
Update update-deps.yml with dockerfile update

jingxu97:
This change set up a job to update go version in dockerfile. This only updates 1.22 patch version.

### DIFF
--- a/.github/workflows/update-deps.yml
+++ b/.github/workflows/update-deps.yml
@@ -43,3 +43,38 @@ jobs:
           labels: ok-to-test
           body: |
             Updating go.mod with latest dependencies...
+  update-go-version:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Get current Go version from Dockerfile
+        id: current-go-version
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/update-deps.yml
```diff
@@ -43,3 +43,38 @@ jobs:
           labels: ok-to-test
           body: |
             Updating go.mod with latest dependencies...
+  update-go-version:
```
wangzhen127, line 46: Does this need to be a separate job?

wangzhen127 (commented):
> This only updates 1.22 patch version.

Does this mean golang minor version update is still manual? And when that happens, we also need to change update-deps.yml together?

.github/workflows/update-deps.yml
jingxu97: maybe not, but separate out to avoid one failure affect the other?

jingxu97:
> > This only updates 1.22 patch version.
> 
> Does this mean golang minor version update is still manual? And when that happens, we also need to change update-deps.yml together?

Yes for now. But minor version update is about once a year. So the overhead is small.

PR #943, opened 2024-08-22
Update update-deps.yml to run update dependencies job at 5 pm every T…

jingxu97:
…hursday

Update update-deps.yml to run update dependencies job at 17pm every Thursday


Random-Liu:
Could you squash your commits?

jingxu97:
close it

PR #946, opened 2024-09-13
Update update-deps.yml

jingxu97:
- Remove path, and use find since path is not supported here
- All pull_request so that local test can be triggered. We can remove it later.


hakman:
Pull requests are no longer supported. https://github.com/kubernetes/node-problem-detector/pull/956
/close

Issue #960, opened 2024-09-29
cpu overhead increase 10X when log increase outburst


smileusd:
We find the regexp tool take lots cpu if log flush too much. We make a test for 6400 logs in 1 second to test the performance for simulating the case of process issue to report too much logs. 
```
root@hostname:~# tail  -n 1000000 /var/log/messages  |egrep "Sep 29 01:02:40" |grep "audit"| wc -l 
6400
```
It take 1.95 cpu and 88.8 from regexp execution:
![Screenshot 2024-09-29 at 17 07 30](https://github.com/user-attachments/assets/1b97d270-4a4c-4f58-8bda-2a04e9ca2b01)
![Screenshot 2024-09-29 at 16 41 31](https://github.com/user-attachments/assets/d63d8db4-a808-4b2a-86fb-7932247ccb62)

Then I try to filter the audit log by strings.containe funcition to skip the regexp, it is reduced to 0.15 cpu which reduce exceed 10X:
```Golang
		line = buffer.String()
		buffer.Reset()
		if strings.Contains(line, " audit:") || strings.Contains(line, " audit[") {
			continue
		}
		log, err := s.translator.translate(strings.TrimSuffix(line, "\n"))
		if log == nil {
			if err != nil {
				glog.Warningf("Unable to parse line: %q, %v", line, err)
			}
			continue
		}
```
![Screenshot 2024-09-29 at 15 58 56](https://github.com/user-attachments/assets/c81f9e9b-c19c-4976-b62b-e3e63f6900a7)

So I think we need add a black list to simple fliter the log to skip regexp to protect the npd perforcemance. I will make a pr soon.

PR #961, opened 2024-09-29
add black list to aviod take too much efforts to translate in file log watcher

smileusd:
This pr is to add black list in file log watcher to skip the translat which take too much overhead in cpu when log flush outburst https://github.com/kubernetes/node-problem-detector/issues/960

### DIFF
--- a/pkg/systemlogmonitor/logwatchers/filelog/log_watcher.go
+++ b/pkg/systemlogmonitor/logwatchers/filelog/log_watcher.go
@@ -116,6 +116,9 @@ func (s *filelogWatcher) watchLoop() {
 		}
 		line = buffer.String()
 		buffer.Reset()
+		if s.filterSkipList(line) {
+			continue
+		}
 		log, err := s.translator.translate(strings.TrimSuffix(line, "\n"))
 		if err != nil {
 			klog.Warningf("Unable to parse line: %q, %v", line, err)
@@ -128,4 +131,13 @@ func (s *filelogWatcher) watchLoop() {
 		}
 		s.logCh <- log
 	}
+}
+
+func (s *filelogWatcher) filterSkipList(line string) bool {
+	for _ , skipItem := range s.cfg.SkipList {
+		if strings.Contains(line, skipItem) {
+			return true
+		}
+	}
+	return false
 }
--- a/pkg/systemlogmonitor/logwatchers/types/log_watcher.go
+++ b/pkg/systemlogmonitor/logwatchers/types/log_watcher.go
@@ -36,6 +36,8 @@ type WatcherConfig struct {
 	// PluginConfig is a key/value configuration of a plugin. Valid configurations
 	// are defined in different log watcher plugin.
 	PluginConfig map[string]string `json:"pluginConfig,omitempty"`
+	// Skip the log lines containing any of the strings in the list to avoid running unnecessary regex.
+	SkipList []string `json:"skipList,omitempty"`
 	// LogPath is the path to the log
 	LogPath string `json:"logPath,omitempty"`
 	// Lookback is the time log watcher looks up
--- a/config/disk-log-message-filelog.json
+++ b/config/disk-log-message-filelog.json
@@ -1,0 +1,28 @@
+{
+	"plugin": "filelog",
+	"pluginConfig": {
+		"timestamp": "^.{15}",
+		"message": "(?i)Currently unreadable.*sectors|(?i)Offline uncorrectable sectors",
+		"timestampFormat": "Jan _2 15:04:05"
+	},
+	"logPath": "/var/log/messages",
+	"lookback": "10h",
+	"bufferSize": 1,
+	"source": "disk-monitor",
+	"skipList": [ " audit:", " audit[" ],
+	"conditions": [
+		{
+			"type": "DiskBadBlock",
+			"reason": "DiskBadBlock",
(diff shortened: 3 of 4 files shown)


wangzhen127:
Do you always know what pattern to skip in the log? Can you provide a sample config file with the blacklist for your target use case?

/ok-to-test

pkg/systemlogmonitor/logwatchers/filelog/log_watcher.go
```diff
@@ -129,3 +132,15 @@ func (s *filelogWatcher) watchLoop() {
 		s.logCh <- log
 	}
 }
+
+func (s *filelogWatcher) filterBlackList(line string) bool {
+	if s.cfg.BlackList == nil || len(s.cfg.BlackList) == 0 {
+		return false
+	}
```
hakman, line 139: I don't think this adds much value. Range should be able to handle nil or empty slices.
```suggestion
```

pkg/systemlogmonitor/logwatchers/types/log_watcher.go
```diff
@@ -36,6 +36,8 @@ type WatcherConfig struct {
 	// PluginConfig is a key/value configuration of a plugin. Valid configurations
 	// are defined in different log watcher plugin.
 	PluginConfig map[string]string `json:"pluginConfig,omitempty"`
+	// Skip the black list by simple filter to avoid running more efforts in regex
+	BlackList []string `json:"blackList,omitempty"`
```
hakman, line 40:
```suggestion
	// Skip the log lines containing any of the strings in the list to avoid running unnecessary regex.
	SkipList []string `json:"skipList,omitempty"`
```

hakman (changes requested):

smileusd:
> Do you always know what pattern to skip in the log? Can you provide a sample config file with the blacklist for your target use case?
> 
> /ok-to-test

In our cluster, the users deploy the audit beat pod and record the log with all files changed in /var/log/messages. So we skip log simply contains the "audit:" and "audit[" .

pkg/systemlogmonitor/logwatchers/filelog/log_watcher.go
smileusd: done

pkg/systemlogmonitor/logwatchers/types/log_watcher.go
smileusd: done

pkg/systemlogmonitor/logwatchers/filelog/log_watcher.go
```diff
@@ -116,6 +116,9 @@ func (s *filelogWatcher) watchLoop() {
 		}
 		line = buffer.String()
 		buffer.Reset()
+		if s.filterSkipList(line) {
+			continue
```
wangzhen127, line 120: Can you add some test for this?

pkg/systemlogmonitor/logwatchers/types/log_watcher.go
```diff
@@ -36,6 +36,8 @@ type WatcherConfig struct {
 	// PluginConfig is a key/value configuration of a plugin. Valid configurations
 	// are defined in different log watcher plugin.
 	PluginConfig map[string]string `json:"pluginConfig,omitempty"`
+	// Skip the log lines containing any of the strings in the list to avoid running unnecessary regex.
+	SkipList []string `json:"skipList,omitempty"`
```
wangzhen127, line 40: Can you add a sample config file in https://github.com/kubernetes/node-problem-detector/tree/master/config to illustrate how to use this?

pkg/systemlogmonitor/logwatchers/filelog/log_watcher.go
smileusd: done

pkg/systemlogmonitor/logwatchers/types/log_watcher.go
smileusd: done

config/docker-monitor-filelog.json
```diff
@@ -9,6 +9,7 @@
 	"lookback": "5m",
 	"bufferSize": 10,
 	"source": "docker-monitor",
+	"skipList": [],
```
wangzhen127, line 12: No need to add empty list in existing files. We need to make sure when it is missing, those files still work for backward compatibility.

config/kernel-monitor-filelog.json
```diff
@@ -9,6 +9,7 @@
 	"lookback": "5m",
 	"bufferSize": 10,
 	"source": "kernel-monitor",
+	"skipList": [ " audit:", " audit[" ],
```
wangzhen127, line 12: Can you remove from this file and add a new json file with this?

wangzhen127 (commented):
Can you also squash the commits?

wangzhen127:
BTW, we plan to make a new release this week. If you can get the PR ready in a day or two, we can include it in the upcoming release. Otherwise, it can wait for the next release.

/retest

config/kernel-monitor-filelog.json
smileusd: done

config/docker-monitor-filelog.json
smileusd: done

PR #980, opened 2024-10-27
feature: add QPS Burst flags for client cfg

googs1025:
- add QPS Burst flags for client cfg

### DIFF
--- a/pkg/exporters/k8sexporter/problemclient/problem_client.go
+++ b/pkg/exporters/k8sexporter/problemclient/problem_client.go
@@ -74,7 +74,8 @@ func NewClientOrDie(npdo *options.NodeProblemDetectorOptions) Client {
 	}
 
 	cfg.UserAgent = fmt.Sprintf("%s/%s", filepath.Base(os.Args[0]), version.Version())
-	// TODO(random-liu): Set QPS Limit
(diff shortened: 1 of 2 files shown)


pkg/exporters/k8sexporter/problemclient/problem_client.go
```diff
@@ -74,7 +74,8 @@ func NewClientOrDie(npdo *options.NodeProblemDetectorOptions) Client {
 	}
 
 	cfg.UserAgent = fmt.Sprintf("%s/%s", filepath.Base(os.Args[0]), version.Version())
-	// TODO(random-liu): Set QPS Limit
+	cfg.QPS = float32(npdo.QPS)
```
andyxning, line 77: can we just make `npdo.QPS` as `float32`?

googs1025: thanks for reply. :)
The flag package currently only seems to provide float64 types. If you want to directly support float32, you need to implement the interface yourself or use another package. For these reasons, I think simple type conversion is more concise.
FYI: https://pkg.go.dev/flag#Float64

andyxning: goti t.

PR #983, opened 2024-11-06
chore: qps flag: use float32 instead of float64

googs1025:
/kind cleanup

- use float32 instead of float64


googs1025:
/assign @andyxning 
sorry for this.....
Awkward... I made a small mistake, using fs has float32 methods...
FYI: https://github.com/kubernetes/node-problem-detector/pull/980#discussion_r1819982150

PR #999, opened 2024-12-26
chore: refactor custom plugin monitor method

googs1025:
- refactor custom plugin monitor method


pkg/custompluginmonitor/custom_plugin_monitor.go
```diff
@@ -296,18 +296,22 @@ func toConditionStatus(s cpmtypes.Status) types.ConditionStatus {
 	}
 }
 
-// initializeStatus initializes the internal condition and also reports it to the node problem detector.
-func (c *customPluginMonitor) initializeStatus() {
```
googs1025, line 300: `initializeStatus`  naming seems confusing, and the `initialConditions` does not need to be repeated

googs1025:
friendly ping @andyxning

PR #1009, opened 2025-01-10
feat(kmsgLogWatcher): add logger to parser

daveoy:
another example of potential internal application health enhancements to satisfy #1006


daveoy:
@wangzhen127 looks like the tests passed here, are we good to request member review?

PR #1021, opened 2025-02-04
feat(k8sExporter): Options to allow disabling Events or Node Conditions

nickbp:
Both outputs are currently hardcoded to being enabled, this allows disabling one or the other. Defaults to both enabled to retain current behavior.

Larger clusters can save some etcd I/O by skipping one of these outputs if they aren't being consumed. In our case we aren't consuming the Events so writing them just creates more churn, for example if node problems are flapping.


daveoy (approved):
🚀  looks good to me, seems useful

PR #1022, opened 2025-02-04
feat(k8sexporter/problemclient): allow deletion of deprecated condition types

### LINKED ISSUE #1011: Clearing Old Conditions
we have observed that when changing our system log monitor configurations to omit a previously watched condition, the condition persists on the node object.

I have added a bool flag `--delete-deprecated-conditions` and stringSliceFlag `--deprecated-condition-types`, plus a handler into the `k8sexporter` that will delete conditions from a node object on exporter initialization. 

would this community be interested in a PR that supplies this feature?

daveoy:
this PR adds two flags,

1. enable deprecated condition type deletion (bool, defaults to false)
2. a CSV-string of condition type names to delete (string, ignored if 1 is false)

if enabled and given a string, on problemclient initialization we loop over the .status.conditions field of our node and recreate the list without the types present in the new switch (2) outlined above. we `update` .status.conditions with the new list (patching doesn't work here)

closes #1011

### DIFF
--- a/cmd/options/options.go
+++ b/cmd/options/options.go
@@ -88,6 +88,9 @@ type NodeProblemDetectorOptions struct {
 	MonitorConfigPaths types.ProblemDaemonConfigPathMap
 
 	// application options
+	// DeleteDeprecatedConditions is the flag determining whether to delete deprecated Conditions and Events from the node status.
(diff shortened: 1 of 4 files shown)


daveoy:
@wangzhen127 can I get an ok-to-test?

Issue #945, opened 2024-09-12
Non matching lines logged as error. 


nikhil-bhat:
In  [filelog](https://github.com/kubernetes/node-problem-detector/blob/master/pkg/systemlogmonitor/logwatchers/filelog/log_watcher.go#L121) lines that don`t match the regex are logged as error with level warning.

This results in lot of logs generated from the npd daemonset.
Shouldn`t the log level in the case be debug instead of warning

hakman:
@nikhil-bhat Could you add a little more context here, maybe a sample from a log file? Thanks!

PR #1032, opened 2025-02-27
Eliminate logspam using filelog monitor

skaven81:
When using a filelog monitor, the top-level `pluginConfig.message` filter provides a limited stream of log events to be checked against the `rules[].pattern`.  I thus expect it to be normal for most log events to fail to match the top-level `pluginConfig.message` filter. Such a condition should not trigger a warning-level log message.  And in fact such log messages should be suppressed by default, unless `-v=5` or higher is used for troubleshooting/debug purposes.

### DIFF
--- a/pkg/systemlogmonitor/logwatchers/filelog/log_watcher.go
+++ b/pkg/systemlogmonitor/logwatchers/filelog/log_watcher.go
@@ -121,7 +121,7 @@ func (s *filelogWatcher) watchLoop() {
 		}
 		log, err := s.translator.translate(strings.TrimSuffix(line, "\n"))
 		if err != nil {
-			klog.Warningf("Unable to parse line: %q, %v", line, err)
+			klog.V(5).Infof("Unable to parse line: %q, %v", line, err)
 			continue
 		}
 		// Discard messages before start time.


hakman:
@wangzhen127 I am not sure if `-v=5` the exact value, but I agree with the approach.
Ref: https://github.com/kubernetes/node-problem-detector/issues/945
/lgtm

pkg/systemlogmonitor/logwatchers/filelog/log_watcher.go
```diff
@@ -121,7 +121,7 @@ func (s *filelogWatcher) watchLoop() {
 		}
 		log, err := s.translator.translate(strings.TrimSuffix(line, "\n"))
 		if err != nil {
-			klog.Warningf("Unable to parse line: %q, %v", line, err)
+			klog.V(5).Infof("Unable to parse line: %q, %v", line, err)
```
wangzhen127, line 124: There are 2 types of errors: timestamp parsing error and message filtering error. This hides both errors. But I think you only want to hide the message filtering error, right? https://github.com/kubernetes/node-problem-detector/blob/master/pkg/systemlogmonitor/logwatchers/filelog/translator.go#L59

I think in the original design, we only considered the pattern where the message regular expression is common in the file log. This is true for kernel log for example. Can you share more of the use case here? It's probably valid and we may want to log here instead of erroring out: https://github.com/kubernetes/node-problem-detector/blob/master/pkg/systemlogmonitor/logwatchers/filelog/translator.go#L74

hakman:
CC @nikhil-bhat

pkg/systemlogmonitor/logwatchers/filelog/log_watcher.go
skaven81: > Can you share more of the use case here? 

I'll update the introductory comment on the PR with the use cases where this became a problem.

skaven81: See also #1038

PR #1037, opened 2025-03-11
Removed draino from README

guettli:
The draino project was not updated for 5 years.


kundan2707 (commented):
it should be removed as no updates since long time

PR #1045, opened 2025-04-02
imagePullPolicy doesn't need to be set to Always

kokes:
Since images are pinned to specific versions, setting `imagePullPolicy` to `Always` is needlessly wasteful. Changing to the default `IfNotPresent` instead.


SergeyKanzhelev (approved):
/lgtm
/approve

PR #1057, opened 2025-05-19
fix: Prevent custom plugin monitor from stucking when plugin process hangs

### LINKED ISSUE #1050: util.kill hang
The customer plugin may end up in the D (uninterruptible) state. Killing it will lead to a hang, and no more rules will be executed.

Adding timeout check for util.kill is possible ？

jessehu:
## Problem
If the plugin execution timeout is caused by the plugin process or sub-process hung in D state due to GPU device errors or other reasons, wg.Wait() will be blocked forever, so we need to add a timeout to the wait group. This PR improves the current timeout handling logic.

Fix https://github.com/kubernetes/node-problem-detector/issues/1050.

## Test
We verified this PR in local env to simulate a process hung in D state via reading from a suspended disk. The code returns cpmtypes.Unknown as expected instead of blocked forever.

### DIFF
--- a/pkg/custompluginmonitor/plugin/plugin.go
+++ b/pkg/custompluginmonitor/plugin/plugin.go
@@ -17,7 +17,6 @@ limitations under the License.
 package plugin
 
 import (
-	"context"
 	"fmt"
 	"io"
 	"os/exec"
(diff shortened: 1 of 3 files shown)


jessehu:
hi @hakman @wangzhen127 @googs1025 could you please take a look at this PR ? Thanks. 👍

SergeyKanzhelev:
Any chance you can create a test for this?

PR #1061, opened 2025-05-30
feat!: Set default host address value to `localhost`.

### LINKED ISSUE #1060: NPD kubelet check fails when kubelet healthzBindAddress is `::1`
With IPv6 cluster setup it's kind of expected that kubelet will expose healthcheck on IPv6 address. Te default value for HOST is right now `127.0.0.1` which requires setting the `HOST_ADDRESS` env.

This still might be problematic when cluster is dual stack IPv4/IPv6 and some monitored resources expose healthcheck on IPv4 and some IPv6. The workaround right now is to use `HOST_ADDRESS=localhost` which then use resolver to translate `localhost` to both `127.0.0.1` and `::1` and check will succeed regardless if target service is listening on ipv4 or ipv6 local address.

The proposal is to change the current defaultHostAddress from `127.0.0.1` to `localhost`.

marqc:
Usage of `localhost` is family agnostic and will work regardless if cluster is IPv4 or IPv6. The current value of `127.0.0.1` only works for IPv4 clusters.

BREAKING CHANGE: It may break in rare cases where `localhost` does not resolve as `127.0.0.1` (if OS config does not follow RFC5735 and RFC6761).

Fixes #1060.

### DIFF
--- a/pkg/healthchecker/types/types.go
+++ b/pkg/healthchecker/types/types.go
@@ -18,6 +18,7 @@ package types
 
 import (
 	"fmt"
+	"net"
 	"os"
 	"sort"
 	"strconv"
@@ -44,7 +45,7 @@ const (
 	kubeletPortKey          = "KUBELET_PORT"
 	kubeProxyPortKey        = "KUBEPROXY_PORT"
 
-	defaultHostAddress   = "127.0.0.1"
+	defaultHostAddress   = "localhost"
 	defaultKubeletPort   = "10248"
 	defaultKubeproxyPort = "10256"
 )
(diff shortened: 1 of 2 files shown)


pkg/healthchecker/types/types.go
```diff
@@ -44,7 +44,7 @@ const (
 	kubeletPortKey          = "KUBELET_PORT"
 	kubeProxyPortKey        = "KUBEPROXY_PORT"
 
-	defaultHostAddress   = "127.0.0.1"
+	defaultHostAddress   = "localhost"
```
aojea, line 47: there is a bug in line 81 and 82

```
	kubeletHealthCheckEndpoint = fmt.Sprintf("http://%s:%s/healthz", hostAddress, kubeletPort)
	kubeProxyHealthCheckEndpoint = fmt.Sprintf("http://%s:%s/healthz", hostAddress, kubeProxyPort)

```

that url should be build using `net.JoinHostPort` to support ipv6, do you mind fixing it?

marqc: Fixed

marqc:
e2e test failures are most likely unrelated. They doesn't look like the result of this change.

wangzhen127:
/retest

@marqc can you fix the invalid commit message issue?

PR #1066, opened 2025-06-04
update sig node approvers to the current list

SergeyKanzhelev:
/assign @wangzhen127


SergeyKanzhelev:
keeping reviewers as-is for now.

PR #1090, opened 2025-08-10
Fix e2e test

hakman:


hakman:
Experiment done. https://github.com/kubernetes/node-problem-detector/pull/1091 and https://github.com/kubernetes/test-infra/pull/35293 should fix all failing tests.

Issue #1093, opened 2025-08-11
Branch for 1.34


SergeyKanzhelev:
Let's discuss the 1.34 branch.

The current diff from 8.2.1: https://github.com/kubernetes/node-problem-detector/compare/v0.8.21...master is a single bug fix for IPv6, few deps updates.

I suggest we ensure the test green. I hope two recent PRs were merged.

I also feel it is a good idea to bump all dependencies to latest before forking and fix CVE:

- https://github.com/kubernetes/node-problem-detector/pull/1089
- https://github.com/kubernetes/node-problem-detector/issues/1085

Anything else should we take for 1.34?

It may be a good idea to pick up the 1.34 deps. Similar to https://github.com/kubernetes-sigs/cri-tools/pull/1884 we can start by picking up RC and later just bump to 0.34 when available and release the version at that time

@hakman @wangzhen127 @daveoy

hakman:
we can also wait until 1.34 is GA and branch after bumping the deps.

SergeyKanzhelev:
yes. Starting with rc can catch any potential issues. Making it easier to do the final step once k8s GAs the 1.34

SergeyKanzhelev:
@hakman what is left for branch creation?

hakman:
@SergeyKanzhelev I think we should merge https://github.com/kubernetes/node-problem-detector/pull/1120 first.
Before cutting new releases, the `debian-base` image that we use should be rebuilt, see https://github.com/kubernetes/release/issues/4118.

wangzhen127:
Should we cut the new release now?

hakman:
yep, I will go do that today.

hakman:
The branch was created and tag PRs created.

hakman:
First release is published: https://github.com/kubernetes/node-problem-detector/releases/tag/v1.34.0

PR #1113, opened 2025-09-08
upgraded reference to k8s 1.34

SergeyKanzhelev:
Related to https://github.com/kubernetes/node-problem-detector/issues/1093


SergeyKanzhelev:
Note, now we have three yaml implementaiton references =). One of them is being held by opencensus that will be eliminated in other workstream

PR #1123, opened 2025-09-11
Update Cloud Build config

hakman:
This should allow us to use the images that are built by `node-problem-detector-push-images` and pushed to `gcr.io/k8s-staging-npd`.
This requires #1122 and https://github.com/kubernetes/test-infra/pull/35512.

/cc @SergeyKanzhelev @wangzhen127 @ameukam


hakman:
Should work ok now, the remaining prerequisite is just nice to have.
/unhold

PR #1141, opened 2025-09-26
feat(exporters): replace opencensus with opentelemetry

### LINKED ISSUE #1008: Migrate OpenCensus to OpenTelemetry
OenCensus has been deprecated. We should migrate to OpenTelemetry. See https://opentelemetry.io/blog/2023/sunsetting-opencensus/

daveoy:
should close #1008 

- replace opencensus with opentelemetry in: promethetus exporter, stackdriver exporter and metrics abstraction utilities used by other components
- fake metrics replaced to preserve problem metrics manager stub used for testing

to note:

- daemon and exporter initialization have been reordered since problemmetrics global manager initialization order matters (we need to register all exporters with the global manager before starting the problem daemon -- this has been run on a few clusters in my env and does not affect node-problem-detector
- make test is passing
- make lint is passing
- i have no way of testing the stackdriver / google compute engine metrics refactor, i hope the e2e test provides this?
- systemstatsmonitor, healthchecker etc are not in use in my environment and i don't know if the tests are enough to ensure they're still working too (some common components, i.e. the metrics_int64 and float64 representations might be at play)


daveoy:
@wangzhen127 can i get an /ok-to-test please

daveoy:
e2e test is complaining about the presence of otel's default `go_` namespaced metrics -- we can configure them off they are new/default with otel.

daveoy:
@hakman @wangzhen127 looks like we're ready for review over here

be gentle

daveoy:
Can we get a review?

hakman:
@daveoy Sorry, but I am not very familiar with this area. I can give it a try, but probably @wangzhen127 or someone from Google has better context.
CC @SergeyKanzhelev

wangzhen127:
Thanks for uploading this PR! Sorry for the delay. Got busy in the past week. Will try to get to this within this week.

SergeyKanzhelev:
@dashpole can you help review or suggest somebody?

SergeyKanzhelev:
> * i have no way of testing the stackdriver / google compute engine metrics refactor, i hope the e2e test provides this?

I would think we should start testing OpenTelemetry metrics with OpenTelemetry collector. It should be reasonably easy to integrate to e2e. @dashpole are there a good examples where OTel colelctor is integrated in e2e tests?

We may also eliminate the Stackdriver portion since Otel protocol must be reasonalbly universal

pkg/util/otel/resource_test.go
```diff
@@ -0,0 +1,52 @@
+/*
+Copyright 2019 The Kubernetes Authors All rights reserved.
```
SergeyKanzhelev, line 2: we may need to eliminate the year completely

wangzhen127:
Adding @rkolchmeyer as well

go.mod
```diff
@@ -70,19 +65,14 @@ require (
...
 	github.com/yusufpapurcu/wmi v1.2.4 // indirect
 	go.opentelemetry.io/auto/sdk v1.1.0 // indirect
 	go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc v0.61.0 // indirect
-	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.61.0 // indirect
-	go.opentelemetry.io/otel v1.36.0 // indirect
-	go.opentelemetry.io/otel/metric v1.36.0 // indirect
-	go.opentelemetry.io/otel/trace v1.36.0 // indirect
+	go.opentelemetry.io/otel/trace v1.37.0 // indirect
 	go.yaml.in/yaml/v2 v2.4.2 // indirect
 	go.yaml.in/yaml/v3 v3.0.4 // indirect
```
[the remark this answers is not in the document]
SergeyKanzhelev: maybe I was thinking about the gopkg.in/yaml

dashpole:
Not sure if you've seen https://opentelemetry.io/docs/specs/otel/compatibility/opencensus/#migration-path, but that might be helpful.

dashpole:
You can test OTLP in an integration test similar to this, but for metrics: https://github.com/kubernetes/kubernetes/blob/9624c3dcdc9a9e1286e8ca32d07d231e69ed2f0c/test/integration/apiserver/tracing/tracing_test.go#L354.  Not sure if integrating the OTel collector will be easier or not.

dashpole:
For unit tests, consider using go.opentelemetry.io/otel/sdk/metric/metricdata/metricdatatest.  The test can set up an SDK with a `ManualReader`, manually call `Collect`, and then make assertions on the resulting metric data.

dashpole:
If you need to write integration tests for google cloud monitoring, you can do something similar to https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/7130d1aded77c51f613a9d949f166d119cd595e9/internal/cloudmock/metrics.go

pkg/util/otel/common.go
```diff
@@ -0,0 +1,92 @@
...
+// GetMeterProvider returns the global meter provider, initializing it if necessary
+func GetMeterProvider() *sdkmetric.MeterProvider {
+	if globalMeterProvider == nil {
+		return InitializeMeterProvider()
+	}
+	return globalMeterProvider
+}
+
+// MeterName is the standard meter name used across the application
+const MeterName = "node-problem-detector"
```
dashpole, line 87: We usually recommend that the meter name is the package name where metrics are defined.  But given this is used as an implementation of your metric interface, that might not make sense.

dashpole:
Overall, this looks correct to me.  With better testing of the outputs (ideally done as a separate PR prior to this migration PR), including prometheus and GCM you could do this migration with higher confidence that it isn't breaking anything

daveoy:
thank you all for your comments, keep them coming!

i should mention that this is running in my environment and the problemdaemons and metrics (problem_counter and problem_gauge) are operating as expected.

i'll see if i can mock the Google cloud monitoring stuff for higher confidence on the stackdriver side.

pkg/util/metrics/metric_int64.go
```diff
@@ -36,67 +37,90 @@ type Int64MetricRepresentation struct {
...
-	name    string
-	measure *stats.Int64Measure
+// Type alias added for backward compatibility
+type Int64Metric = OTelInt64Metric
+
+type Int64MetricInterface interface {
+	Record(labelValues map[string]string, value int64) error
 }
 
-// NewInt64Metric create a Int64Metric metric, returns nil when viewName is empty.
```
rkolchmeyer, line 44: It looks like the "returns nil when viewName is empty" behavior wasn't retained - the program exits in this situation in the new implementation. Would it make sense to keep this behavior? Same for the float64 metric.

rkolchmeyer (commented):
It looks like there are issues on GCE:

```
rpc error: code = InvalidArgument desc = One or more TimeSeries could not be written: timeSeries[57] (metric.type="compute.googleapis.com/guest/disk/operation_bytes_count", metric.labels={"direction": "write", "device_name"
: "sda8", "service_name": "node-problem-detector"}): unrecognized metric labels [service_name];
```

and that error repeats for a bunch of other metrics. I guess this PR adds a new "service_name" label that Cloud Monitoring doesn't expect. Is that label necessary for all opentelemetry clients, or can node-problem-detector avoid sending that on GCE?

dashpole:
YOu will want to add the `WithFilteredResourceAttributes(NoAttributes())` option to the google cloud exporter to fix https://github.com/kubernetes/node-problem-detector/pull/1141#pullrequestreview-3300493346

wangzhen127:
@daveoy Do you get any chance to make progress on this PR? Just checking if we could make this into 1.35. :)

wangzhen127:
/remove-lifecycle stale

@daveoy have you got time to move this forward? :)

daveoy:
Yessir, I will pick this up tomorrow and see if I can’t get these last few things closed out by end of next week. Apologies for the delay

daveoy:
working on the rebase now

daveoy:
okay, i big time cheesed up that rebase but i think its better now. waiting on tests

daveoy:
> If you need to write integration tests for google cloud monitoring, you can do something similar to https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/7130d1aded77c51f613a9d949f166d119cd595e9/internal/cloudmock/metrics.go

should this be moved out of internal? or shall i recreate here so we can build on it

dashpole:
feel free to recreate it here

dashpole:
https://cloud.google.com/blog/products/management-tools/otlp-opentelemetry-protocol-for-google-cloud-monitoring-metrics is probably relevant here.  Cloud monitoring now supports OTLP.  The only "special" thing you need to do is use gcp auth with the OTLP grpc exporter ([example](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/tree/main/example/metric/otlpgrpc)).

By default, it writes to google managed prometheus, which is different from what the exporter you currently use exports to (it is cheaper, and writes to a prometheus_target resource).  I'm not sure what you are trying to maintain consistency with, so you might need a bit more config on the OTLP exporter to have exact parity.

daveoy:
i'll let @wangzhen127 weigh in on that -- mostly opencensus+stackdriver stuff to be otel+stackdriver(gcm). happy to explore using OTLP here but wonder if it may cause issues in some customer envs over at GCP?

daveoy:
need some GCP eyes on the failing GCP e2e stuff -- looks like unrelated errors to me

daveoy:
> need some GCP eyes on the failing GCP e2e stuff -- looks like unrelated errors to me

ope, they appear happier now? flaky tests maybe?

daveoy:
not sure if github is having issues but i added a commit with `make gomod` changes to my fork that hasn't arrived here yet

daveoy:
@hakman wonder if you could throw a review on here too

hakman (changes requested):
@daveoy I took a pass at the PR and looks mostly fine, added some suggestions.

Additionally, it seems that the old exporter attached GCEMetadata.Zone, InstanceID, and InstanceName, but the new global OTel resource only contains service.*, so Cloud Monitoring will classify compute.googleapis.com/guest/* metrics under a generic resource instead of the node's gce_instance. Could you take a look, please?

hakman:
#1297 was merged and contains all the changes in this PR.
Thanks again @daveoy for stating and testing this. It was of great help.
/close

PR #1213, opened 2026-01-06
add future time filter for log watchers when time across the year

smileusd:
This pr is to fix this specail case:
1. kernel log file: 
```
root@:~# tail -n 1 /var/log/messages
Jan  6 02:06:53 kubelet[8137]: E0106 02:06:53.560506    8137 summary_sys_containers.go:89] "Failed to get system container stats" err="failed to get cgroup stats for \"/systemd/system.slice\": failed to get container info for \"/systemd/system.slice\": unknown container \"/systemd/system.slice\"" containerName="/systemd/system.slice"
```
2. journal log:
```
root@:~# journalctl -u kubelet -n 1
Jan 06 02:07:53 kubelet[8137]: E0106 02:07:53.698973    8137 summary_sys_containers.go:89] "Failed to get system container stats" err="failed to get cgroup stats for \"/systemd/system.slice\": failed to get co>
```
These logs are not record "year" of time, so when the time from 2025 to 2026, the NPD will think that the log in 2025-12-31 is after than 2026-01-01, so the logs from 2025 will be captured into 2026.

### DIFF
--- a/pkg/systemlogmonitor/logwatchers/filelog/log_watcher.go
+++ b/pkg/systemlogmonitor/logwatchers/filelog/log_watcher.go
@@ -131,6 +131,13 @@ func (s *filelogWatcher) watchLoop() {
 			klog.V(5).Infof("Throwing away msg %q before start time: %v < %v", log.Message, log.Timestamp, s.startTime)
 			continue
 		}
+
+		// Discard messages after now, lots of log files are not record year. 1 min cover log write latency
+		if log.Timestamp.After(time.Now().Add(time.Minute)) {
+			klog.V(5).Infof("Throwing away msg %q after current time: %v < %v", log.Message, log.Timestamp, time.Now())
+			continue
+		}
+
 		s.logCh <- log
 	}
 }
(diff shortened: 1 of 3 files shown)


SergeyKanzhelev:
> These logs are not record "year" of time, so when the time from 2025 to 2026, the NPD will think that the log in 2025-12-31 is after than 2026-01-01, so the logs from 2025 will be captured into 2026.

Can you please clarify what happens? Does NPD think that those Jan logs are from 2025 or that Dec logs are for 2026? Simply discarding events as a solution sounds wrong if we need to process logs - we may loose something interesting there.

smileusd:
> > These logs are not record "year" of time, so when the time from 2025 to 2026, the NPD will think that the log in 2025-12-31 is after than 2026-01-01, so the logs from 2025 will be captured into 2026.
> 
> Can you please clarify what happens? Does NPD think that those Jan logs are from 2025 or that Dec logs are for 2026? Simply discarding events as a solution sounds wrong if we need to process logs - we may loose something interesting there.

Yes, the kernel and journal log not record the year, NPD can not decide the 1.1.2026 is later than 12.31.2025. For example if we set the log loopback is only 3 mintnues, we will recived exceed 3 mintnues ago logs from year of 2025 if we restart NPD after 1.1.2026. 

We have a case is that, we set a pattern, one log at 30.12.2025 match this pattern, we recived the event again at 7.1.2026 because restart NPD.

Issue #474, opened 2020-10-08
Gauge for FilesystemIsReadOnly not downgraded to 0 after fixing the problem


sharonosbourne:
The problem occurred when filesystem went to read only mode. That was fixed, but still in the metrics I was able to see the counter and gauge set up to 1.
I conducted a test and multiple times injected the FileSystemIsReadOnly to the /dev/kmsg (https://github.com/kubernetes/node-problem-detector/blob/master/config/kernel-monitor.json):

1 log_monitor.go:160] New status generated: &{Source:kernel-monitor Events:[{Severity:info Timestamp:2020-10-08 06:44:16.09315274 +0000 UTC m=+1331754.148888064 Reason:FilesystemIsReadOnly Message:Node condition ReadonlyFilesystem is now: True, reason: FilesystemIsReadOnly}] Conditions:[{Type:KernelDeadlock Status:False Transition:2020-09-22 20:48:21.98500453 +0000 UTC m=+0.040739839 Reason:KernelHasNoDeadlock Message:kernel has no deadlock} {Type:ReadonlyFilesystem Status:True Transition:2020-10-08 06:44:16.09315274 +0000 UTC m=+1331754.148888064 Reason:FilesystemIsReadOnly Message:Remounting filesystem read-only}]}

Still the metrics were shown as 1 and it did not downgraded to 0. Even the the issue with ro filesystem was fixed, still the metric was 1:

problem_counter{reason="FilesystemIsReadOnly"} 1
problem_gauge{reason="FilesystemIsReadOnly",type="ReadonlyFilesystem"} 1

As a workaround the pod was deleted and after that metrics were reset to 0.
What is the reason of that behaviour? The type "permanent"? Is deleting a pod the only solution?

kernel-monitor.json

		{
			"type": "permanent",
			"condition": "ReadonlyFilesystem",
			"reason": "FilesystemIsReadOnly",
			"pattern": "Remounting filesystem read-only"
		}

adesaegher:
We are also facing a similar issue but with many occurences as we are extensively using PVC with GCP disks. 
With a ot of mounting/unmounting operations, the kernel catches many readonly disks events (not on the node root disk) and consequently node-problem-detector set the node as not ready. 
We may find a more precise pattern in kernel-monitor.json to only catch root filesystem events.

TaichiHo:
This looks like a long standing bug that is still happening. Any suggestions here? 

/remove-lifecycle rotten

[...]

PR #1234, opened 2026-02-12
ReadonlyFilesystem Condition with Auto-Recovery Detection

CharudathGopal:
ISSUE: https://github.com/kubernetes/node-problem-detector/issues/474

**Problem Statement**
The existing NPD **SystemLogMonitor** for **ReadonlyFilesystem** is a one-way trigger: it sets the condition to True when detecting `Remounting filesystem read-only` in kernel logs, but never clears it. The condition remains stuck until NPD pod restart or node reboot, causing some Kubernetes platform to delete the node.

**Solution: CustomPluginMonitor with Active Recovery Detection**

This PR introduces a CustomPluginMonitor (check_ro_filesystem.sh) that actively monitors both detection and recovery of read-only filesystems.

**Key Features**

1. NPD Design Adherence & Detection (5-minute lookback)

    - Mimics NPD's SystemLogMonitor by scanning `/dev/kmsg` for "Remounting filesystem read-only" messages
    - Uses 5-minute lookback window (300 seconds) calculated from `/proc/uptime`
    - Filters kernel messages by timestamp to match NPD's behavior
    - Extracts device names from kernel messages (e.g., pxd!pxd208624217308167071, dm-3, sda1)
    - Checks if extracted devices are currently mounted read-only in `/host/proc/1/mounts`
    - If any device is RO: Exit 1 → Condition = True

2. Recovery Check (all-time lookback)

    - If no recent errors found, performs second scan of `/dev/kmsg` with no time limit
    - Retrieves ALL historical "remounting filesystem read-only" messages since boot
    - Checks if ANY historically-problematic devices are still mounted read-only
    - If all devices recovered or unmounted: Exit 0 → Condition = False
    - **Enables automatic recovery detection - the key differentiator from SystemLogMonitor**

3. Targeted Approach (Avoids False Positives)

    - Only monitors devices that kernel explicitly reported as having errors
    - Does NOT scan all mounts in `/proc/mounts` (which would flag legitimate read-only mounts like /boot, CD-ROMs, ConfigMaps)
    - More accurate and simpler than filesystem-wide scanning
    - Prevents false positives from intentionally read-only filesystems

**Deployment Changes**

ReadonlyFilesystem CustomPluginMonitor is not enabled by default, users can enable this feature by following steps documented in ` docs/readonly-recovery-plugin-monitor.md` 

  - Adds /host/proc volume mount to access host mount table
  - Replace `--config.system-log-monitor=/config/readonly-monitor.json` with `--config.custom-plugin-monitor=/config/readonly-recovery-plugin-monitor.json`

### DIFF
--- a/config/plugin/check_ro_filesystem.sh
+++ b/config/plugin/check_ro_filesystem.sh
@@ -1,0 +1,278 @@
+#!/bin/sh
+
+# ReadonlyFilesystem detector with auto-recovery capability.
+#
+# Strategy:
+# 1. Find all devices remounted read-only in last 5 minutes (from /dev/kmsg)
+# 2. Check if ANY of those devices are CURRENTLY mounted read-only
+# 3. If ANY device is RO -> set ReadonlyFilesystem=True (exit 1)
+# 4. Get all devices remounted read-only in history (from /dev/kmsg)
+# 5. Check if ALL of those devices are CURRENTLY mounted read-only OR not mounted at all
+# 6. If ALL devices recovered -> set ReadonlyFilesystem=False (exit 0)
+
+
+readonly OK=0 # All devices recovered, clear condition (condition: False)
+readonly NONOK=1 # At least one device still RO (condition: True)
+readonly MOUNTS_FILE_HOST="/host/proc/1/mounts"
+readonly MOUNTS_FILE_LOCAL="/proc/mounts"
+readonly LOOKBACK_SEC=300  # 5 minutes lookback
+
+# Extract device name from kernel message
+# Example: "EXT4-fs (sda1): remounting filesystem read-only"
+# Example: "XFS (dm-3): remounting filesystem read-only"
+extract_device_name() {
+    _msg="$1"
+    # Extract device from parentheses: (device)
+    _device=$(printf '%s\n' "$_msg" | sed -n 's/.*(\([^)]*\)).*/\1/p')
+    if [ -n "$_device" ]; then
+        printf '%s\n' "$_device"
+        return 0
+    fi
+    return 1
+}
+
+# Extract device names from kmsg output
+# Input: kmsg messages (one per line)
+# Output: unique device names (sorted)
+extract_devices_from_messages() {
+    _messages="$1"
+    printf '%s\n' "$_messages" | while IFS= read -r _line; do
+        # Skip empty lines
+        [ -z "$_line" ] && continue
+
+        # Extract device name from the message
+        _dev=$(extract_device_name "$_line")
+        if [ -n "$_dev" ]; then
+            printf '%s\n' "$_dev"
+        fi
+    done | sort -u
+}
+
+# Get all possible device paths for a given device name
+# Handles dm-X devices by resolving symlinks via udevadm
+# Returns: space-separated list of device paths to check
+get_device_paths() {
+    _dev="$1"
+    _paths="$_dev"
+
+    # Add common path prefixes
+    _paths="$_paths /dev/$_dev"
+
+    # For dm-X devices, get symlinks from udevadm
+    case "$_dev" in
+        dm-*)
+            # Try to get symlinks for this dm device
+            if command -v udevadm >/dev/null 2>&1; then
+                _symlinks=$(udevadm info --query=symlink --name="/dev/$_dev" 2>/dev/null || true)
+                if [ -n "$_symlinks" ]; then
+                    # Add each symlink as a potential path
+                    for _link in $_symlinks; do
+                        _paths="$_paths /dev/$_link"
+                    done
+                fi
+            fi
+            ;;
+    esac
+
+    # For devices with ! character (Portworx dm-name format)
+    # Example: pxd!123 → /dev/mapper/pxd!123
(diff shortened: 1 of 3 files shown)


CharudathGopal:
@Random-Liu @sjenning Can you please take a look at this PR

CharudathGopal:
Hello @hakman,

This feature is important for Portworx Enterprise Product which provides a comprehensive, software-defined, persistent container storage and data management platform, specifically designed for Kubernetes-based applications across on-premises, hybrid, and multi-cloud environments.

The changes we are making as part of this PR can help multiple platform like Anthos, Gardener to operate seamlessly with Portworx.

Appreciate any help with reviewing this change.

hakman:
Hi @CharudathGopal — thanks for putting this together, and for the detailed work on it. I went through the PR, and I had to spend some time understanding the problem statement and the operational impact of the change.

Here’s my feedback:

1. **Split ownership of a single condition across monitors**
   Having one monitor set a condition and another clear it can be race-prone in NPD (for a given condition type, the latest update wins). I’d strongly prefer that a *single* monitor owns both the **assert** and **clear** transitions for `ReadonlyFilesystem`, so the behavior is deterministic.

2. **Unreliable signal from `/dev/kmsg`**
   The script treats “no remount messages found” as recovery. Since `/dev/kmsg` is a rotating ring buffer, older remount events can disappear even while the filesystem is still mounted read-only. That could lead to clearing `ReadonlyFilesystem` incorrectly.
   More generally, plugins ideally shouldn’t tail `/dev/kmsg`, and should rely on stronger signals (e.g., explicitly tracking and checking mount state / filesystem status).

3. **Unreadable `/dev/kmsg` reported as healthy**
   If `/dev/kmsg` is unreadable, the script currently returns `OK`, which could clear an active `ReadonlyFilesystem=True` condition without evidence of recovery. I think this should be reported as **Unknown / non-OK** (or otherwise avoid clearing) rather than “healthy”.

4. **Complexity / maintainability (Bash + awk)**
   The plugin logic is fairly large and not trivial to maintain/debug. It might be possible to simplify the logic and reduce the amount of parsing/state handling.

That said, NPD is modular. If this approach is working well for you and your users/customers as-is, I don't see any reason to not use it. I do agree it would be valuable to have this capability upstream, but I think the implementation details need a bit more discussion before we merge something like this.

@wangzhen127 @SergeyKanzhelev — curious if you have thoughts on the right approach here (especially around the signal/source of truth and condition ownership).

CharudathGopal:
Hello @hakman,

Thanks for the detailed review! 

Before answering your questions/concerns I would like to mention I have tried to keep the design as similar to existing SystemLogMonitor (readonly-monitor.json).  

**1. Condition Ownership**
We will explicitly ask customers to disable the existing SystemLogMonitor (readonly-monitor.json) when deploying our CustomPluginMonitor to avoid split ownership. Only one monitor manages the condition at a time, no race condition occurs.

**2. /dev/kmsg Ring Buffer Limitation**
Acknowledged. However, we're following the same principle as the existing readonly-monitor.json, both rely on `/dev/kmsg` as the source of truth. The difference: we add recovery detection on top of the same signal.
The ring buffer issue exists in the current implementation too. Our approach is: if the kernel message disappeared AND the device is no longer mounted RO (or unmounted), it's reasonable to clear the condition. This matches real-world recovery scenarios (Portworx unmounts failed volumes).

Even in case of SystemLogMonitor (readonly-monitor.json), when the POD is restarted and ring-buffer do not have the error message then it resets the node condition to false.

**3. Unreadable /dev/kmsg = OK**
This matches existing NPD behavior. When /dev/kmsg is unreadable:
Existing SystemLogMonitor: kmsgparser.NewParser() fails → Watch() returns error → monitor is skipped → no condition reported
Our plugin: /dev/kmsg unreadable → exit OK → no condition reported
Both approaches avoid false alarms. This is critical because some Kubernetes platforms (e.g., certain cloud providers) delete nodes when ReadonlyFilesystem=True is reported. Better to not report anything than risk node deletion on a monitoring failure.

**4. Complexity**
The complexity stems from handling diverse device naming schemes in production environments:
Portworx uses: pxd!pxd123, dm-X, /dev/mapper/pxd!pxd123, /dev/pxd/pxd123
Device-mapper symlinks require udevadm resolution
WWID-based devices, multipath devices, etc.

**Initial approach:** Scan all /proc/mounts for read-only filesystems. 
Problem: Required extensive device name normalization and still produced false positives (legitimate RO mounts like /boot, ConfigMaps, CD-ROMs).

**Current approach:** Only check devices that kernel explicitly flagged in error messages. More targeted, fewer false positives, but requires robust path resolution.

I 100% agree that this needs a upstream solution, but getting this change in helps a lot of customers and multiple K8s Platform Providers to use this solution.

CharudathGopal:
@hakman FYI, forgot to mention earlier.

The doc I have added already talks [about disabling readonly-monitor.json ](https://github.com/kubernetes/node-problem-detector/pull/1234/changes#diff-b8ca08b7cb761de8cb1f0b8eb9b34f495f0e9c77ee615947426ed84b18d0793dR67)before using the customPluginMonitor we have here in this PR.

@SergeyKanzhelev @wangzhen127 Appreciate if you review the change.

PR #1256, opened 2026-04-01
chore(readme): update chart install instructions

TPXP:
The chart registry moved to github package registry according to the chart readme, this PR updates setup instructions on this repository as well to use `oci://ghcr.io/deliveryhero/helm-charts/node-problem-detector`

Ref: https://github.com/deliveryhero/helm-charts/tree/master/stable/node-problem-detector


hakman (approved):
Thanks @TPXP, this will help many users. 🚀

PR #1266, opened 2026-04-21
custom-plugin-monitor: Honor default condition status

halaney:
Users can define a default status, but that's being stomped on currently. Let's honor what's defined, if defined correctly... otherwise continue to default to False and log a warning.

### DIFF
--- a/pkg/custompluginmonitor/custom_plugin_monitor.go
+++ b/pkg/custompluginmonitor/custom_plugin_monitor.go
@@ -317,7 +317,15 @@ func initialConditions(defaults []types.Condition) []types.Condition {
 	conditions := make([]types.Condition, len(defaults))
 	copy(conditions, defaults)
 	for i := range conditions {
-		conditions[i].Status = types.False
+		switch conditions[i].Status {
+		case types.True, types.False, types.Unknown:
+		case "":
+			conditions[i].Status = types.False
+		default:
+			klog.Warningf("Condition %q has invalid initial status %q, defaulting to %q",
+				conditions[i].Type, conditions[i].Status, types.False)
+			conditions[i].Status = types.False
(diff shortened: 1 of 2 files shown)


halaney:
Hi! Can I get some review on this?

pkg/custompluginmonitor/custom_plugin_monitor.go
```diff
@@ -317,7 +317,13 @@ func initialConditions(defaults []types.Condition) []types.Condition {
...
 	copy(conditions, defaults)
 	for i := range conditions {
-		conditions[i].Status = types.False
+		switch conditions[i].Status {
+		case types.True, types.False, types.Unknown:
+		default:
+			klog.Warningf("Condition %q has invalid initial status %q, defaulting to %q",
+				conditions[i].Type, conditions[i].Status, types.False)
+			conditions[i].Status = types.False
+		}
```
DigitalVeer, line 328: Currently none of the existing [configs/*.json files](https://github.com/kubernetes/node-problem-detector/tree/master/config) with `"plugin": "custom"`  actually set `status` on a condition. This means `Status` would now come in as `""` which doesn't match `True`/`False`/`Unknown`, and by the new logic, go to the default branch and log a warning.

That means each CustomPluginMonitor condition without `status` will log a warning once at startup ( `Condition "NTPProblem" has invalid initial status "", defaulting to "False"`. )

It may be worth honoring the default as such:

```
    case "":
        conditions[i].Status = types.False
```

halaney: hah, duh! thanks I'll make that change. We've been running this (plus my other open PRs) in our setup, but we use our own custom plugins and _do_ set that.

PR #1269, opened 2026-04-22
custom-plugin-monitor: Really skip initial statuses

halaney:
Right now with SkipInitialStatus we skip the initial status update, but it doesn't really achieve the goal of skipping sending initial status values for conditions overall. For example, right now:

    1. We set conditions to their default values internally
    2. We run our checks
    3. The first check returns a Result
    4. This Result is used to updated the conditions slice we hold internally
    5. We send out all the conditions

meaning we are sending the default value out still for most of the conditions.

Instead in this skip initial status config we should only add conditions that have actually reported a result.


halaney:
Hi! Can I get some review on this?

PR #1278, opened 2026-05-08
fix: handle SIGTERM gracefully on unix

halaney:
We weren't catching SIGTERM and cancelling our context, so the go runtime would have us exit with a non-zero code when the signal was received. Let's catch it and try to exit gracefully.

i.e. before:

    (...)
    I0508 11:08:00.426917 1787442 log_monitor.go:237] Initialize condition generated: []
    ^C
    $ echo $?
    130

and now:

    (...)
    I0508 11:07:42.150562 1787146 log_monitor.go:237] Initialize condition generated: []
    ^C
    I0508 11:07:43.372093 1787146 log_monitor.go:123] Stop log monitor /tmp/npd-test.json
    I0508 11:07:43.652876 1787146 log_watcher.go:104] Stop watching filelog
    I0508 11:07:43.652923 1787146 log_monitor.go:144] Log monitor stopped: /tmp/npd-test.json
    $ echo $?
    0

### DIFF
--- a/cmd/nodeproblemdetector/node_problem_detector_unix.go
+++ b/cmd/nodeproblemdetector/node_problem_detector_unix.go
@@ -21,6 +21,9 @@ package main
 import (
 	"context"
 	"flag"
+	"os"
+	"os/signal"
+	"syscall"
 
 	"github.com/spf13/pflag"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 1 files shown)


hakman (approved):
Thank you for your contribution @halaney!

Issue #833, opened 2023-10-25
Feature Request: GPU Support


ZongqiangZhang:
This feature request aims to enhance the Node Problem Detector with the ability to monitor GPUs on nodes and detect issues.

Currently NPD does not have direct visibility into GPUs. However, many workloads are GPU accelerated which makes GPU health an important part of node health. e.g. GPUs are widely used in machine learning training and inference. Especially for LLM training which may using tens of thousands of GPU cards. The entire training cluster should be restarted from previous checkpoint if any one of the GPUs in the cluster is gone bad.

This feature request adds the following capabilities:

- GPU device monitoring: NPD will collect GPU device info periodically and look for crashes or errors via nvidia-smi/nvml/dcgm tools.

- GPU device monitoring: NPD will check GPU device info periodically to detect if a GPU is "stuck" (e.g. nvidia-smi command hangs).

- TBD: GPU runtime monitoring: NPD will check for crashes or OOM issues reported in nvidia logs.

Specifically, this feature request includes:
- Code for the gpu_monitor plugin
- A Dockerfile to build an NPD image with GPU support
- Other dependencies

Looking forward to your feedback!

zz913922:
Exactly the same situation, did you resolve this issue already?

AllenXu93:
I think NPD should support different config for different device or runtime;
For example , I have both GPU and none GPU worker node in one cluster, or containerd and docker nodes in one cluster; currently we need to deploy 2 npd daemonset for different type of node;

AllenXu93:
BTW, for GPU, we don't need to install more dependencies, it just add env `NVIDIA_VISIBLE_DEVICES` and `NVIDIA_DRIVER_CAPABILITIES` , and use `nvidia-smi` command to check GPU state

wangzhen127:
Thanks for filing the feature request! I think this totally makes sense. Do you have any more concrete proposal?

/cc @SergeyKanzhelev

SergeyKanzhelev:
yes, accelerators health is an important functionality and would be great to have it in NPD

Need to design it carefully though. There is already some health checking in a device plugin (like https://github.com/NVIDIA/k8s-device-plugin/blob/bf58cc405af03d864b1502f147815d4c2271ab9a/internal/rm/health.go#L39) that we need to work nicely with. Even simple detection of a device plugin health is a good starting point here.

@AllenXu93 @ZongqiangZhang do you want to work on more detailed design? I definitely will be interested to join the effort

[...]

PR #1281, opened 2026-05-13
Initial cut of external plugin support for NPD

dims:
xref: https://github.com/kubernetes/node-problem-detector/issues/833

As a first step to GPU support, we want to avoid running scripts repeatedly and have something running that NPD can talk to over GRPC. I did some experiments in:
https://github.com/dims/npd-ext

Filing this PR as a first step. I've included a sample `examples/external-plugins/gpu-monitor` that just runs nvidia-smi.

NOTE that the protos are based off the same handshake that external plugins currently use. So this mechanism is NOT just for GPUs in any way. 

Please see `docs/external-plugins.md` for more info.


examples/external-plugins/gpu-monitor/Dockerfile
```diff
@@ -0,0 +1,51 @@
+# Build stage
+FROM golang:1.21-alpine AS builder
```
ArangoGutierrez: 1.26

dims:
cc @wangzhen127 @hakman

hakman:
Tests should be fixed by https://github.com/kubernetes/node-problem-detector/pull/1294.

googs1025 (commented):
first review:  

Thanks for working on this. We have a similar need around external/GPU-style checks, so I’m trying to understand the intended direction here. 😄 

  One question from reading the docs: is this PR meant to cover only external monitors for now, or external exporters as well? I see `ExternalMonitorProxy` is implemented, but the docs/protos also mention `ExternalExporter`. If exporter support is future work, maybe it would help to call that out explicitly.

  I also wonder if the external monitor contract could be made a bit clearer, especially around whether NPD validates returned sources/condition types, and what should happen when the external plugin is unavailable.

  Overall the direction looks useful to me.

dims:
@hakman please take a look?

PR #1285, opened 2026-05-29
fix(logwatchers/journald): back off when sd_journal_wait fails

chchen-x:
sd_journal_wait can return immediately with a negative errno when sd_journal_get_fd is unable to allocate an inotify instance (e.g. when the host has exhausted fs.inotify.max_user_instances). The previous code discarded Wait's return value, so the watchLoop spun at 100% CPU until inotify pressure cleared.

Check the return code and sleep for waitLogTimeout when Wait did not actually wait, preserving the intended 5s polling cadence.

### DIFF
--- a/pkg/systemlogmonitor/logwatchers/journald/log_watcher.go
+++ b/pkg/systemlogmonitor/logwatchers/journald/log_watcher.go
@@ -114,7 +114,16 @@ func (j *journaldWatcher) watchLoop() {
 		}
 		// If next reaches the end, wait for waitLogTimeout.
 		if n == 0 {
-			j.journal.Wait(waitLogTimeout)
+			// sd_journal_wait can return immediately with a negative
+			// errno (e.g. -ENOSPC when inotify user instances are
+			// exhausted, so sd_journal_get_fd cannot create the
+			// inotify instance it needs). Fall back to a manual sleep
+			// so the loop honors waitLogTimeout instead of spinning
+			// at 100% CPU.
+			if r := j.journal.Wait(waitLogTimeout); r < 0 {
+				klog.Errorf("sd_journal_wait failed (%d), backing off for %v", r, waitLogTimeout)
+				time.Sleep(waitLogTimeout)
+			}
 			continue
 		}
 


pkg/systemlogmonitor/logwatchers/journald/log_watcher.go
```diff
@@ -114,7 +114,16 @@ func (j *journaldWatcher) watchLoop() {
...
 		if n == 0 {
-			j.journal.Wait(waitLogTimeout)
+			// sd_journal_wait can return immediately with a negative
+			// errno (e.g. -ENOSPC when inotify user instances are
+			// exhausted, so sd_journal_get_fd cannot create the
+			// inotify instance it needs). Fall back to a manual sleep
+			// so the loop honors waitLogTimeout instead of spinning
+			// at 100% CPU.
+			if r := j.journal.Wait(waitLogTimeout); r < 0 {
+				klog.Errorf("sd_journal_wait failed (%d), backing off for %v", r, waitLogTimeout)
```
DigitalVeer, line 124: It may be worth decoding the errno:

```go
if r := j.journal.Wait(waitLogTimeout); r < 0 {
    klog.Errorf("sd_journal_wait failed (%d: %s), backing off for %v", r, syscall.Errno(-r), waitLogTimeout)
    ...
}
```

if we use syscall.Errno(-r),  "no space left on device" is more descriptive than -28.

pkg/systemlogmonitor/logwatchers/journald/log_watcher.go
```diff
@@ -114,7 +114,16 @@ func (j *journaldWatcher) watchLoop() {
...
 		// If next reaches the end, wait for waitLogTimeout.
 		if n == 0 {
-			j.journal.Wait(waitLogTimeout)
+			// sd_journal_wait can return immediately with a negative
+			// errno (e.g. -ENOSPC when inotify user instances are
+			// exhausted, so sd_journal_get_fd cannot create the
+			// inotify instance it needs). Fall back to a manual sleep
+			// so the loop honors waitLogTimeout instead of spinning
+			// at 100% CPU.
+			if r := j.journal.Wait(waitLogTimeout); r < 0 {
```
DigitalVeer, line 123: this logs every ~5s for as long as the condition lasts (~720/hr). would it be better to log once when it fails and once when it recovers instead?

PR #1288, opened 2026-06-04
Honor max_output_length for custom plugin stdout capture

kv-cache:
**What this PR does / why we need it**:

Custom-plugin stdout is read into a hardcoded 4 KiB buffer
(`maxCustomPluginBufferBytes = 1024 * 4`) and only *then* truncated to the
configured `max_output_length`, so the effective limit is
`min(4096, max_output_length)`. Any plugin configured with
`max_output_length > 4096` is silently capped at 4096 bytes — the configured
value has no effect. For a plugin that emits JSON, the cut lands mid-token and
produces invalid output that a strict consumer discards entirely.

This drives the stdout read limit from `max_output_length`, bounded by a 1 MiB
safety ceiling to protect against a runaway plugin. stderr keeps a small fixed
cap since it is used only for diagnostic logging (`logPluginStderr`) and is never
included in the returned output. No behavior change for the default config
(`max_output_length = 80`).

**Which issue(s) this PR fixes**:

NONE

**Special notes for your reviewer**:

Adds `TestPluginRunCaptureRespectsMaxOutputLength` plus a fixture that emits
16384 bytes. With `max_output_length = 8192` the captured output must be exactly
8192 bytes — this fails on the current code (capped at 4096) and passes with the
fix. `go test -short -race ./pkg/custompluginmonitor/...` passes; gofmt/vet clean.

This raises the capture ceiling so a plugin's configured `max_output_length`
takes effect; it does not change that output beyond `max_output_length` is still
byte-truncated.

/kind bug

**Does this PR introduce a user-facing change?**:

```release-note
Custom plugin monitor now honors a `max_output_length` larger than 4096 bytes. Previously plugin stdout was silently capped at a hardcoded 4 KiB internal buffer regardless of the configured `max_output_length`.
```

**Note**:

Made with AI

### DIFF
--- a/pkg/custompluginmonitor/plugin/plugin.go
+++ b/pkg/custompluginmonitor/plugin/plugin.go
@@ -33,9 +33,18 @@ import (
 	"k8s.io/node-problem-detector/pkg/util/tomb"
 )
 
-// maxCustomPluginBufferBytes is the max bytes that a custom plugin is allowed to
-// send to stdout/stderr. Any bytes exceeding this value will be truncated.
-const maxCustomPluginBufferBytes = 1024 * 4
+// maxCustomPluginStdoutCeilingBytes is a hard safety ceiling on how many bytes
+// of stdout NPD reads from a custom plugin, bounding memory use from a runaway
+// plugin. The per-plugin max_output_length is what actually governs the message
+// size; this ceiling only caps the (misconfigured) case where max_output_length
+// itself exceeds it. It must be >= the largest supported max_output_length so a
+// plugin's configured limit is never silently truncated by the read buffer.
+const maxCustomPluginStdoutCeilingBytes = 1024 * 1024
+
+// maxCustomPluginStderrBytes caps stderr. run() uses stderr only for diagnostic
+// logging (logPluginStderr) and never includes it in the returned output, so it
+// does not need to honor max_output_length and keeps a small fixed cap.
+const maxCustomPluginStderrBytes = 1024 * 4
 
 type Plugin struct {
 	config     cpmtypes.CustomPluginConfig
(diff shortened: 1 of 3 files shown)


pkg/custompluginmonitor/plugin/plugin.go
```diff
@@ -210,14 +219,22 @@ func (p *Plugin) run(rule cpmtypes.CustomRule) (exitStatus cpmtypes.Status, outp
 		stderrErr error
 	)
 
+	// Capture enough stdout to honor the configured max_output_length, bounded
+	// by a hard safety ceiling. Previously this was a fixed 4 KiB buffer that
+	// silently truncated plugins configured with a larger max_output_length.
+	stdoutCapture := int64(*p.config.PluginGlobalConfig.MaxOutputLength)
```
DigitalVeer, line 225: This changes the default config behavior. [run()](https://github.com/kubernetes/node-problem-detector/blob/97d58d2cb54dc1143eeaf895302852b845e28484/pkg/custompluginmonitor/plugin/plugin.go#L245) trims whitespace after capture so the old code always had a 4 KiB window to find non-whitespace content in. Now that window is max_output_length itself which is 80 bytes by default.

I reproduced this is by having a plugin prints 80 spaces then `DiskHung: /dev/sda1 unresponsive` and exiting w/ a status code of 1. Right now NPD reports the message, but this PR reports `""`. The output is lost when the plugin fires.

Keeping the old 4 KiB as a floor fixes the >4 KiB truncation without changing anything else:

```go
stdoutCapture := int64(*p.config.PluginGlobalConfig.MaxOutputLength)
if stdoutCapture < 4096 {
	stdoutCapture = 4096
}
if stdoutCapture > maxCustomPluginStdoutCeilingBytes {
	stdoutCapture = maxCustomPluginStdoutCeilingBytes
}
```

A leading whitespace test case that tests this is worthwhile

pkg/custompluginmonitor/plugin/plugin.go
```diff
@@ -33,9 +33,18 @@ import (
...
-// maxCustomPluginBufferBytes is the max bytes that a custom plugin is allowed to
-// send to stdout/stderr. Any bytes exceeding this value will be truncated.
-const maxCustomPluginBufferBytes = 1024 * 4
+// maxCustomPluginStdoutCeilingBytes is a hard safety ceiling on how many bytes
+// of stdout NPD reads from a custom plugin, bounding memory use from a runaway
+// plugin. The per-plugin max_output_length is what actually governs the message
+// size; this ceiling only caps the (misconfigured) case where max_output_length
+// itself exceeds it. It must be >= the largest supported max_output_length so a
+// plugin's configured limit is never silently truncated by the read buffer.
+const maxCustomPluginStdoutCeilingBytes = 1024 * 1024
```
DigitalVeer, line 42: `max_output_length` > 1 MiB gets capped here with no log. Would it be better to check it once at config load and log or reject out-of-range values there?

1 mib may also be too big. The captured results still end up in [`resultChan`](https://github.com/kubernetes/node-problem-detector/blob/97d58d2cb54dc1143eeaf895302852b845e28484/pkg/custompluginmonitor/plugin/plugin.go#L53) which holds 1000 entries. That is 1000 * 1 MiB =~ about 1 GiB of RAM held by a node daemon. And the output goes verbatim into NodeCondition.Message and event messages.

DigitalVeer (commented):
Thanks for this fix! I notice two points worth addressing:

1. The read window now shrinks below the old 4 KiB for small max_output_length, so default-config behavior change
2. The 1 mib ceiling silently ignores larger configured values and may be too large

PR #1297, opened 2026-07-05
feat(exporters): replace opencensus with opentelemetry

### LINKED ISSUE #1008: Migrate OpenCensus to OpenTelemetry

hakman:
Based on @daveoy's #1141, adding remaining fixes.
Closes #1008
/cc @SergeyKanzhelev @wangzhen127


pkg/exporters/prometheusexporter/prometheus_exporter.go
```diff
@@ -21,35 +21,65 @@ import (
...
+// registry. Using a dedicated registry (instead of the global
+// prometheus.DefaultRegisterer) keeps the default Go runtime and process
+// collectors out of NPD's scrape output.
+func newExporterAndHandler() (*otelprometheus.Exporter, http.Handler, error) {
+	reg := prometheus.NewRegistry()
+
+	// Create Prometheus exporter with options to prevent automatic suffixing
+	promExporter, err := otelprometheus.New(
+		otelprometheus.WithRegisterer(reg),                                   // Use a dedicated registry
+		otelprometheus.WithTranslationStrategy(otlptranslator.NoTranslation), // Don't add suffixes or escape
```
DigitalVeer: I recorded the same metric and took a diff on the output between this PR & master.


```
# any plain scraper w/o escaping
curl -s localhost:20257/metrics | grep uptime
#   host_uptime 12345.6            (same on both branches)

# when using Prometheus 3.x 
curl -s -H 'Accept: text/plain; version=1.0.0; escaping=allow-utf-8' \
     localhost:20257/metrics | grep uptime
#   PR head:  {"host/uptime"} 12345.6     <- renamed
#   master:   host_uptime 12345.6          <- unchanged
```
---

We should account for users scraping with Prometheus 3.x.

Given that Prometheus 3.x supports [utf-8](https://prometheus.io/docs/guides/utf8/) and has `escaping=allow-utf-8` by [default](https://prometheus.io/docs/guides/utf8/#:~:text=At%20scrape%20time%2C%20the%20scraping%20system%20must%20pass%20escaping%3Dallow%2Dutf%2D8%20in%20the%20Accept%20header%20in%20order%20to%20be%20served%20UTF%2D8%20names.%20If%20scrape%20target%20does%20not%20see%20this%20header%2C%20it%20will%20automatically%20convert%20UTF%2D8%20names%20to%20legacy%2Dcompatible%20using%20underscore%20replacement.), we should make sure the metric sent out is the same before and after this change.

pkg/util/otel/common.go
```diff
@@ -0,0 +1,108 @@
...
+		// Add all registered readers
+		for _, reader := range readers {
+			opts = append(opts, sdkmetric.WithReader(reader))
+		}
+
+		// Create meter provider with all readers and resource
+		globalMeterProvider = sdkmetric.NewMeterProvider(opts...)
+
+		// Set as global meter provider
+		otel.SetMeterProvider(globalMeterProvider)
```
DigitalVeer, line 93: We should pass [`option.WithTelemetryDisabled()`](https://github.com/kubernetes/node-problem-detector/blob/76310355bd741592d7deb7c2fedc579f2157dea2/vendor/google.golang.org/api/option/option.go#L390) in [monitoringClientOptions](https://github.com/kubernetes/node-problem-detector/blob/76310355bd741592d7deb7c2fedc579f2157dea2/pkg/exporters/stackdriver/stackdriver_exporter.go#L174).

  By default, `google.golang.org/api` [adds an otelgrpc handler](https://github.com/kubernetes/node-problem-detector/blob/76310355bd741592d7deb7c2fedc579f2157dea2/vendor/google.golang.org/api/transport/grpc/dial.go#L384) to each gRPC client. Thus the exporter makes `rpc.client.*` metrics for its own calls. These metric names are not in [NPDMetricToSDMetric](https://github.com/kubernetes/node-problem-detector/blob/76310355bd741592d7deb7c2fedc579f2157dea2/pkg/exporters/stackdriver/stackdriver_exporter.go#L59).

Cloud Monitoring rejects unknown metrics on export cycles and each failed export makes more rpc.client.* metrics so this can cause a failure loop.

DigitalVeer (commented):
Thanks for this! I work with NPD at GKE and I hope I can add my two cents here.

pkg/exporters/prometheusexporter/prometheus_exporter.go
hakman: Good catch. Fixed.

pkg/util/otel/common.go
hakman: Sounds good, done.

cmd/nodeproblemdetector/node_problem_detector.go
```diff
@@ -59,9 +55,19 @@ func npdMain(ctx context.Context, npdo *options.NodeProblemDetectorOptions) erro
 		defaultExporters = append(defaultExporters, pe)
 		klog.Info("Prometheus exporter started.")
 	}
-
 	plugableExporters := exporters.NewExporters()
 
+	// Initialize OpenTelemetry meter provider with all registered readers
+	// This must be called after all exporters have been created and registered their readers
+	otelutil.InitializeMeterProvider()
```
wangzhen127: Add a defer function to call Shutdown()?

pkg/util/otel/common.go
```diff
@@ -0,0 +1,108 @@
...
+	meterProviderOnce   sync.Once
+	readers             []sdkmetric.Reader
+	readersMutex        sync.Mutex
+	extraResourceAttrs  []attribute.KeyValue
+)
+
+// AddMetricReader adds a metric reader to the global meter provider configuration
+// This should be called before InitializeMeterProvider()
+// Accepts both pull readers (like Prometheus) and push readers (like Stackdriver)
+func AddMetricReader(reader sdkmetric.Reader) {
```
wangzhen127: Does `InitializeMeterProvider()` must run before this function, so that it can pick up all the readers? If so, maybe consider error out if `globalMeterProvider` is non-nil in this function to prevent silent misconfiguration during future refactors or plugins

pkg/util/otel/common.go
```diff
@@ -0,0 +1,108 @@
...
+	readersMutex.Lock()
+	defer readersMutex.Unlock()
+	readers = append(readers, reader)
+}
+
+// AddResourceAttributes registers additional resource attributes that are merged
+// into the global OpenTelemetry resource by InitializeMeterProvider().
+// This should be called before InitializeMeterProvider().
+func AddResourceAttributes(attrs ...attribute.KeyValue) {
```
wangzhen127: Ditto. Consider checking `globalMeterProvider` being non-nil and error out.

wangzhen127 (commented):
Mostly lgtm. Just a few nits.

wangzhen127:
Can you also rebase to head and resolve conflicts?

cmd/nodeproblemdetector/node_problem_detector.go
hakman: done

pkg/util/otel/common.go
hakman: done

pkg/util/otel/common.go
hakman: done

hakman:
> Mostly lgtm. Just a few nits.

All done and rebased, thanks for taking another pass.

wangzhen127:
/lgtm
/hold

Thanks a lot for the work! Feel free to unhold to merge.

hakman:
Thanks, @wangzhen127!

I think this has been open long enough for anyone interested in the topic to provide feedback. Let’s merge it and continue improving it from there.

/unhold

daveoy:
🔥

PR #1303, opened 2026-07-08
fix: keep bundled custom plugin scripts outside /config

immanuwell:
The README says to overwrite `/config` with a ConfigMap. That hides `./config/plugin/*`, so the shipped custom plugin configs become a small footgun and NPD falls over on startup.

Fix
Copy the bundled scripts to `/home/kubernetes/bin/plugin` and point the shipped custom plugin configs there.
Also add coverage for those paths, and keep the filelog watcher test from blowing up when host `inotify` is tapped out.

Repro
1. Build the image.
2. Mount only `custom-plugin-monitor.json` at `/config`.
3. Run `/node-problem-detector --enable-k8s-exporter=false --config.custom-plugin-monitor=/config/custom-plugin-monitor.json`.
4. Before this patch it fails with `rule path "./config/plugin/check_ntp.sh" does not exist`.

Test
`go test ./...`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,6 +31,7 @@ COPY --from=builder /src/bin/node-problem-detector /node-problem-detector
 
 ARG LOGCOUNTER
 COPY --from=builder /src/bin/health-checker /src/${LOGCOUNTER} /home/kubernetes/bin/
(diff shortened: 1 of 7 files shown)


immanuwell:
@dchen1107 @klueska please take a look

Issue #1176, opened 2025-11-28
scaling down will send unnecessary alerts


BraceCY:
The npd component checks the health status of kube-proxy via the /healthz interface
https://github.com/kubernetes/node-problem-detector/blob/2baac6c06751cc6dccd8b0ad9b6d08b6780f1a13/pkg/healthchecker/types/types.go#L83

When cluster-autoscaler performs elastic scaling down, it will add the ToBeDeletedByClusterAutoscaler taint to nodes. When a node has the ToBeDeletedByClusterAutoscaler taint, the /healthz interface of kube-proxy will consider itself to be in an unhealthy state.
https://github.com/kubernetes/kubernetes/blob/d4672d9a9db5687c4561048c4840ef1f8ccb3c75/pkg/proxy/healthcheck/proxy_health.go#L188

The elastic scaling down of cluster-autoscaler is standard behavior and should not trigger any alarms, as it can cause confusion.

hakman:
@BraceCY thanks for reporting this.
Why do you think it's a NPD issue and not a kube-proxy issue?
CC @wangzhen127 @SergeyKanzhelev

BraceCY:
@hakman thank you for your reply!

The Kubernetes SIG-Network has discussed the behavior of kube-proxy at https://github.com/kubernetes/enhancements/issues/3836.
I believe that KEP-3836 explains the scenario.

**Goals:**
_Offer a better capability for connection draining when terminating nodes, especially for load balancers that support this._

Kube-proxy provides two interfaces: /healthz and /livez. The /healthz interface checks node eligibility, excluding nodes that are being deleted or tainted with "ToBeDeletedTaint", among other conditions. The /livez interface only checks the status of kube-proxy.

I think it is reasonable for nodes to be in a state of being deleted or tainted with "ToBeDeletedTaint" as part of normal operations. Therefore, it would be more appropriate for NPD to send alarms based on the /livez interface.

CC @wangzhen127 @SergeyKanzhelev

dpacgdm:
Opened a fix in https://github.com/kubernetes/node-problem-detector/pull/1306 — switch the kube-proxy probe from `/healthz` to `/livez` so Cluster Autoscaler scale-down no longer triggers false unhealthy alerts.

PR #1306, opened 2026-07-09
healthchecker: use kube-proxy /livez to avoid scale-down false alerts

dpacgdm:
## Summary
Cluster Autoscaler marks nodes with `ToBeDeletedByClusterAutoscaler` during scale-down. kube-proxy `/healthz` intentionally fails in that case for load-balancer connection draining ([KEP-3836](https://github.com/kubernetes/enhancements/issues/3836)), which caused NPD to report kube-proxy as unhealthy and fire unnecessary alerts.

Switch the kube-proxy health check endpoint from `/healthz` to `/livez`. `/livez` still checks process and dataplane synchronization health; it only excludes the node-eligibility check used for scale-down draining.

## Related issue
https://github.com/kubernetes/node-problem-detector/issues/1176

## Test plan
- [x] `go test ./pkg/healthchecker/... ./cmd/healthchecker/...`
- [x] Updated endpoint unit tests to expect `/livez` for kube-proxy

### DIFF
--- a/pkg/healthchecker/types/types.go
+++ b/pkg/healthchecker/types/types.go
@@ -82,7 +82,8 @@ func setKubeEndpoints() {
 	kubeletHealthCheckEndpoint = fmt.Sprintf("http://%s/healthz", net.JoinHostPort(hostAddress, kubeletPort))
 	// Use /livez for kube-proxy so Cluster Autoscaler scale-down (ToBeDeletedByClusterAutoscaler
 	// taint) does not make NPD report kube-proxy as unhealthy. /healthz intentionally fails in
-	// that case for load-balancer connection draining (KEP-3836); /livez only reflects process health.
+	// that case for load-balancer connection draining (KEP-3836). /livez still checks process and
+	// dataplane synchronization health; it only excludes the node-eligibility check.
 	kubeProxyHealthCheckEndpoint = fmt.Sprintf("http://%s/livez", net.JoinHostPort(hostAddress, kubeProxyPort))
 }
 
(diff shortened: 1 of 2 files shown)


dpacgdm:
Friendly ping: this PR is still waiting on `/ok-to-test` from a kubernetes org member so CI can run. Happy to address any feedback — thanks for taking a look when you have a chance.

pkg/healthchecker/types/types.go
```diff
@@ -80,7 +80,10 @@ func setKubeEndpoints() {
 	}
 
 	kubeletHealthCheckEndpoint = fmt.Sprintf("http://%s/healthz", net.JoinHostPort(hostAddress, kubeletPort))
-	kubeProxyHealthCheckEndpoint = fmt.Sprintf("http://%s/healthz", net.JoinHostPort(hostAddress, kubeProxyPort))
+	// Use /livez for kube-proxy so Cluster Autoscaler scale-down (ToBeDeletedByClusterAutoscaler
+	// taint) does not make NPD report kube-proxy as unhealthy. /healthz intentionally fails in
+	// that case for load-balancer connection draining (KEP-3836); /livez only reflects process health.
```
hakman:
```suggestion
	// that case for load-balancer connection draining (KEP-3836); /livez reports kube-proxy
	// dataplane health without considering node eligibility.
```

hakman (changes requested):
Looks good overall. Small comment nit: `/livez` still checks dataplane synchronization health, it only excludes node eligibility. Could we adjust the wording accordingly?

dpacgdm:
Thanks — will fix the wording. You're right: `/livez` still covers dataplane synchronization health; it only drops the node-eligibility / scale-down draining check that `/healthz` includes. Updating the comment (and PR description) accordingly.

hakman:
Thanks @dpacgdm, this will be a nice improvement.
/lgtm
/approve

PR #1307, opened 2026-07-09
config: include --cri-timeout in containerd health-checker defaults

dpacgdm:
## Summary
`--cri-timeout` is already supported by the health-checker binary and is correctly passed through custom plugin rule `args`. The default containerd health-checker configs omitted the flag, so `crictl` fell back to the 2s built-in default.

This change:
- Adds `--cri-timeout=10s` to the Linux and Windows containerd sample configs
- Adds a unit test that `--cri-timeout` parses correctly

## Related issue
Fixes https://github.com/kubernetes/node-problem-detector/issues/1248

Note: when the flag is present in rule `args`, NPD already propagates it (`util.Exec(ctx, rule.Path, rule.Args...)`). The reporter's `--timeout=2s` observation matches the default when the flag is absent from the config they deployed.

## Test plan
- [x] `go test ./cmd/healthchecker/options -count=1`

### DIFF
--- a/config/health-checker-containerd.json
+++ b/config/health-checker-containerd.json
@@ -25,6 +25,7 @@
         "--component=cri",
         "--enable-repair=true",
         "--cooldown-time=2m",
+        "--cri-timeout=10s",
         "--health-check-timeout=60s"
       ],
       "timeout": "3m"
--- a/config/windows-health-checker-containerd.json
+++ b/config/windows-health-checker-containerd.json
@@ -25,10 +25,10 @@
           "--component=cri",
           "--enable-repair=true",
           "--cooldown-time=2m",
+          "--cri-timeout=10s",
           "--health-check-timeout=60s"
         ],
         "timeout": "3m"
       }
     ]
   }
-  


config/health-checker-containerd.json
```diff
@@ -1,4 +1,4 @@
-{
+﻿{
```
hakman: The first line of `config/health-checker-containerd.json` now starts with `\xEF\xBB\xBF`, which Go's `encoding/json` rejects, so NPD fails to load the config with `invalid character 'ï'`. It likely came from your editor saving as "UTF-8 with BOM". You can strip it with:
   ```sh
   perl -i -pe 's/^\x{FEFF}//' config/health-checker-containerd.json
   ```

cmd/healthchecker/options/options_test.go
```diff
@@ -74,3 +76,20 @@ func TestIsValid(t *testing.T) {
 		})
 	}
 }
+
+func TestCriTimeoutFlagParsing(t *testing.T) {
```
hakman: Please drop changes to this file, it does not add much value.

hakman (changes requested):
Thanks for the PR @dpacgdm! The change itself is the right call, and scoping `--cri-timeout` to just the containerd configs is correct.
I added a few small comments, please update the PR and squash the commits.

dpacgdm:
@hakman Thanks for the review — glad the containerd-only scope looks right.

I'll push an update that:
1. Strips the UTF-8 BOM from `config/health-checker-containerd.json`
2. Reverts the `options_test.go` changes
3. Squashes to a single commit

dpacgdm:
Update pushed: BOM stripped, `options_test.go` reverted, single commit. PTAL when you have a chance.

hakman (approved):

PR #1318, opened 2026-07-13
Fix invalid config JSON and enforce canonical config format

hakman:
`config/disk-log-message-filelog.json` contained trailing commas and no trailing newline, so it could not be parsed by `encoding/json` and NPD would fail to start when configured with it.

This PR:

- Fixes the invalid JSON.
- Adds `TestShippedConfigs`, which verifies that every shipped JSON config under `config/` — top level and one directory down (e.g. `config/exporter`, `config/guestosconfig`) — parses (catching trailing commas, UTF-8 BOMs and other syntax errors) and is in the canonical format (2-space indent, trailing newline), showing a diff on failure.
- Adds `make fmt-configs` to rewrite nonconforming files in place (`UPDATE_EXPECTED=y`).
- Normalizes the existing config files to the canonical format (whitespace-only changes).

### DIFF
--- a/pkg/systemlogmonitor/config_json_test.go
+++ b/pkg/systemlogmonitor/config_json_test.go
@@ -1,0 +1,90 @@
+/*
+Copyright 2026 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package systemlogmonitor
+
+import (
+	"bytes"
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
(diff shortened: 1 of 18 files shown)


pkg/systemlogmonitor/config_json_test.go
```diff
@@ -0,0 +1,83 @@
...
+	return buf.Bytes(), nil
+}
+
+// TestShippedConfigs verifies that every JSON file shipped in the config/ directory parses with
+// encoding/json, the same way the monitors load them at startup, and that it is written in the
+// canonical format. The parse check catches trailing commas, UTF-8 BOMs and other syntax errors
+// that would make NPD fail to start with the config. The format check keeps diffs in PRs minimal.
+// Run `make fmt-configs` to rewrite nonconforming files in the canonical format.
+func TestShippedConfigs(t *testing.T) {
+	configs, err := filepath.Glob(filepath.Join("..", "..", "config", "*.json"))
```
DigitalVeer: This doesn't find [`config/exporter/stackdriver-exporter.json`](https://github.com/kubernetes/node-problem-detector/blob/c96c10360201bfbc23a99b35d4f4dfb1c083bfde/config/exporter/stackdriver-exporter.json) and [`config/guestosconfig/known-modules.json`](https://github.com/kubernetes/node-problem-detector/blob/c96c10360201bfbc23a99b35d4f4dfb1c083bfde/config/guestosconfig/known-modules.json). The stackdriver exporter config can crash NPD on startup if the config is not correctly formatted, so it's worth catching here. We may need a second glob for `config/*/*.json`.

hakman: Thanks, good catch.

PR #1320, opened 2026-07-20
Add per-rule custom plugin intervals

DigitalVeer:
## Summary

Custom plugin rules can set an optional invoke interval and rules without this field will use the global invoke interval.
The scheduler groups rules with equal parsed durations.
Each group runs one synchronized batch for each tick.

## Compatibility

The default schedule keeps one shared batch for all rules.
The scheduler creates all tickers before the combined boot batch.
Downstream result handling does not change.

## Tests

This PR adds fake-clock tests for the scheduler.

The lifecycle bot closed prior attempts #357 and #358.

```release-note
Custom plugin rules can set their own invoke intervals.
```


hakman (approved):
Thanks @DigitalVeer!

PR #1322, opened 2026-07-20
Cache compiled regular expressions in the log buffer

DigitalVeer:
## Summary

The log buffer keeps each compiled regular expression in a cache.
The cache size is not more than the number of patterns given to `Match`.
This PR removes the TODO in `log_buffer.go`.

## Benchmark

One `Match` call with a default kernel monitor pattern:

| | Time per call | Memory per call | Allocations per call |
| --- | --- | --- | --- |
| Before | 2590 ns | 9265 B | 49 |
| After | 211 ns | 776 B | 1 |

```release-note
NONE
```

### DIFF
--- a/pkg/systemlogmonitor/log_buffer.go
+++ b/pkg/systemlogmonitor/log_buffer.go
@@ -39,6 +39,8 @@ type logBuffer struct {
 	msg     []string
 	max     int
 	current int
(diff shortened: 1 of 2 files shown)


hakman (approved):
Thanks @DigitalVeer, LGTM!

PR #1323, opened 2026-07-22
Precompile log buffer regular expressions

hakman:
Compile fixed patterns during monitor and log counter initialization, pass compiled expressions to the log buffer, and reject invalid log counter patterns before watching logs.

/cc @wangzhen127 @ameukam

### DIFF
--- a/pkg/logcounter/log_counter_test.go
+++ b/pkg/logcounter/log_counter_test.go
@@ -20,25 +20,65 @@ limitations under the License.
 package logcounter
 
 import (
+	"strings"
 	"testing"
 	"time"
 
 	testclock "k8s.io/utils/clock/testing"
 
+	"k8s.io/node-problem-detector/cmd/logcounter/options"
 	"k8s.io/node-problem-detector/pkg/logcounter/types"
 	"k8s.io/node-problem-detector/pkg/systemlogmonitor"
 	systemtypes "k8s.io/node-problem-detector/pkg/systemlogmonitor/types"
 )
 
-func NewTestLogCounter(pattern string, startTime time.Time) (types.LogCounter, *testclock.FakeClock, chan *systemtypes.Log) {
+func newTestLogCounter(t *testing.T, pattern string, startTime time.Time) (types.LogCounter, *testclock.FakeClock, chan *systemtypes.Log) {
+	t.Helper()
+	compiledPattern, err := systemlogmonitor.CompilePattern(pattern)
+	if err != nil {
+		t.Fatalf("failed to compile pattern %q: %v", pattern, err)
+	}
 	logCh := make(chan *systemtypes.Log)
 	clock := testclock.NewFakeClock(startTime)
 	return &logCounter{
 		logCh:   logCh,
 		buffer:  systemlogmonitor.NewLogBuffer(bufferSize),
-		pattern: pattern,
+		pattern: compiledPattern,
 		clock:   clock,
 	}, clock, logCh
+}
+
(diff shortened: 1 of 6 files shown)


pkg/systemlogmonitor/log_monitor.go
```diff
@@ -73,7 +75,7 @@ func NewLogMonitorOrDie(configPath string) types.Monitor {
 	}
 	// Apply default configurations
 	(&l.config).ApplyDefaultConfiguration()
-	err = l.config.ValidateRules()
```
DigitalVeer, line 76: After this change, no code calls `ValidateRules` anymore. Let's delete it or add a reason to keep it.

pkg/systemlogmonitor/log_buffer.go
```diff
@@ -49,26 +47,25 @@ type logBuffer struct {
...
+		buffer: make([]*types.Log, maxLines),
+		msg:    make([]string, maxLines),
+		max:    maxLines,
 	}
 }
 
+// CompilePattern compiles a log buffer pattern that must match to the end of
+// the buffered logs.
+func CompilePattern(expr string) (*regexp.Regexp, error) {
+	return regexp.Compile(expr + `\z`)
```
DigitalVeer, line 62: We should compile the expression first. I tried the regex `"foo\"` on this branch and master. This fails on master but succeeds here.

`foo\` (invalid) became `foo\\z` (literal: `foo\z`) which is valid but incorrect. 

---
```go
func CompilePattern(expr string) (*regexp.Regexp, error) {
	if _, err := regexp.Compile(expr); err != nil {
		return nil, err
	}
	return regexp.Compile(expr + `\z`)
}
```

pkg/logcounter/log_counter_test.go
```diff
@@ -25,22 +25,55 @@ import (
...
+			pattern: "[",
+		},
+		{
+			name:          "revert pattern",
+			revertPattern: "[",
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			_, err := NewJournaldLogCounter(&options.LogCounterOptions{
+				Pattern:       tc.pattern,
```
DigitalVeer, line 71: This test also passes on master with the error `empty source is not allowed`. There is no [JournaldSource](https://github.com/kubernetes/node-problem-detector/blob/eeb802f03739c37b7b31cba02c97b96f60e70e88/cmd/logcounter/options/options.go#L32) in the test, so the construction fails before we even run the pattern check. Funnily enough, since we get an error message (even though it's the wrong one), the test is passing.

We should assert the actual error message.

DigitalVeer (commented):
nice! glad we're moving pattern compilation to startup now :) - a few catches

pkg/logcounter/log_counter_test.go
hakman: Updated the table test to assert the specific `invalid pattern` and `invalid revert pattern` errors, so unrelated watcher errors no longer satisfy it.

pkg/systemlogmonitor/log_monitor.go
hakman: Removed the unused `ValidateRules` method; monitor construction now calls `compileRules` directly.

pkg/systemlogmonitor/log_buffer.go
hakman: Updated `CompilePattern` to validate the original expression before appending `\z`, and added a regression test for a trailing backslash.

PR #1325, opened 2026-07-23
Set the Node UID on recorded events

### LINKED ISSUE #1324: Node events are missing from 'kubectl describe node' because involvedObject.uid is not set
### What happened?
NPD successfully creates k8s events for node problems, but NPD-generated events are not shown in the `Events` section of:
```bash
kubectl describe node <node-name>
```

The Events can still be found using `kubectl get events`, but their `involvedObject` does not contain the actual Node UID.

As a result, clients that correlate events with k8s objects by UID cannot associate these Events with the current Node.

### Environment
- NPD version: `v1.35.2`
- Kubernetes version: `v1.35.4`

### Root cause
The relevant code is [here](https://github.com/kubernetes/node-problem-detector/blob/eeb802f03739c37b7b31cba02c97b96f60e70e88/pkg/exporters/k8sexporter/problemclient/problem_client.go#L156-L164). The Node `ObjectReference` is constructed using only the Node name, so the actual Node UID is not populated.

Previously, the UID was removed by [PR #1163: Remove the invalid node.UID](https://github.com/kubernetes/node-problem-detector/pull/1163), because a Node name is not its UID. But the invalid value was not replaced with the actual UID from `node.metadata.uid`. As a result, the Event reference now has no UID at all.

Any feedback or suggestions would be greatly appreciated.

/kind bug

DigitalVeer:
Recorded events refer to the node with an `ObjectReference` that has no UID. The client already reads the Node object at startup to check that kube-apiserver is ready, and this PR stores the UID from that read. The node status patch returns the Node object, so the client stores that UID too. This also removes a pending TODO in `problem_client.go`.

Fixes #1324

### DIFF
--- a/pkg/exporters/k8sexporter/k8s_exporter.go
+++ b/pkg/exporters/k8sexporter/k8s_exporter.go
@@ -123,7 +123,9 @@ func (ke *k8sExporter) startHTTPReporting(npdo *options.NodeProblemDetectorOptio
 func waitForAPIServerReadyWithTimeout(ctx context.Context, c problemclient.Client, npdo *options.NodeProblemDetectorOptions) error {
 	return wait.PollUntilContextTimeout(ctx, npdo.APIServerWaitInterval, npdo.APIServerWaitTimeout, true, func(ctx context.Context) (done bool, err error) {
 		// If NPD can get the node object from kube-apiserver, the server is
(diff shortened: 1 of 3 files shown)


hakman:
Thanks for fixing this @DigitalVeer
/lgtm
/approve

PR #1329, opened 2026-07-27
Cache the joined log buffer and skip it for single-line rules

DigitalVeer:
This PR makes two changes on top of #1323:

1. The buffer caches the joined string. `Push` clears the cache.
2. `CompilePattern` marks the rules that can match only one line. `Match` tests these rules against the last line only.

This will help NPD act more reliably during log floods.

## Benchmark

`BenchmarkPushAndMatchAll` (linux/arm64, Go 1.26):

| | Time per line | Memory per line | Allocations per line |
| --- | --- | --- | --- |
| Before, buffer size 10 | 18484 ns | 9221 B | 26 |
| After, buffer size 10 | 1659 ns | 0 B | 0 |
| Before, buffer size 100 | 174067 ns | 99118 B | 28 |
| After, buffer size 100 | 1687 ns | 0 B | 0 |

## API change

- `CompilePattern` returns `*Pattern` instead of `*regexp.Regexp`.
- This removes `String()` from the `LogBuffer` interface as it is not used anymore.

### DIFF
--- a/pkg/logcounter/log_counter.go
+++ b/pkg/logcounter/log_counter.go
@@ -21,7 +21,6 @@ package logcounter
 
 import (
 	"fmt"
-	"regexp"
 	"time"
 
 	"k8s.io/utils/clock"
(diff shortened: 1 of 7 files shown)


hakman:
Thanks @DigitalVeer!
/lgtm
/approve

Issue #1328, opened 2026-07-27
NPD needs significant unit test coverage


DigitalVeer:
We should improve the overall test coverage present in NPD.

---

Currently, NPD's unit test coverage is 46.0% of 2299 statements. There are important packages, such as the custom plugin monitor, that have close to no test coverage at all.

Recent PRs such as #1322 and #1323 showed real headroom to make NPD faster and memory effecient, and more optimization work like that is worth pursuing. NPD also stands to benefit from agent-assisted development which can help us unlock more performance and reliability improvements over time. This type of work is only safe on top of reliable test infrastructure.

The table below are packages worth targeting in a first round of unit test improvement due to their importance in NPD workloads:

| Package | Coverage | Planned work |
| --- | --- | --- |
| `pkg/exporters/k8sexporter` | 0.0% | export test with the fake problem client |
| `pkg/exporters/prometheusexporter` | 0.0% | export test |
| `pkg/custompluginmonitor` | 0.8% | table test for `generateStatus` |
| `pkg/healthchecker` | 20.6% | tests for `getRepairFunc`, `getUptimeFunc`, `checkForPattern` |
| `pkg/systemlogmonitor/logwatchers/journald` | 23.2% | tests for the watch loop |
| `pkg/systemstatsmonitor` | 45.2% | fixture test for the OS feature collector |

Every other measured package is in the collapsed table. A lot of the remaining 0% rows are thin  wrappers, build-time metadata, or type declarations where a unit test would not help us much immediately.

<details>
<summary>Full package test coverage</summary>

| Package | Coverage |
| --- | --- |
| `pkg/exporters/stackdriver/gce` | 0.0% |
| `pkg/systemlogmonitor/logwatchers` | 0.0% |
| `pkg/systemlogmonitor/logwatchers/testing` | 0.0% |
| `pkg/version` | 0.0% |
| `cmd/healthchecker` | 0.0% |
| `cmd/logcounter` | 0.0% |
| `cmd/logcounter/options` | 0.0% |
| `pkg/exporters/stackdriver` | 3.3% |
| `pkg/exporters/k8sexporter/problemclient` | 5.4% |
| `pkg/problemdetector` | 23.3% |
| `cmd/healthchecker/options` | 26.5% |
| `cmd/nodeproblemdetector` | 36.4% |
| `pkg/util/metrics` | 44.2% |
| `pkg/problemdaemon` | 45.0% |
| `pkg/systemlogmonitor` | 52.2% |
[...]

PR #1330, opened 2026-07-29
Add unit tests for the custom plugin monitor

DigitalVeer:
This PR bumps the unit test coverage for custom plugin monitor from 0.8% to 95.3%

---

This PR adds tests for `generateStatus`, the monitor lifecycle and the metrics path.

This is the first item of #1328.

| Package | Before | After |
| --- | --- | --- |
| `pkg/custompluginmonitor` | 0.8% | 95.3% |

## What the tests check

- `TestGenerateStatusForConditions` replays plugin results
against one monitor and checks the condition and the event after each step for five scenarios.

- `TestGenerateStatusForTemporaryProblem` checks that a temporary rule creates an event and
moves no condition.

- `TestGenerateStatusMetrics` checks the problem counter and the problem gauge through the
existing metrics stub. The counter increments for new problems only.

- `TestMonitorLoopReportsPluginResults` and `TestMonitorLoopExitsWhenResultChannelCloses`
check the goroutine lifecycle. `Stop` must return. A closed result channel must end the
loop.

- `TestNewCustomPluginMonitorOrDie` writes a config file and checks that the constructor
applies the defaults.

The lifecycle tests use the plugin scripts in `pkg/custompluginmonitor/plugin/test-data`.

## Validation

| Check | Result |
| --- | --- |
| `go test -cover ./pkg/custompluginmonitor/` | 95.3% |
| `go test -timeout=1m -race -short -count=5` | pass, 3.6s |
| `golangci-lint run --config .golangci.yml` | 0 issues |

### DIFF
--- a/pkg/custompluginmonitor/custom_plugin_monitor_test.go
+++ b/pkg/custompluginmonitor/custom_plugin_monitor_test.go
@@ -17,15 +17,680 @@ limitations under the License.
 package custompluginmonitor
 
 import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"runtime"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
+	"k8s.io/node-problem-detector/pkg/custompluginmonitor/plugin"
+	cpmtypes "k8s.io/node-problem-detector/pkg/custompluginmonitor/types"
 	"k8s.io/node-problem-detector/pkg/problemdaemon"
+	"k8s.io/node-problem-detector/pkg/problemmetrics"
+	"k8s.io/node-problem-detector/pkg/types"
+	"k8s.io/node-problem-detector/pkg/util"
+	"k8s.io/node-problem-detector/pkg/util/metrics"
+	"k8s.io/node-problem-detector/pkg/util/tomb"
 )
+
+const (
+	testSource         = "cpm-test"
+	testConfigPath     = "test-config-path"
+	testCondition      = "TestCondition"
+	testConditionOK    = "TestConditionOK"
+	testConditionOKMsg = "test condition is ok"
+	otherCondition     = "OtherCondition"
+	otherConditionOK   = "OtherConditionOK"
+	otherOKMsg         = "other condition is ok"
+	testProblemReason  = "TestConditionProblem"
+	testTempReason     = "TestTempProblem"
+	testWait           = 3 * time.Second
+	settleWait         = 200 * time.Millisecond
+)
+
+// testEpoch is a fixed past instant stamped into Transition before a test runs.
+var testEpoch = time.Date(2020, 1, 1, 0, 0, 0, 0, time.UTC)
 
 func TestRegistration(t *testing.T) {
 	assert.NotPanics(t,
 		func() { problemdaemon.GetProblemDaemonHandlerOrDie("custom-plugin-monitor") },
 		"Custom plugin monitor failed to register itself as a problem daemon.")
+}
+
+// testOptions holds fields the tests and structs below may need.
+type testOptions struct {
+	defaultConditions      []types.Condition
+	rules                  []*cpmtypes.CustomRule
+	messageChange          bool
+	enableMetricsReporting bool
+	skipInitialStatus      bool
+}
+
+// defaultTestConditions returns two conditions. The condition under test is second.
+func defaultTestConditions() []types.Condition {
+	return []types.Condition{
+		{Type: otherCondition, Reason: otherConditionOK, Message: otherOKMsg},
+		{Type: testCondition, Reason: testConditionOK, Message: testConditionOKMsg},
+	}
+}
+
+// newTestConfig builds a config with the defaults applied.
+func newTestConfig(t *testing.T, o testOptions) cpmtypes.CustomPluginConfig {
+	t.Helper()
+	cfg := cpmtypes.CustomPluginConfig{
+		Plugin:            "custom",
+		Source:            testSource,
+		DefaultConditions: o.defaultConditions,
+		Rules:             o.rules,
+	}
+	require.NoError(t, (&cfg).ApplyConfiguration())
+	messageChange := o.messageChange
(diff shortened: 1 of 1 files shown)


pkg/custompluginmonitor/custom_plugin_monitor_test.go
```diff
@@ -17,15 +17,680 @@ limitations under the License.
...
+				assert.Equal(t, otherConditionOK, other.Reason, "step %d", i)
+				assert.Equal(t, otherOKMsg, other.Message, "step %d", i)
+				assert.True(t, other.Transition.Equal(testEpoch), "step %d", i)
+
+				if !s.wantEvent {
+					assert.Empty(t, got.Events, "step %d", i)
+					continue
+				}
+				require.Len(t, got.Events, 1, "step %d", i)
+				event := got.Events[0]
```
hakman, line 376: The expected event is built with the same helper production uses, so this doesn't independently check severity. If that helper always returned `Warn`, resolution events would become warnings and the whole suite would still pass.

```suggestion
				event := got.Events[0]
				// Assert severity literally: `want` below is built with the same helper
				// generateStatus uses, so it cannot catch a bad severity mapping.
				wantSeverity := types.Info
				if s.wantStatus == types.True {
					wantSeverity = types.Warn
				}
				assert.Equal(t, wantSeverity, event.Severity, "step %d", i)
```

pkg/custompluginmonitor/custom_plugin_monitor_test.go
```diff
@@ -17,15 +17,680 @@ limitations under the License.
...
+	require.NoError(t, err)
+
+	status := receiveStatus(t, statusChan)
+	require.Len(t, status.Conditions, 2)
+	assert.Equal(t, types.True, status.Conditions[1].Status)
+	assert.Equal(t, testProblemReason, status.Conditions[1].Reason)
+	require.Len(t, status.Events, 1)
+	assert.Equal(t, types.Warn, status.Events[0].Severity)
+
+	requireStop(t, c)
```
hakman, line 583: `requireStop` shows that `Stop()` returned, but not that the plugin it owns actually stopped. Delete `c.plugin.Stop()` at `custom_plugin_monitor.go:148` and this package still passes under `-race`, with the plugin's ticker and rule goroutines left running.

```suggestion
	requireStop(t, c)

	// Stop must also shut down the plugin it owns.
	select {
	case _, open := <-c.plugin.GetResultChan():
		assert.False(t, open, "Stop did not close the plugin result channel")
	case <-time.After(testWait):
		t.Fatal("Stop did not close the plugin result channel")
	}
```

pkg/custompluginmonitor/custom_plugin_monitor_test.go
```diff
@@ -17,15 +17,680 @@ limitations under the License.
...
+}
+
+func TestMonitorLoopExitsWhenResultChannelCloses(t *testing.T) {
+	c := newTestMonitor(t, testOptions{defaultConditions: defaultTestConditions()})
+	// monitorLoop must initialize the conditions itself.
+	c.conditions = nil
+	statusChan, err := c.Start()
+	require.NoError(t, err)
+
+	status := receiveStatus(t, statusChan)
```
hakman: This doesn't observe whether `monitorLoop` returns. With no rules, no result-derived status can follow the initial one, so `requireNoStatus` at the end passes either way — change the closed-channel `return` at `custom_plugin_monitor.go:141` to `break` and the loop spins forever while this test stays green. Running the loop directly makes its completion observable; `Start` is still covered by `TestMonitorLoopReportsPluginResults`.

```suggestion
	startedAt := time.Now()
	// Start is not used here: monitorLoop runs directly so its return is observable.
	// Stop would deadlock, because the closed-channel path never calls tomb.Done.
	go c.plugin.Run()
	returned := make(chan struct{})
	go func() {
		c.monitorLoop()
		close(returned)
	}()

	status := receiveStatus(t, c.statusChan)
```

pkg/custompluginmonitor/custom_plugin_monitor_test.go
```diff
@@ -17,15 +17,680 @@ limitations under the License.
...
+	assert.Equal(t, testSource, status.Source)
+	assert.Empty(t, status.Events)
+	require.Len(t, status.Conditions, 2)
+	for _, cond := range status.Conditions {
+		assert.Equal(t, types.False, cond.Status)
```
hakman, line 612: These are the only conditions in the suite with a real initialization time — everywhere else `newTestMonitor` overwrites `Transition` with `testEpoch`. Drop the assignment at `custom_plugin_monitor.go:321` and every condition ships a zero `LastTransitionTime`, with the suite still passing. Uses the `startedAt` captured above.

```suggestion
		assert.Equal(t, types.False, cond.Status)
		assert.WithinRange(t, cond.Transition, startedAt, time.Now())
```

pkg/custompluginmonitor/custom_plugin_monitor_test.go
```diff
@@ -17,15 +17,680 @@ limitations under the License.
...
+	}
+
+	// Stopping the plugin closes its result channel.
+	c.plugin.Stop()
+	requireNoStatus(t, statusChan)
```
hakman: This is what makes the test check what its name says. The deadline keeps a regression here a focused failure instead of a hang that times out the whole run.

```suggestion
	c.plugin.Stop()
	requireNoStatus(t, c.statusChan)
	select {
	case <-returned:
	case <-time.After(testWait):
		t.Fatal("monitorLoop did not return after the result channel closed")
	}
```

hakman (changes requested):
Thanks @DigitalVeer, I added a few suggestions.

pkg/custompluginmonitor/custom_plugin_monitor_test.go
DigitalVeer: Thanks! I agree with all your suggestions and have applied them.

hakman:
There will be quite useful, thanks @DigitalVeer.
/lgtm
/approve

PR #1333, opened 2026-08-03
feat: run HTTP server standalone without Kubernetes

### LINKED ISSUE #1332: Flag --enable-k8s-exporter=false causes loss of /healthz, /conditions, and /debug/pprof endpoints
When running NPD outside of Kubernetes (e.g. on a bare host or in a VM without a cluster), passing `--enable-k8s-exporter=false` is required to avoid a panic at startup. However, this flag also silently removes the HTTP diagnostic server on `--port` (default `20256`), losing access to `/healthz`, `/conditions`, and `/debug/pprof/`.

## Root cause

`startHTTPReporting()`, the function that starts the `/healthz`, `/conditions`, and pprof handlers, is called exclusively inside `k8sexporter.NewExporterOrDie()`.
When `--enable-k8s-exporter=false`, that function returns `nil` immediately and the HTTP server is never started, even though `--port` is set to a non-zero value.

## Steps to reproduce

```sh
❯ docker run --rm --name npd --privileged -v /dev/kmsg:/dev/kmsg:ro registry.k8s.io/node-problem-detector/node-problem-detector:v1.36.0 \
  --enable-k8s-exporter=false \
  --prometheus-port=20257 \
  --config.custom-plugin-monitor=config/custom-plugin-monitor.json

❯ docker exec -ti npd bash
# Prometheus metrics work:
$ curl -Is http://127.0.0.1:20257/metrics
HTTP/1.1 200 OK
Content-Type: text/plain; version=0.0.4; charset=utf-8; escaping=underscores
Date: Mon, 03 Aug 2026 12:06:56 GMT
Content-Length: 1255

# But the HTTP server is gone:
$ curl http://127.0.0.1:20256/healthz
curl: (7) Failed to connect to 127.0.0.1 port 20256 after 0 ms: Couldn\'t connect to server
$ curl http://127.0.0.1:20256/conditions
curl: (7) Failed to connect to 127.0.0.1 port 20256 after 0 ms: Couldn\'t connect to server
$ curl http://127.0.0.1:20256/debug/pprof/
curl: (7) Failed to connect to 127.0.0.1 port 20256 after 0 ms: Couldn\'t connect to server
$
```

## Expected behavior

`/healthz`, `/conditions`, and `/debug/pprof` should be available on `--port` regardless of whether `--enable-k8s-exporter` is true or false.
The Prometheus metrics endpoint (problem_counter, problem_gauge, etc.) already works without Kubernetes because those metrics are written directly by the problem daemons via OpenCensus; the HTTP diagnostic server should behave the same way.

## Fix

Move the HTTP server out of the K8s exporter into a standalone httpexporter that tracks conditions in-memory via `ExportProblems` and is started independently of `--enable-k8s-exporter`.

tobiasgiese:
fixes #1332

Extract /healthz, /conditions and /debug/pprof into a new httpexporter package so those endpoints are available when --enable-k8s-exporter=false. Previously the HTTP server was started only inside the k8s exporter, making it unavailable in non-Kubernetes environments. The new httpexporter tracks conditions in-memory via ExportProblems and starts on --port regardless of whether the Kubernetes API server is reachable.

### DIFF
--- a/cmd/nodeproblemdetector/node_problem_detector.go
+++ b/cmd/nodeproblemdetector/node_problem_detector.go
@@ -25,6 +25,7 @@ import (
 	_ "k8s.io/node-problem-detector/cmd/nodeproblemdetector/problemdaemonplugins"
 	"k8s.io/node-problem-detector/cmd/options"
 	"k8s.io/node-problem-detector/pkg/exporters"
+	"k8s.io/node-problem-detector/pkg/exporters/httpexporter"
 	"k8s.io/node-problem-detector/pkg/exporters/k8sexporter"
 	"k8s.io/node-problem-detector/pkg/exporters/prometheusexporter"
 	"k8s.io/node-problem-detector/pkg/problemdaemon"
@@ -58,6 +59,10 @@ func npdMain(ctx context.Context, npdo *options.NodeProblemDetectorOptions) erro
 	if pe := prometheusexporter.NewExporterOrDie(npdo); pe != nil {
 		defaultExporters = append(defaultExporters, pe)
 		klog.Info("Prometheus exporter started.")
+	}
+	if he := httpexporter.NewExporterOrDie(npdo); he != nil {
+		defaultExporters = append(defaultExporters, he)
+		klog.Info("HTTP exporter started.")
 	}
 
 	plugableExporters := exporters.NewExporters()
--- a/cmd/nodeproblemdetector/node_problem_detector.go
+++ b/cmd/nodeproblemdetector/node_problem_detector.go
@@ -25,6 +25,7 @@ import (
 	_ "k8s.io/node-problem-detector/cmd/nodeproblemdetector/problemdaemonplugins"
 	"k8s.io/node-problem-detector/cmd/options"
 	"k8s.io/node-problem-detector/pkg/exporters"
+	"k8s.io/node-problem-detector/pkg/exporters/httpexporter"
 	"k8s.io/node-problem-detector/pkg/exporters/k8sexporter"
 	"k8s.io/node-problem-detector/pkg/exporters/prometheusexporter"
 	"k8s.io/node-problem-detector/pkg/problemdaemon"
@@ -51,6 +52,10 @@ func npdMain(ctx context.Context, npdo *options.NodeProblemDetectorOptions) erro
 
 	// Initialize exporters.
 	defaultExporters := []types.Exporter{}
+	if he := httpexporter.NewExporterOrDie(npdo); he != nil {
+		defaultExporters = append(defaultExporters, he)
+		klog.Info("HTTP exporter started.")
+	}
 	if ke := k8sexporter.NewExporterOrDie(ctx, npdo); ke != nil {
 		defaultExporters = append(defaultExporters, ke)
 		klog.Info("K8s exporter started.")
--- a/pkg/exporters/httpexporter/http_exporter.go
+++ b/pkg/exporters/httpexporter/http_exporter.go
@@ -1,0 +1,104 @@
+/*
+Copyright 2026 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package httpexporter
+
+import (
+	"net"
+	"net/http"
+	"net/http/pprof"
+	"strconv"
+	"sync"
+
+	"k8s.io/klog/v2"
+
+	"k8s.io/node-problem-detector/cmd/options"
+	"k8s.io/node-problem-detector/pkg/types"
+	"k8s.io/node-problem-detector/pkg/util"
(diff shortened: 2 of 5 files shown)


tobiasgiese:
With a local built image I can see

```sh
❯ docker run --rm --name npd --privileged -v /dev/kmsg:/dev/kmsg:ro gcr.io/k8s-staging-npd/node-problem-detector:v1.37.0-alpha.1-31-g7bff5a4c-dirty \
  --enable-k8s-exporter=false \
  --prometheus-port=20257 \
  --config.custom-plugin-monitor=config/custom-plugin-monitor.json

❯ docker exec -ti npd bash
root@5bcbcdda7f64:/# apt update -qq; apt install -qq -y curl jq &>/dev/null
2 packages can be upgraded. Run 'apt list --upgradable' to see them.
root@5bcbcdda7f64:/# curl -s http://127.0.0.1:20256/conditions | jq
[
  {
    "type": "ReadonlyFilesystem",
    "status": "False",
    "transition": "2026-08-03T12:41:01.208945846Z",
    "reason": "FilesystemIsNotReadOnly",
    "message": "Filesystem is not read-only"
  },
  {
    "type": "KernelDeadlock",
    "status": "False",
    "transition": "2026-08-03T12:41:01.208911429Z",
    "reason": "KernelHasNoDeadlock",
    "message": "kernel has no deadlock"
  },
  {
    "type": "XfsShutdown",
    "status": "False",
    "transition": "2026-08-03T12:41:01.208911554Z",
    "reason": "XfsHasNotShutDown",
    "message": "XFS has not shutdown"
  },
  {
    "type": "CperHardwareErrorFatal",
    "status": "False",
    "transition": "2026-08-03T12:41:01.208911596Z",
    "reason": "CperHardwareHasNoFatalError",
    "message": "UEFI CPER has no fatal error"
  },
  {
    "type": "NTPProblem",
    "status": "Unknown",
    "transition": "2026-08-03T12:41:01.214200471Z",
    "reason": "NTPIsUp",
    "message": "Could not find 'systemctl' - require systemd"
  }
]
```

pkg/exporters/httpexporter/http_exporter.go
```diff
@@ -0,0 +1,98 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package httpexporter provides a standalone HTTP server exposing /healthz,
+// /conditions and /debug/pprof without requiring a Kubernetes API server.
+package httpexporter
```
DigitalVeer, line 17: Can we add tests for this package?

cmd/nodeproblemdetector/node_problem_detector.go
DigitalVeer, line 60: We should move initializing the httpexporter above k8sexporter. If k8s is first, the apiserver connection can hang for `--apiserver-wait-timeout` (5 minutes by default). We could have no access to `/healthz` during that time.

This is reproducible by pointing `--apiserver-override` at an unroutable address (so the connection hangs) with a 20s wait timeout. `/healthz` refuses connections during the wait and answers right after the timeout expires.

```suggestion
      if he := httpexporter.NewExporterOrDie(npdo); he != nil {
              defaultExporters = append(defaultExporters, he)
              klog.Info("HTTP exporter started.")
      }
      if ke := k8sexporter.NewExporterOrDie(ctx, npdo); ke != nil {
              defaultExporters = append(defaultExporters, ke)
              klog.Info("K8s exporter started.")
      }
      if pe := prometheusexporter.NewExporterOrDie(npdo); pe != nil {
              defaultExporters = append(defaultExporters, pe)
              klog.Info("Prometheus exporter started.")
      }
```

DigitalVeer (changes requested):
Thanks for the changes! We should also update the README. The README still has `--address` and `--port` under the ["For Kubernetes exporter" heading](https://github.com/kubernetes/node-problem-detector/blob/5fde43a6a1994ee5356a57105c65ad62c794494d/README.md#L113-L126).

cmd/nodeproblemdetector/node_problem_detector.go
tobiasgiese: ~~Why do you think this is important? If we don't have access to the `kube-apiserver`, I'm not sure the `/healthz` endpoint is particularly relevant either, right?~~

~~For the Kubernetes use case, we should have access to the `kube-apiserver`, so the exporter would start immediately.~~

~~Do you see any edge cases where this could cause issues? Specifically, what problem would you expect if the `/healthz` endpoint isn't accessible during that time? I assume the exporter would simply be considered unhealthy until it becomes accessible.~~

I just reread the code and you're right. Let me update it

pkg/exporters/httpexporter/http_exporter.go
tobiasgiese: We don't have unit-tests for the other exporters. I can start adding unit-tests for sure, maybe it makes sense to cover unit-tests for the other exporters in a followup as well.

DigitalVeer (approved):
The HTTP handler you removed seems to have been the only caller for [`ConditionManager.GetConditions`](https://github.com/kubernetes/node-problem-detector/blob/dc0018e5fbda29c1f663386cfdf7579a03d97b23/pkg/exporters/k8sexporter/condition/manager.go#L103-L111). 

In [manager.go](https://github.com/kubernetes/node-problem-detector/blob/master/pkg/exporters/k8sexporter/condition/manager.go), we should drop the interface method, implementation, and lock comment  (line 66). We can also drop `TestGetConditions`.

You can do this in this PR or a follow-up.

PR #1338, opened 2026-08-23
Fix log pattern end anchoring

hakman:
`CompilePattern` previously appended `\z` directly to each configured expression. Because `|` has lower precedence, `oom-kill|Out of memory\z` anchored only the second branch. The first branch could match a retained line again after every subsequent push until that line was overwritten, repeating temporary events and log-counter results.

Group the expression before adding the end anchor:

- **Before:** `oom-kill|Out of memory\z`
- **After:** `(?:oom-kill|Out of memory)\z`

This makes every branch honor the end-of-buffer contract and removes the extra parse-tree comparison. Shipped configurations are unaffected because they already group their alternatives. Custom patterns with an unparenthesized top-level `|` intentionally change behavior; use `(?:oom-kill|Out of memory).*` when either phrase should match anywhere before the end of the newest line.

/cc @DigitalVeer

### DIFF
--- a/pkg/systemlogmonitor/log_buffer.go
+++ b/pkg/systemlogmonitor/log_buffer.go
@@ -17,6 +17,7 @@ limitations under the License.
 package systemlogmonitor
 
 import (
+	"fmt"
 	"regexp"
 	"regexp/syntax"
 	"slices"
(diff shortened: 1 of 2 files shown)


pkg/systemlogmonitor/log_buffer.go
```diff
@@ -74,7 +74,8 @@ func CompilePattern(expr string) (*Pattern, error) {
 	if _, err := regexp.Compile(expr); err != nil {
 		return nil, err
 	}
-	anchored := expr + `\z`
+	// Group the expression so the end anchor applies to every top-level branch.
+	anchored := `(?:` + expr + `)\z`
 	reg, err := regexp.Compile(anchored)
 	if err != nil {
 		return nil, err
```
DigitalVeer, line 76: Everything LGTM. Only one comment - I think with this change we should wrap the error:

```
fmt.Errorf("invalid pattern %q: %w", expr, err)
```

This is based on running a before/after with a string with \Q in it:
```
old  regexp.Compile(`error\Q` + `\z`)     → no err
new  regexp.Compile(`(?:error\Q)` + `\z`) → missing closing ): `(?:error\Q)\z`
```

The error message shows our internal `(?:...)\z` pattern which the user did not write.

hakman: good point, will update. thanks for checking this.

hakman:
CC @ameukam @DigitalVeer - one more LGTM, please

PR #1346, opened 2026-08-25
README.md Remove the old badges add single new line

AlexBaranowski:
Howdy,

- As far as I see the repository does not use TRAVIS.ci anymore
- The goreportcard is sunset

Best,
Alex


hakman (approved):
Thanks for the update!

